PR #360, opened 2013-10-29
Basic http preauthentication

andris-szatmari:
Provides an optional basic http authentication.
It can be used to host private packagist service behind any kind of http authentication (htpasswd, apache with mod_ldap, etc).
This modification solves #265 without having bigger modifications on the current running instances.

The two new options are preauthenticated_provider.enabled and preauthenticated_provider.default_email_domain.
With the enabled flag you can enable http authentication listener and user provider. If you leave it false, no further modifications required.
If enabled, the userprovider is being put behind a userproviderproxy object, which proxies all the method calls to the userprovider instance, but when no user is found, it creates one.
The default_email_domain is used to create an email address for the new user as email is not nullable.

### DIFF
--- a/src/Packagist/WebBundle/Package/Dumper.php
+++ b/src/Packagist/WebBundle/Package/Dumper.php
@@ -82,11 +82,11 @@ class Dumper
     /**
      * Constructor
      *
-     * @param RegistryInterface $doctrine
-     * @param Filesystem $filesystem
+     * @param RegistryInterface     $doctrine
+     * @param Filesystem            $filesystem
      * @param UrlGeneratorInterface $router
-     * @param string $webDir web root
-     * @param string $cacheDir cache dir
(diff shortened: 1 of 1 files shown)


src/Packagist/WebBundle/Security/Provider/PreAuthenticatedUserProviderProxy.php
```diff
@@ -0,0 +1,159 @@
...
+     * @param string $userName
+     *
+     * @return \Symfony\Component\Security\Core\User\UserInterface
+     */
+    private function autoCreateUserWithDefaultData($userName)
+    {
+        $user = $this->userManager->createUser();
+        /* @var PackagistUser $user */
+        $user->setUsername($userName);
+        $user->setPlainPassword($user->getUsername());
```
xproax: 👍

PR #361, opened 2013-10-29
Remove password fields when logged in with Github

dshoreman:
If a user has used Github authentication to create their account, they are
not able to update their email as a password does not exist. This resolves the
issue by removing the Change Password form, and also the Password input on
the Edit Profile page.

One possible improvement that could be made in future would be additional
checks to ensure the user is still authenticated with Github, though I've
spent all day on this already so I'll leave that to somebody else to add in
;)

### DIFF
--- a/src/Packagist/WebBundle/Form/Type/ProfileFormType.php
+++ b/src/Packagist/WebBundle/Form/Type/ProfileFormType.php
@@ -52,20 +52,13 @@ class ProfileFormType extends BaseType
         $builder->add('failureNotifications', null, array('required' => false, 'label' => 'Notify me of package update failures'));
     }
 
-    private function addPasswordField(&$event)
+    private function addPasswordField($event)
     {
-        if (class_exists('Symfony\Component\Security\Core\Validator\Constraints\UserPassword')) {
-            $constraint = new UserPassword();
-        } else {
-            // Symfony 2.1 support with the old constraint class
-            $constraint = new OldUserPassword();
-        }
-
         $event->getForm()->add('current_password', 'password', array(
             'label' => 'form.current_password',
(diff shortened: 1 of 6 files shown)


chq81:
Any progress on that? Currently I am having the same problem.

alcohol:
Is this PR still relevant since the rework/redesign?

Seldaek:
I believe so yes, the problem is still there, that accounts initially registered with github need to use the password reset form to create a password before they can change their username or email. Not a very common problem though, but it is still there.

src/Packagist/WebBundle/Form/Type/ProfileFormType.php
```diff
@@ -14,19 +14,61 @@
...
 
         $builder->add('failureNotifications', null, array('required' => false, 'label' => 'Notify me of package update failures'));
     }
 
+    private function addPasswordField(&$event)
+    {
+        if (class_exists('Symfony\Component\Security\Core\Validator\Constraints\UserPassword')) {
+            $constraint = new UserPassword();
+        } else {
+            // Symfony 2.1 support with the old constraint class
```
xproax: #758 
[patternpack-1.0.6.zip](https://github.com/composer/packagist/files/691889/patternpack-1.0.6.zip)

xproax:
[patternpack-1.0.6.zip](https://github.com/composer/packagist/files/691892/patternpack-1.0.6.zip)

alcohol:
@xproax what are you attempting to do? Approving this won't help, since it needs to be rebased.

dshoreman:
If this is still relevant and someone wants to take over, feel free. I'm unlikely to have time in the foreseeable future.

Seldaek:
Fixed by #895

PR #447, opened 2014-10-08
Add quick links for other login options in top bar

ss23:


Seldaek:
Made irrelevant by the new design

PR #456, opened 2014-11-12
Keep dev-versions up to date for valid branches

aminin:
Don't block all package updates due to one invalid branch


Seldaek:
I don't really like this because it means people will ignore failures but ok I can see in some cases it's impractical so I'll merge, thanks.

PR #463, opened 2014-12-10
New splitting strategy for providers

lavoiesl:
Old format:
 - **latest**: until 10 hours before 1st day of current month.
 - **active**: 10-60 hours before 1st day of current month.
 - **stale**: 60-180 hours before 1st day of current month.
 - **archived**: more than 180 hours before 1st day of current month.

New format:
 - **latest**: until monday last week.
 - **YYYY-MM**: month block (03, 06, 09, 12), excluding what is in latest.
 - **YYYY**: year, excluding current
 - **archived**: Everything older than 2012

Rationale:

 * Split into smaller groups, resulting in faster download. This becomes a problem when Packagist is overloaded and/or the user has a slow connection - Composer times out.
 * Greatly reduces the changes to old providers. With the previous strategy, packages were switching constantly, causing cascading changes to all providers.

There is still the problem that the dumper relies on file modification time, it would be best to switch to actual package modification date.

### DIFF
--- a/src/Packagist/WebBundle/Package/SymlinkDumper.php
+++ b/src/Packagist/WebBundle/Package/SymlinkDumper.php
@@ -552,27 +552,35 @@ class SymlinkDumper
 
     private function getTargetListing($file)
     {
-        static $firstOfTheMonth;
-        if (!$firstOfTheMonth) {
-            $date = new \DateTime;
-            $date->setDate($date->format('Y'), $date->format('m'), 1);
-            $date->setTime(0, 0, 0);
-            $firstOfTheMonth = $date->format('U');
+        static $limitLatest, $thisYear, $limitArchived;
+
+        if (!$limitLatest) {
+            $limitLatest   = new \DateTime('monday last week');
+            $thisYear      = new \DateTime(date('Y') . '-01-01');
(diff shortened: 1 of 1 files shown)


Seldaek:
I deployed this last week, and to be honest the results in terms of traffic are mitigated. I did not see a conclusive change for better or worse. I think the difference might be that given our package number and especially the rate of changes packages keep getting bounced from the 2014 provider to the 2015 one for example (if a package had no commit for >2months then suddenly gets one), every time it changes both files.

lavoiesl:
I understand what you are talking about, the problem is definitely more visible at the beginning of the year, perhaps splitting the previous year in 2 could help, I’m not sure.

What I know though, is that the first download is only a tiny bit slower, but much more reliable since the individual downloads never fail anymore. The following updates are also smaller. Generally, I am very pleased with the changes. Perhaps the problem is more visible to me because my internet connection is so bad. ;)

I have an idea, I will get back to you.

Seldaek:
Well I think if it really lead to a lot less bandwidth usage I would see it. I actually checked log files and summed up all traffic from provider files and overall it didn't go down (nor up) since the change.

lavoiesl:
Here @Seldaek, #474 should fix that issue.

PR #469, opened 2015-01-13
Remove dead compiler pass

webflo:
The RepositoryPass is not used anywhere. The service was removed in fec4475 and 90e70e5 a long time ago.


stof:
Why is it marked as a WIP ?

webflo:
Because i thought there it other unused code. But thats not the case. I think this is ready.

Seldaek:
Thanks

PR #473, opened 2015-02-10
Fix: Undefined classes

localheinz:
This PR

* [x] fixes an inline comment which referred to non-existent `Entity\Maintainer`, but should be `Entity\User`
* [x] imports `Doctrine\ORM\NoResultException` so `NoResultException` can actually be caught
* [x] removes a condition depending on an interface which was removed in `hwi/oauth-bundle:0.3.0-alpha1`


Seldaek:
Thanks! Although FYI the HWI oauth removal was a bit greedy, I restored it in b6175ae2ce8f9575c4b5132e9b59273999e323e4

PR #474, opened 2015-02-13
Splitting strategy: always keep 4 last trimesters

lavoiesl:
This is a respin of #463

@Seldaek mentioned that the 2014 listing was always getting updated, this is true because we are at the beginning of the year.

This fixes it by calculating the last 4 trimesters before today. The rest of the logic remains unchanged.

This PR also adds a `getTargetListingBlocks` method that declares all blocks, this simplifies the code of `getTargetListing` and allows for easier testing.

And yeah, unit tests were added too.

Sample of generated blocks:
```
2014-12-31:
    'latest'  => '2014-12-22'
    '2014-10' => '2014-10-01'
    '2014-07' => '2014-07-01'
    '2014-04' => '2014-04-01'
    '2014-01' => '2014-01-01'
    '2013'    => '2013-01-01'
    '2012'    => '2012-01-01'

2015-01-01:
    'latest'  => '2014-12-22'
    '2015-01' => '2015-01-01'
    '2014-10' => '2014-10-01'
    '2014-07' => '2014-07-01'
    '2014-04' => '2014-04-01'
    '2014'    => '2014-01-01'
    '2013'    => '2013-01-01'
    '2012'    => '2012-01-01'

2015-05-31:
    'latest'  => '2015-05-25'
    '2015-04' => '2015-04-01'
    '2015-01' => '2015-01-01'
    '2014-10' => '2014-10-01'
    '2014-07' => '2014-07-01'
    '2014'    => '2014-01-01'
    '2013'    => '2013-01-01'
    '2012'    => '2012-01-01'
```

### DIFF
--- a/src/Packagist/WebBundle/Tests/Package/SymlinkDumperTest.php
+++ b/src/Packagist/WebBundle/Tests/Package/SymlinkDumperTest.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Packagist\WebBundle\Tests\Package;
+
+class SymlinkDumperTest extends \PHPUnit_Framework_TestCase
+{
+    private $mockDumper;
+
+    public function setUp()
+    {
+        $this->mockDumper = $this->getMockBuilder('Packagist\WebBundle\Package\SymlinkDumper')
+            ->disableOriginalConstructor()
+            ->getMock();
+    }
+
+    public function tearDown()
+    {
+        $this->mockDumper = null;
+    }
+
+    /**
+     * @dataProvider getTestGetTargetListingBlocks
+     */
+    public function testGetTargetListingBlocks($now, array $expected)
+    {
+
+        $blocks = self::invoke($this->mockDumper, 'getTargetListingBlocks', $now);
+
+        $blocks = array_map(function($timestamp) { return date('Y-m-d', $timestamp); }, $blocks);
+
+        $this->assertEquals($expected, $blocks);
+    }
+
+    public function getTestGetTargetListingBlocks()
+    {
+        return array(
+            array(
+                strtotime('2014-12-31'),
+                array(
+                    'latest'  => '2014-12-22',
+                    '2014-10' => '2014-10-01',
+                    '2014-07' => '2014-07-01',
+                    '2014-04' => '2014-04-01',
+                    '2014-01' => '2014-01-01',
+                    '2013'    => '2013-01-01',
+                    '2012'    => '2012-01-01',
+                ),
+            ),
+            array(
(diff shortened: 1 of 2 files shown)


src/Packagist/WebBundle/Tests/Package/SymlinkDumperTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+    protected static $dumper;
+
+    public static function setUpBeforeClass()
+    {
+         self::$kernel = static::createKernel();
+         self::$kernel->boot();
+
+         $container = self::$kernel->getContainer();
+         self::$dumper = $container->get('packagist.package_dumper');
```
stof: instead of booting a kernel here and getting the full object graph, you should rather create mocks for dependencies to make it a unit test (your tested code does not involve any of the dependencies anyway)

lavoiesl: Good point, fixed.

Seldaek:
Alright thanks let's see what happens.

Seldaek:
Seems the file size distribution is better like this at least.. but it remains to be seen if it will help bandwidth usage at all in average update conditions :)

```
  184673 Feb 14 16:14 p-provider-2012.json         
  889429 Feb 14 16:14 p-provider-2013.json         
  546985 Feb 14 16:14 p-provider-2014-04.json      
  693177 Feb 14 16:14 p-provider-2014-07.json      
 1086016 Feb 14 16:14 p-provider-2014-10.json      
  452788 Feb 14 16:14 p-provider-2014.json         
  702896 Feb 14 16:14 p-provider-2015-01.json      
    1356 Feb 14 16:14 p-provider-archived.json     
  729344 Feb 14 16:14 p-provider-latest.json       
    1181 Feb 14 16:15 packages.json                
```

lavoiesl:
Seems like archived is way too small, it can be merged with 2012 which is also small. Change https://github.com/composer/packagist/pull/474/files#diff-d9ebe9087825484f66586f90f115c12aR572 to 2013.

Seldaek:
Right good point that'll save a useless request, I'll re-dump and then re-evaluate traffic in a few days (weekends are anyway fairly low traffic comparatively so it's hard to tell).

lavoiesl:
So, have you seen improvements?

Seldaek:
Yeah I am not entirely sure if it's just coincidental because of less usage or not, don't really have time to run stats now but overall traffic was lower so that's good :)

lavoiesl:
Alright, keep me posted for next week. On my side, it is much smoother.

PR #476, opened 2015-02-19
Added JSONP support to search

rdohms:
`/search.jsonp?q=something&callback=my_call`

Will return a valid JSONP response that can be read from a JS call.


stof:
``/search.jsonp`` looks weird to me. Why not keeping ``search.json?callback=my_call``, which would also keep the whole logic much simpler ?

Btw, your commit appear to be done by a user named ``travis-ci`` with an email not associated with your account. Is it expected ?

rdohms:
Ah dang.... i messed up my config it seems.

As for json or jsonp, i'm not sure what standard practice is, i can go either way.

rdohms:
@stof I do see here some implementations doing that, only adding the callback parameter. I'll adjust this and open a new PR

stof:
Well, you could just update this PR though rather than opening a new one

PR #477, opened 2015-02-19
Adding JSONP support

rdohms:
`search.json?q=something&callback=my_call`

Will trigger the addition of a wrapper on the json return, allowing api calls from JS.


stof:
do we have other actions where the jsonp support might be a win ?

rdohms:
@stof probably, but i do not have the freetime to allocate to fixing all of them, this PR could be a "this is how you do it" guide, but i can't guarantee more time this week.

Flyingmana:
already noted this via twitter, CORS may be the more modern solution here.
http://stackoverflow.com/questions/12296910/so-jsonp-or-cors

PR #478, opened 2015-02-28
Prepare WebBundle for subtree split

derhasi:
For Drupal we are currently building a [Drupal Packagist](https://github.com/drupal-composer/drupal-packagist). At the moment this is an ugly fork of the original Packagist. For improving that and being able to keep up to date, it would be great, the WebBundle could be available as a subtree split. Therefore I added a composer.json to src/Packagist/WebBundle.

I'm currently checking out, if we can run an automated build of the subtree split, e.g. via Travis CI, but that may be some task for a separate issue.


derhasi:
Obsolete.

PR #480, opened 2015-03-18
save a lot of memory when retrive packages.json file

Zjie:
These days, I tried so hard to set up a local composer repo by packagist. But the memory problem happened after adding to much package to local repo. 
Could you believe that a single request to get packages.json would took up over 1G memory?
So that why I made this pull request.
I thought this would help other guys to solve these memory-problem.

### DIFF
--- a/src/Packagist/WebBundle/Entity/PackageRepository.php
+++ b/src/Packagist/WebBundle/Entity/PackageRepository.php
@@ -239,17 +239,9 @@ class PackageRepository extends EntityRepository
     public function getFullPackages(array $ids = null, $filters = array())
     {
         $qb = $this->getEntityManager()->createQueryBuilder();
-        $qb->select('p', 'v', 't', 'a', 'req', 'devReq', 'sug', 'rep', 'con', 'pro')
+        $qb->select('p', 'v')
             ->from('Packagist\WebBundle\Entity\Package', 'p')
             ->leftJoin('p.versions', 'v')
-            ->leftJoin('v.tags', 't')
-            ->leftJoin('v.authors', 'a')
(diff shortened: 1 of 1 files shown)


stof:
This is not getting full packages anymore, so the function is lying (and you will make thousands of SQL requests because of the lazy-loading)

And Packagist is not meant to give you a packages.json generated on the fly with lots of packages (there is no way this can work properly, be it for memory or time).
Thus, the packages.json file generated on the fly contains the whole content of the DB, which means your composer client will also have to load all packages in memory. When using the dump command, it dumps into a more advanced format for the packages.json, allowing composer to load only a small subset of the metadata (on packagist.org, loading the whole metadata in memory in the client would require way more than 1GB).

You should really be dumping the repository files with the command.

Btw, on packagist.org, the code will reject the fallback logic entirely: https://github.com/composer/packagist/blob/17530758db98e0a63f0fae9f96662fda02602ec2/src/Packagist/WebBundle/Controller/ApiController.php#L43-57
@Seldaek would it be worth considering that Packagist should always rely on the dumped files and removing the fallback controller entirely ? Or at least adding the alert-level logging for any host ?

Btw, note that hosting your own Packagist instance is not a supported use case, so it is at your own risk: https://getcomposer.org/doc/05-repositories.md#packagist

Zjie:
Thanks for your explanation. I properly misunderstood the functionality of packages.json. The reason why I tried to set up a local repo is that development machine can not access the internet and a private repo will help to solve this problem.
In any case, thank you so much.

stof:
@Zjie using ToranProxy or Satis would better fit your use case IMO.

Issue #365, opened 2013-11-20
Sorting of search results


nfx:
In order to get better package selection, it would be good to sort results based on downloads, number of stars, forks, any other things that promote package maturity.

marceloandrader:
+1 if you need some hands to help also

MTon:
+1

Grynn:
+1

forsethc:
+1

Arsey:
+1

Kampfbereit:
+1

jakubpas:
+1

stopsopa:
+1

stopsopa:
i have my brutal solution for this.
- go to page http://httpd.pl/bookmarklets#toc_11
- drag button "packagistsort" to bookmarks toolbar in browser
- search eg. "curl" in packagist.org
- press just created button in bookmarks "packagistsort"

Please don't use this when phrase is not precise and packagist return 100 pages of results

laland:
+1

vimishor:
I created a [gist](https://gist.github.com/vimishor/e770294b39f4e83a3136) with a (_temporary_) solution for CLI, which I'm using. Maybe someone could find it useful.

nickdietel:
+1

Seldaek:
You can now sort by Stars and Downloads

justinkwanlee:
@Seldaek OMG THIS IS THE GREATEST WOOOOOOO

jakubpas:
Cool ;-)

saurini:
THANK YOU SO MUCH!

stopsopa:
Great :)

BlakeGardner:
So much yes

stopsopa:
great, and thanks to remove my solution ;) indeed it's not useful anymore.

zonky2:
hmm ...??
I can't find anything to sort on the site https://packagist.org/

I would like to sort by 
* last added
* by date of release

stonetwig:
How do you sort by downloads / stars? It's not very obvious

alexkuc:
I think the feature used to be there (sorting) but it is gone now https://github.com/composer/packagist/issues/818 ?

PR #481, opened 2015-03-19
 Sorting of search results

marceloandrader:
This pull request has the changes needed to support sort by downloads or favorites count when searching packages.

By downloads:

by-download

By favorites:

by-favorites

I've added the downloads and favorites field on the schema.xml on solr and on the packagist:index action to get those values and send them to solr. So when this change goes to production a full re-index will be needed.

Not sure if this can be a good thing for the repo, but also included some files to create the packagist dev environment as docker contaiers using `docker-compose` to orchestrate the 4 services (web, database, redis, solr) and be up and running with 1 command.

### DIFF
--- a/doc/schema.xml
+++ b/doc/schema.xml
@@ -236,6 +236,8 @@
    <field name="description" type="text_general_rev" indexed="true" stored="true"/>
    <field name="tags" type="text_general_rev" indexed="true" stored="true" multiValued="true"/>
    <field name="type" type="text_general_rev" indexed="true" stored="true"/>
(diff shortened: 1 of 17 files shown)


marceloandrader:
This pull request fixes the issue #365

Seldaek:
Thanks @marceloandrader but it turns out #493 came about with a slightly better solution to this so I merged that one instead. In any case the feature is now available on packagist.org :)

marceloandrader:
No problem @seldaek what is important is that we have the feature now. Thanks for merging.

PR #493, opened 2015-05-04
Sorting of search results by multiple ordered fields

bmichalski:
This pull request fixes the issue #365, in a slight different way than https://github.com/composer/packagist/pull/481.

It enables to sort by multiple ordered field, using HTML or JSON.

Even though the user interface does not permit it, as it may not be user friendly, the API does support a combination of ordering parameters.

Even if at the moment it may not be needed, it might come in handy when you decide to add other criterias to order by.

Below are some screenshots:

By downloads:

screenshot from 2015-05-04 21 11 47

By favorites:

screenshot from 2015-05-04 21 12 36

It also comes with automated tests.

Please be aware that these tests should not be run on a production machine, as they first drop / create the database to ensure isolation.

### DIFF
--- a/src/Packagist/WebBundle/Command/IndexPackagesCommand.php
+++ b/src/Packagist/WebBundle/Command/IndexPackagesCommand.php
@@ -12,8 +12,6 @@
 
 namespace Packagist\WebBundle\Command;
 
-use DateTime;
-use Exception;
 use Packagist\WebBundle\Entity\Package;
 use Packagist\WebBundle\Model\DownloadManager;
 use Packagist\WebBundle\Model\FavoriteManager;
@@ -128,8 +126,8 @@ class IndexPackagesCommand extends ContainerAwareCommand
                     $this->updateDocumentFromPackage($document, $package, $redis, $downloadManager, $favoriteManager);
                     $update->addDocument($document);
 
-                    $package->setIndexedAt(new DateTime);
-                } catch (Exception $e) {
+                    $package->setIndexedAt(new \DateTime);
+                } catch (\Exception $e) {
                     $output->writeln('<error>Exception: '.$e->getMessage().', skipping package '.$package->getName().'.</error>');
                 }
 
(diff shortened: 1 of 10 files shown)


stof:
This PR conflicts with master. Looks like you need a rebase

src/Packagist/WebBundle/Command/IndexPackagesCommand.php
```diff
@@ -12,12 +12,16 @@
 
 namespace Packagist\WebBundle\Command;
 
+use DateTime;
+use Exception;
```
stof: this goes against the Symfony coding standards, which are used in the Composer projects. It should be reverted

src/Packagist/WebBundle/Controller/WebController.php
```diff
@@ -248,6 +248,111 @@ protected function setupPager($query, $page)
     }
 
     /**
+     * @param mixed $orderBys
```
stof: this cannot be mixed. It requires passing an array

src/Packagist/WebBundle/Resources/public/js/search.js
```diff
@@ -49,11 +57,57 @@
...
+
+            pushStateArguments = [
+                null,
+                'Search',
+                urlPrefix + '/search/?q=' + q + joinedOrderBysQryStrPart
+            ];
+
             if (firstQuery) {
-                window.history.pushState(null, "Search", "/search/?q=" + encodeURIComponent($('input[type="search"]', form).val()));
+                window.history.pushState.apply(window.history, pushStateArguments);
```
stof: you don't really need to use ``.apply`` here. You could just use ``.pushState(null, "Search", url)``, after creating the ``url`` variable

bmichalski:
I fixed the PR according to your comments.
I also merged composer/packagist master into my branch, there should no longer be a conflict.

Seldaek:
Thanks @bmichalski! I did a couple tweaks as you can see in the commit linked above, but overall it's working quite well and will surely help people make better decisions.

PR #506, opened 2015-06-12
Make navigation dropdown click area larger

giggsey:
It now fills the entire row, instead of just the text/icon.


Seldaek:
Thanks

PR #515, opened 2015-06-13
Make search field 100% wide on mobile

kix:
This makes the search form fill all the width on mobile devices. A wider search form looks way more usable to me:
Before:
screenshot 2015-06-13 15 52 23
After:
screenshot 2015-06-13 15 52 46


Seldaek:
Thanks, I had to tweak a bit though as you can see in the commit above, because when you actually search we need some space for icons on the right, that's why it wasn't full width in the first place.

PR #518, opened 2015-06-14
Set subpixel antialiasing for navbar

calcinai:
Forcing subpixel-antialiasing stops Safari falling back to traditional anti-aliasing during animation of the dropdown menus.  The usual anti-aliasing switch makes the navbar text appear thin and wiry.


Seldaek:
Thanks

PR #525, opened 2015-06-16
Add a link for Package Versions usage documentation

Soullivaneuh:
Could be nice to have it on quick info homepage.

### DIFF
--- a/src/Packagist/WebBundle/Resources/views/Web/index.html.twig
+++ b/src/Packagist/WebBundle/Resources/views/Web/index.html.twig
@@ -15,6 +15,8 @@
     }
 }</code></pre>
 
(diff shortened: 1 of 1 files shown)


src/Packagist/WebBundle/Resources/views/Web/index.html.twig
```diff
@@ -15,6 +15,8 @@
     }
 }</code></pre>
 
+            <p>For more information about packages versions usage, see <a href="https://getcomposer.org/doc/01-basic-usage.md#package-versions">composer documentation</a>.</p>
```
stof: missing ``the``

Soullivaneuh: Do you mean `the composer documentation`?

stof: yes, so that it reads like an English sentence.

Soullivaneuh:
Commits squashed after fix. Ready to merge.

PR #527, opened 2015-06-17
UI tweaks

jakefolio:
I wanted to make some small tweaks to the layout:
- Re-organize navigation to have sign in at the far right. This also puts "browse" in the first position which is the most common usage
- Increased the font size and padding of search input. Also added a drop shadow for a clean look :)
- Added left border to code samples to be a better call to action.

Mobile Navigation
- Fixed the mobile nav to no longer be floated and to cover entire width of screen
- Removed the right borders since they serve no purpose in mobile nav
- Hide login drop down on mobile, force mobile users to click through to login page

### DIFF
--- a/src/Packagist/WebBundle/Resources/public/css/main.css
+++ b/src/Packagist/WebBundle/Resources/public/css/main.css
@@ -29,7 +29,14 @@ html, body {
 }
 
 pre, code {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, however I tweaked a couple things which didn't fit IMO and reverted your sneaky attempt at switching to php  purple :p

PR #531, opened 2015-06-21
apply use statement instead of full class name

Ma27:


Seldaek:
Why though? It's more line, not really more or less readable..

PR #537, opened 2015-06-23
'Watches' should be 'Watchers'

JayBizzle:


Seldaek:
I guess so, thanks :) It'll go in with the next deploy

PR #539, opened 2015-06-29
Enhancement: Automatically focus on search input

localheinz:
This PR

* [x] sets the value of the `autofocus` attribute of the search query input to `true`, so we can start typing in whatever package we're looking for already

### Before

screen shot 2015-06-28 at 23 08 23

### After

screen shot 2015-06-28 at 23 08 38


Seldaek:
I removed that in 0ac50c70e09d7cf1c53381ac3f75aedd041ddffa because it caused issues with mobile browsers.. I'd be up for merging again if it was done with JS conditionally (i.e. on desktop only) based on window width maybe.

localheinz:
@Seldaek 

Ha, good to know!

localheinz:
@Seldaek 

To be honest, I'd prefer to just tab into the search bar, rather than fiddling with the display device width via JavaScript.

PR #555, opened 2015-08-02
Fix the reference to the doc about how to update packages

stof:
This documentation is now on the about page, not on the profile page.


alcohol:
:+1:

PR #554, opened 2015-07-31
Autofocus search field

frederikbosch:
When visiting Packagist a user most likely wants to search the repository. Therefore, autofocus on the search field will improve usability.

PR #558, opened 2015-08-11
Autofocus search field if width more 992px

andrey1s:
used the size of the desktop http://getbootstrap.com/css/#responsive-utilities

### DIFF
--- a/src/Packagist/WebBundle/Resources/public/js/search.js
+++ b/src/Packagist/WebBundle/Resources/public/js/search.js
@@ -156,4 +156,8 @@
             $('#search_query_query').focus();
         }
     });
+
+    if ($(document).width() >= 992) {
+        $('#search_query_query').focus();
+    }
 }(jQuery));


alcohol:
Refs #554 #539 #309

alcohol:
Could you please fix the codestyle though?

andrey1s:
@alcohol fixed. could you specify the rules for the styles `js`?

Seldaek:
That's probably safe, thanks

frederikbosch:
Wonderful: great work. Many thanks!

alcohol:
Just to point out, this will not work for any of the iPad retina displays for example (or any other handheld device sporting a similar resolution).

frederikbosch:
@alcohol Then how about this. All browsers without orientation aspects get the auto focus. Others do not.

```
if(typeof window.orientation === 'undefined'){ }
```

andrey1s:
@alcohol f this option with iPad retina doesn't fit, then another option to use [device.js](https://github.com/matthewhudson/device.js) but he adds attributes to `<html>`

alcohol:
I was just pointing it out. I do not know how much of an issue it really is. :-)

Seldaek:
I think @alcohol means iPads won't get the autofocus? Or do you mean they will get wrongly detected because of their high resolution? AFAIK retina uses virtual pixels for all purposes of JS/CSS and not the real device pixels.

Anyway I think it's not a problem, our mobile/tablet usage is very low. It was mostly causing issues on small mobile that had the keyboard popping up, but if it focuses or not on ipad does not matter I think.

PR #553, opened 2015-08-11
Create endpoint for API

timmillwood:
I was speaking to @alcohol on IRC today and he confirmed my thought that there is no way to programatically add packages packagist. So I'm starting the effort towards a PR to add this functionality.

The use case is Drupal. I'm trying to look for a way that as soon as a new module is added to drupal.org it also gets added to packagist under the drupal namespace. I believe @dbuytaert owns the drupal namespace, drupal.org could use his API key to create the packages, then even add the module maintainer as a maintainer on packagist if we store packagist usernames as part of the drupal.org profiles.

We already have an endpoint in the packagist API for updating packages, which is awesome, but many packages are not already there so we need an automated way to add them.

### DIFF
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -82,7 +82,16 @@ class ApiController extends Controller
             return new JsonResponse(array('status' => 'error', 'message' => 'Missing or invalid payload'), 406);
         }
 
-        return $this->receivePost($request, $url, $urlRegex);
+        
+        $package = new Package;
+        $package->setEntityRepository($this->getDoctrine()->getRepository('PackagistWebBundle:Package'));
+        $package->setRouter($this->get('router'));
+        $user = $this->findUser($request);
+        $package->addMaintainer($user);
+        $package->repository = $url;
+        $em = $this->getDoctrine()->getManager();
+        $em->persist($package);
+        $em->flush();
     }
 
     /**
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -82,7 +82,6 @@ class ApiController extends Controller
             return new JsonResponse(array('status' => 'error', 'message' => 'Missing or invalid payload'), 406);
         }
 
-        
         $package = new Package;
         $package->setEntityRepository($this->getDoctrine()->getRepository('PackagistWebBundle:Package'));
         $package->setRouter($this->get('router'));
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -62,7 +62,6 @@ class ApiController extends Controller
      */
     public function createPackageAction(Request $request)
     {
-
         $payload = json_decode($request->getContent(), true);
         if (!$payload) {
             return new JsonResponse(array('status' => 'error', 'message' => 'Missing payload parameter'), 406);
@@ -76,7 +75,7 @@ class ApiController extends Controller
         $package->repository = $url;
         $errors = $this->get('validator')->validate($package)
         if (count($errors) > 0) {
-            foreach( $errors as $error ) {
+            foreach($errors as $error) {
                 $errorArray[$error->getPropertyPath()] =  $error->getMessage();
             }
             return new JsonResponse(array('status' => 'error', 'message' => $errorArray), 406); 
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -57,6 +57,35 @@ class ApiController extends Controller
     }
 
     /**
+     * @Route("/api/create-package", name="generic_create", defaults={"_format" = "json"})
+     * @Method({"POST"})
+     */
+    public function createPackageAction(Request $request)
+    {
+        // parse the payload
+        $payload = json_decode($request->request->get('payload'), true);
+        if (!$payload && $request->headers->get('Content-Type') === 'application/json') {
+            $payload = json_decode($request->getContent(), true);
+        }
+
+        if (!$payload) {
+            return new JsonResponse(array('status' => 'error', 'message' => 'Missing payload parameter'), 406);
+        }
+
+        if (isset($payload['repository']['url'])) { // github/gitlab/anything hook
+            $urlRegex = '{^(?:ssh://git@|https?://|git://|git@)?(?P<host>[a-z0-9.-]+)[:/](?P<path>[\w.-]+/[\w.-]+?)(?:\.git)?$}i';
+            $url = $payload['repository']['url'];
+        } elseif (isset($payload['canon_url']) && isset($payload['repository']['absolute_url'])) { // bitbucket hook
+            $urlRegex = '{^(?:https?://|git://|git@)?(?P<host>bitbucket\.org)[/:](?P<path>[\w.-]+/[\w.-]+?)(\.git)?/?$}i';
+            $url = $payload['canon_url'].$payload['repository']['absolute_url'];
+        } else {
+            return new JsonResponse(array('status' => 'error', 'message' => 'Missing or invalid payload'), 406);
+        }
+
+        return $this->receivePost($request, $url, $urlRegex);
+    }
+
+    /**
      * @Route("/api/update-package", name="generic_postreceive", defaults={"_format" = "json"})
      * @Route("/api/github", name="github_postreceive", defaults={"_format" = "json"})
      * @Route("/api/bitbucket", name="bitbucket_postreceive", defaults={"_format" = "json"})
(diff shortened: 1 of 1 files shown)


timmillwood:
ok, removed all the github/bitbucket stuff, added validation, converted to a PR.

stof:
@timmillwood in general, it is better to create a new PR than converting an existing issue to a PR. It avoids having a mess by mixing the discussion on the feature request with the discussion on the implementation (and it keeps the feature request open in case the proposed implementation gets rejected instead of making the feature request disappear when proposing an implementation btw).

stof:
this is not totally unexpected if the github UI does not provide a way to transform issues into PRs IMO.

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,36 @@ public function packagesAction(Request $req)
...
+     * @Method({"POST"})
+     */
+    public function createPackageAction(Request $request)
+    {
+
+        $payload = json_decode($request->getContent(), true);
+        if (!$payload) {
+            return new JsonResponse(array('status' => 'error', 'message' => 'Missing payload parameter'), 406);
+        }
+        $url = $payload['repository']['url'];
```
stof: do we actually need the ``repository`` part here ? I suggest using ``$payload['url']`` to be simpler

timmillwood:
I think you can tell I'm a Drupal developer ;) not used to github and pull requests. I need to [get off the island](http://www.garfieldtech.com/blog/off-the-island-2013) more.

src/Packagist/WebBundle/Controller/ApiController.php
timmillwood: I was trying to keep the "payload" consistent with the update endpoint, this way I can have a similar function to do create and update but just switch out the endpoint.

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,36 @@ public function packagesAction(Request $req)
...
+        $package->repository = $url;
+        if ($this->get('validator')->validate($package)) {
+            return new JsonResponse(array('status' => 'error', 'message' => 'Invalid package'), 406); 
+        }
+        try {
+            $em = $this->getDoctrine()->getManager();
+            $em->persist($package);
+            $em->flush();
+        } catch (\Exception $e) {
+            $this->get('logger')->crit($e->getMessage(), array('exception', $e));
```
stof: this should return a failure response though

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,36 @@ public function packagesAction(Request $req)
...
+        }
```
stof: you must return a Response from controllers, otherwise you get an exception telling you you messed things

stof:
> I think you can tell I'm a Drupal developer ;) not used to github and pull requests.

you managed to perform an action available only through the API though :smile:

timmillwood:
> you managed to perform an action available only through the API though :smile:

Google + command line

src/Packagist/WebBundle/Controller/ApiController.php
timmillwood: Added

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,39 @@ public function packagesAction(Request $req)
...
+        if ($this->get('validator')->validate($package)) {
+            return new JsonResponse(array('status' => 'error', 'message' => 'Invalid package'), 406); 
+        }
+        try {
+            $em = $this->getDoctrine()->getManager();
+            $em->persist($package);
+            $em->flush();
+        } catch (\Exception $e) {
+            $this->get('logger')->crit($e->getMessage(), array('exception', $e));
+            return new JsonResponse(array('status' => 'error', 'message' => $e->getMessage()), 406);
```
stof: this should be a 500 error actually, and the error message should not be exposed (it is not safe to display it as it can contain sensitive info, especially when coming from the DB)

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,39 @@ public function packagesAction(Request $req)
...
+            return new JsonResponse(array('status' => 'error', 'message' => 'Missing payload parameter'), 406);
+        }
+        $url = $payload['repository']['url'];
+        $package = new Package;
+        $package->setEntityRepository($this->getDoctrine()->getRepository('PackagistWebBundle:Package'));
+        $package->setRouter($this->get('router'));
+        $user = $this->findUser($request);
+        $package->addMaintainer($user);
+        $package->repository = $url;
+        if ($this->get('validator')->validate($package)) {
```
stof: this is wrong. ``validate`` returns a ``ConstraintViolationList`` object, which will always be truthy.

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,39 @@ public function packagesAction(Request $req)
...
+            return new JsonResponse(array('status' => 'error', 'message' => 'Invalid package'), 406);
```
stof: the response should include error messages from violations of the validator to explain what is wrong.

timmillwood: is it safe enough to just do `'message' => $this->get('validator')->validate($package)` (obviously via a variable)?

stof: no, because the ConstraintViolationList does not implement JsonSerializable, and so it would be json-encoded as an empty JS object (there is no public properties)

timmillwood: ok, what would be the suggested way to include the error messages?

timmillwood:
Used a method similar to [the symfony docs](http://symfony.com/doc/current/book/validation.html#using-the-validator-service) to return errors.

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,41 @@ public function packagesAction(Request $req)
...
+        $url = $payload['repository']['url'];
+        $package = new Package;
+        $package->setEntityRepository($this->getDoctrine()->getRepository('PackagistWebBundle:Package'));
+        $package->setRouter($this->get('router'));
+        $user = $this->findUser($request);
+        $package->addMaintainer($user);
+        $package->repository = $url;
+        $errors = $this->get('validator')->validate($package)
+        if (count($errors) > 0) {
+            $errorsString = (string) $errors;
```
stof: casting as string is a bad idea. The symfony doc says it can be handy for debugging. But it is clearly not handy for an API (you have no way to do something with it in your API client, except displaying it as is, and it is not very user-friendly).
You should rather build an array of errors in a format exposing useful info

Crell:
Is the access restriction on the namespace based on the data in the composer.json file, or an account on Packagist itself?  Ideally, *no one* would be able to push to the drupal/ namespace except by going through Drupal.org and the automation there.  I'd prefer to not have 30,000 people who have arbitrary access to push to any drupal/ repository. We got rid of that on Drupal.org years ago...

stof:
@Crell Packagist maintainers are based on packagist accounts only

timmillwood:
I think the current setup will be **ok** for now, but I'm sure *patches welcome*!

alcohol:
Other than some minor PSR2/CS issues, it looks good to me.

dcorb:
Excited to see this. Thanks for working on it Tim

timmillwood:
@alcohol Think I have all the PSR2/CS issues, but please point out the ones I've missed.

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,42 @@ public function packagesAction(Request $req)
...
+        $url = $payload['repository']['url'];
+        $package = new Package;
+        $package->setEntityRepository($this->getDoctrine()->getRepository('PackagistWebBundle:Package'));
+        $package->setRouter($this->get('router'));
+        $user = $this->findUser($request);
+        $package->addMaintainer($user);
+        $package->repository = $url;
+        $errors = $this->get('validator')->validate($package)
+        if (count($errors) > 0) {
+            foreach($errors as $error) {
```
alcohol: space after foreach

alcohol:
Missed one ;-)

alcohol:
Cheers :+1:

Seldaek:
It's live :)

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -57,6 +57,42 @@ public function packagesAction(Request $req)
...
+            return new JsonResponse(array('status' => 'error', 'message' => 'Missing payload parameter'), 406);
+        }
+        $url = $payload['repository']['url'];
+        $package = new Package;
+        $package->setEntityRepository($this->getDoctrine()->getRepository('PackagistWebBundle:Package'));
+        $package->setRouter($this->get('router'));
+        $user = $this->findUser($request);
+        $package->addMaintainer($user);
+        $package->repository = $url;
+        $errors = $this->get('validator')->validate($package)
```
Seldaek: Y'all missed a missing semicolon here btw ;)

@timmillwood you should really make sure you get errors for this in your editor :)

PR #563, opened 2015-08-13
WIP: Use sqlite in dev and test environment

alcohol:
As title suggests.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -5,6 +5,12 @@ framework:
     router:   { resource: "%kernel.root_dir%/config/routing_dev.yml" }
     profiler: { only_exceptions: false }
 
+doctrine:
+    dbal:
+        driver: pdo_sqlite
+        path: %kernel.root_dir%/Resources/sqlite.db
+        charset: UTF8
+
 swiftmailer:
     disable_delivery: true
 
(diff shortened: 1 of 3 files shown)


stof:
-1 on this. Packagist performs a bunch of SQL queries directly, and the supported SQL syntax is not the same in MySQL and Sqlite. Waiting for a prod deployment to discover that your query does not behave the same on MySQL is a bad idea.

alcohol:
I am aware of that, and I am looking into making those queries portable.

stof:
@alcohol you are looking at making *existing* queries portable. but this will not change the fact that *future* queries will always be developed on Sqlite and will run on MySQL only when deploying (because almost everyone runs the dev env locally).
the dev and prod environment should stay as close as possible, and using a different DB platform makes a huge difference.

app/config/config_dev.yml
```diff
@@ -5,6 +5,12 @@ framework:
     router:   { resource: "%kernel.root_dir%/config/routing_dev.yml" }
     profiler: { only_exceptions: false }
 
+doctrine:
+    dbal:
+        driver: pdo_sqlite
+        path: %kernel.root_dir%/Resources/sqlite.db
```
stof: This path is not a writable location, which is very bad for people running Packagist through php-fpm locally rather than using the builtin webserver, as it will mess up permissions.
and btw, it is not an ignored location either, which means people are likely to commit their DB, which is a bad idea

alcohol: It's a work in progress. I'm sorry I did not denote that more clearly. Adding the path to `.gitignore` is trivial. But I'd prefer to determine a suitable location before doing that.

Seldaek:
Yeah meh, not sold on this idea.. Indeed there are quite a few raw SQL queries that have chances of blowing up on sqlite in various ways. If it works optionally cool but I sure won't use it in development.

alcohol:
I can't think of a way to make it optional without making those queries portable. So I guess that puts if off the table :-(

PR #576, opened 2015-09-01
enable precompression of certain files at highest level

alcohol:


src/Packagist/WebBundle/Package/SymlinkDumper.php
```diff
@@ -467,7 +475,17 @@ private function dumpRootFile($file)
...
         }
-        $this->writeFile($file, json_encode($this->rootFile));
+
+        $json = json_encode($this->rootFile);
+        $time = time();
+        $this->writeFile($file, $json, $time);
+
+        $pathgz = $file . '.gz';
+
+        if ($this->compress && !file_exists($pathgz)) {
```
Seldaek: I'd remove the pathgz. We should always recompress if we rewrite the file. Avoids a lot of potential fuckups.

src/Packagist/WebBundle/Package/SymlinkDumper.php
```diff
@@ -480,8 +498,17 @@ private function dumpListing($path)
...
+        $time = time();
+
         if (!file_exists($path)) {
-            $this->writeFile($path, $json);
+            $this->writeFile($path, $json, $time);
+        }
+
+        $pathgz = $path . '.gz';
+
+        if ($this->compress && !file_exists($pathgz)) {
```
Seldaek: Same here remove check

app/config/defaults.yml
```diff
@@ -1,6 +1,7 @@
 parameters:
     packagist_host: ~
     packagist_metadata_dir: "%kernel.cache_dir%/composer-packages-build"
+    packagist_dumper_compress: 0
```
stof: should be a boolean, not an integer

alcohol: Read description of property in class maybe and then let me know if you still feel that way?

stof: ah sorry, I missed the fact that this was not just enabling the compression but also passed to ``gzencode`` itself

PR #586, opened 2015-09-22
Fix deprecated Form::bind method usage

Soullivaneuh:


src/Packagist/WebBundle/Controller/WebController.php
```diff
@@ -48,7 +48,7 @@ public function searchFormAction(Request $req)
         $this->computeSearchQuery($req, $filteredOrderBys);
 
         if ($req->query->has('search_query')) {
-            $form->bind($req);
+            $form->submit($req);
```
alcohol: This is actually also soon to be deprecated.

See https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Form/Form.php#L509

alcohol:
Passing a `Request` instance is going to be deprecated too. Instead I think `handleRequest` could be used in most instances?

Soullivaneuh:
I made this PR to quickfix the deprecation but yes, this could be considered.

stof:
@Soullivaneuh passing a Request object to ``submit()`` is also deprecated. We should still be getting a warning: https://github.com/symfony/symfony/blob/74c24a53152523287cd831cf53476319760884ad/src/Symfony/Component/Form/Form.php#L510

Soullivaneuh:
@stof OK will work on it.

Soullivaneuh:
@alcohol @stof ready for review.

alcohol:
I'm not really able to run a checkout locally, but in so far as I can tell, the changes look good.

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -142,47 +140,44 @@ public function fetchInfoAction(Request $req)
         $user = $this->getUser();
         $package->addMaintainer($user);
 
-        $response = array('status' => 'error', 'reason' => 'No data posted.');
```
stof: this is changing the response when the method is not called as POST. You will need to check for ``$form->isSubmitted()`` for invalid forms to change the response content to keep it

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -142,47 +140,44 @@ public function fetchInfoAction(Request $req)
...
+            /** @var Package $existingPackage */
+            foreach ($existingPackages as $existingPackage) {
+                $similar[] = array(
+                    'name' => $existingPackage->getName(),
+                    'url' => $this->generateUrl('view_package', array('name' => $existingPackage->getName()), true),
+                );
+            }
+
+            $response = array('status' => 'success', 'name' => $package->getName(), 'similar' => $similar);
+        } else {
```
alcohol: As per @stof's comment, I think if you make this an `} elseif ($form->isSubmitted()) {` then you can wrap up with an `} else {` in which you return `$response = array('status' => 'error', 'reason' => 'No data posted.');`.

src/Packagist/WebBundle/Controller/PackageController.php
Soullivaneuh: @stof @alcohol fixed. Thanks for the tip.

Seldaek:
Thanks

Soullivaneuh:
You're welcome. :+1:

Issue #526, opened 2015-06-16
README code highlighting


Seldaek:
Would be nice to have..

PR #590, opened 2015-10-02
added readme highlighting, issue #526

jakoch:
This uses [highlight.js](https://highlightjs.org/) to highlight content in all `pre` and `code` elements inside `div.readme`.

Includes via Cloudflare: https://cdnjs.com/libraries/highlight.js/

### DIFF
--- a/src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
+++ b/src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
@@ -11,6 +11,7 @@
 {% endblock %}
 
 {% block head_additions %}
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/styles/monokai_sublime.min.css">
     <script type="text/javascript">
       WebFontConfig = {
         google: { families: [ 'Source+Code+Pro::latin' ] }
@@ -23,6 +24,31 @@
         var s = document.getElementsByTagName('script')[0];
         s.parentNode.insertBefore(wf, s);
       })();
+      function highlight_code()
+      {
+        if (typeof (Worker) === undefined) return false;
+
+        var workerFunction = new Blob(['(' + highlight_code_worker_function.toString() + ')()'], {type: "text/javascript"});
+        var localWorkerURL = URL.createObjectURL(workerFunction);
+
+        $('div.readme pre, div.readme code').each(function() {
+          var code = $(this);
+          var worker = new Worker(localWorkerURL);
+          worker.onmessage = function(event) { code.html(event.data).addClass('hljs'); }
+          worker.postMessage(code.text()); // start worker
+        });
+      }
+      function highlight_code_worker_function()
+      {
+        onmessage = function(event) {
+          importScripts('https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/highlight.min.js');
+          self.hljs.configure({tabReplace:4});
+          var result = self.hljs.highlightAuto(event.data);
+          postMessage(result.value);
+          close(); // close worker
+        }
+      }
+      $(window).on('load', highlight_code);
     </script>
 {% endblock %}
 


src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -12,6 +12,20 @@
...
+    
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.8.0/styles/monokai_sublime.min.css">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.8.0/highlight.min.js"></script>
+    <script>
+      $(document).ready(function() {
+        hljs.configure({
+          tabReplace: 4,
+          languages: ['apache', 'bash', 'ini', 'javascript', 'json', 'nginx', 'php', 'sql', 'css', 'html']
+        });
+        $('div.readme pre, code').each(function(i, block) {
```
stof: this is looking for ``code`` everywhere, not only in the readme

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -12,6 +12,20 @@
...
 {% block scripts %}
     <script src="{{ asset('bundles/packagistweb/js/view.js')}}"></script>
+    
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.8.0/styles/monokai_sublime.min.css">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.8.0/highlight.min.js"></script>
+    <script>
+      $(document).ready(function() {
+        hljs.configure({
+          tabReplace: 4,
+          languages: ['apache', 'bash', 'ini', 'javascript', 'json', 'nginx', 'php', 'sql', 'css', 'html']
```
stof: should we really restrict the list of languages here ? What is the reason for that ?

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
jakoch: Ok. Changed to `div.readme pre, div.readme code`.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
stof: Btw, this prevents some useful languages (``twig`` for instance)

jakoch: Hmm, i thought that it speeds the highlighting up by reducing the number of languages the auto-detection has to work on. Because it invokes auto-detection on multiple DOM elements for only 10 languages instead of 135 (`console.log(hljs.listLanguages());`).

TomasVotruba: If it's matter of speed, I'd be for restricting.
Adding `twig` and other seems legit.

TomasVotruba:
Thanks for PR. I'm really missing this feature, since the Readme would be much more readable (like on Github). I was about to propose one so thx for saving time :).

TomasVotruba:
What needs to be done here?

alcohol:
Consensus, pretty much. Maybe refine which languages to highlight. Nothing more.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
stof: Well, in most cases, packages are giving an explicit language in their markdown, because it is necessary to have highlighting on github (they don't have auto-detection).
It is too bad that the github markdown renderer does not include the needed class in the rendered HTML to make this info available. This would avoid the need for auto-detection

TomasVotruba: What do you suggest then?

javiereguiluz:
:+1: I like this proposal. However, in the language list:

```
languages: ['apache', 'bash', 'ini', 'javascript', 'json', 'nginx', 'php', 'sql', 'css', 'html']
```

Apart from `twig`, I really miss two very common languages (at least for Symfony packages): `xml` and `yaml`.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
Soullivaneuh: Maybe asynchronously highlight jobs would be a solution for no languages limit.

See: https://github.com/isagalaev/highlight.js/issues/954#issuecomment-150802447

alcohol:
If you could update this to the latest [highlight.js](https://highlightjs.org/) version, I'll go annoy @Seldaek until he merges it :>

jakoch:
- updated HighlightJS to v9.5.0
- added the languages twig, xml, less, scss, yaml as requested by commenters
- the CI failure is [unrelated](https://travis-ci.org/composer/packagist/builds/145497207#L580)
- Note on WebWorker support and answer for Soullivaneuh https://github.com/composer/packagist/pull/590#discussion_r60047040: Yes, a [WebWorker](https://github.com/isagalaev/highlight.js#web-workers) would solve the language detection and rendering freeze problems for large chunks of code. The feature wasn't supported by HighlightJS back in Oct 2015. It landed with [v9.0.0](https://github.com/isagalaev/highlight.js/blob/master/CHANGES.md#version-900). Because this change would affect the requirements, specifically the lower bound for browser compatibility (http://caniuse.com/#feat=webworkers), i'd leave the implementation decision to the maintainers.

Seldaek:
Our user base is 100% devs so we can assume modern browsers. It would be
great if you can enable web workers so people get the best experience
possible. Thanks for the updates!

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -10,24 +10,52 @@
...
-        });
-        $('div.readme pre, div.readme code').each(function(i, block) {
-          hljs.highlightBlock(block);
+    <script type="text/javascript">
+      WebFontConfig = {
+        google: { families: [ 'Source+Code+Pro::latin' ] }
+      };
+      (function() {
+        var wf = document.createElement('script');
+        wf.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1/webfont.js';
```
alcohol: Maybe better to use `//` here instead of `https://`. That way it will use the protocol that the user is currently using to access Packagist, e.g. http or https (if I am not mistaken).

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -10,24 +10,52 @@
...
+          var worker = new Worker(localWorkerURL);
+          worker.onmessage = function(event) { code.html(event.data).addClass('hljs'); }
+          worker.postMessage(code.text()); // start worker
         });
-      });
+      }
+      function highlight_code_worker_function()
+      {
+        onmessage = function(event) {
+          importScripts('https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/highlight.min.js');
```
alcohol: Same as previous note.

alcohol:
Merge conflicts?

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -212,9 +316,6 @@
                 <h1>README</h1>
                 {{ package.readme|raw }}
             </div>
-            <div class="hidden readme-expander">
```
alcohol: Why was the expanded removed? Is the readme no longer partially hidden if it is very  large?

jakoch: Jordi removed it for #607 (show full readme by default) ([changeset](https://github.com/composer/packagist/commit/b7dfeac24e010f27c5b32dc117b1f763bd585122#diff-553142c8b38c82f0b6b1c75c63e9edc3L220)). It's the current state of [`master`](https://github.com/composer/packagist/blob/master/src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig#L291).

jakoch:
The reason for the merge conflict is that i used the file from the master branch, 
because that version provides a `{% block head_additions %}` block and applied my 25 lines on top of it.
Should i open a new PR to have a clear view on the changeset and the option to merge without conflict?

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
stof: @alcohol but Packagist forces using HTTPS anyway (and uses HSTS)

alcohol:
Just rebase it please.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
stof: And always loading the JS file over HTTPS is fine, even when the original site does not use HTTPS btw. I saw posts from Paul Irish saying that he does not recommend using protocol-free URLs anymore in favor of HTTPS one (while he was one of the guy advocating them a few years ago)

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -10,6 +10,48 @@
...
+      function highlight_code()
+      {
+        if (typeof (Worker) === undefined) return false;
+
+        var workerFunction = new Blob(['(' + highlight_code_worker_function.toString() + ')()'], {type: "text/javascript"});
+        var localWorkerURL = URL.createObjectURL(workerFunction);
+
+        $('div.readme pre, div.readme code').each(function() {
+          var code = $(this);
+          var worker = new Worker(localWorkerURL);
```
stof: should you really start a new worker for each separate code block ?

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
alcohol: Fair enough :p

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -10,6 +10,48 @@
     {{ parent() }}
 {% endblock %}
 
+{% block head_additions %}
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/styles/monokai_sublime.min.css">
+    <script type="text/javascript">
+      WebFontConfig = {
```
Soullivaneuh: The indent looks wrong to me.

You start with 4 spaces but use 2 spaces after that.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -212,9 +316,6 @@
                 <h1>README</h1>
                 {{ package.readme|raw }}
             </div>
-            <div class="hidden readme-expander">
-                <i class="glyphicon glyphicon-chevron-down"></i>
-            </div>
         {% endif %}
     </div>
-{% endblock %}
+{% endblock %}
```
Soullivaneuh: Missing new line at end of file.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
alcohol: He mixes 4 (for html) with 2 (for javascript). Personally I'm ok with that.

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
jakoch: Good question. 

You are right, this is a 1 to 1 relation of code block to highlighting worker.
The only downside i noticed is that each worker has to `importScript` the HighlightJS resource.
This leads to multiple additional imports. I'm not sure, why this happens. Maybe my browser settings are set to always no-cache or browsers do not support that(, yet). I would expect that the resource is stored on the first load and loaded from the cache on subsequent requests.

In general, the iteration over code blocks has to go somewhere. 
You either have it on the outside using multiple workers or on the inside of one or two workers.

jakoch: Ok, seems the caching behaviour is my fault. `disable cache (while DevTools) open` was turned on.

stof: I would not iterate over blocks inside workers. this would be hard (as you cannot pass DOM nodes to the worker).

what could be done is to start only 1 worker (when the ``$('div.readme pre, div.readme code')`` collection is not empty), and then  change a bit the worker integration. Instead of posting just the text, you would post a object with the text and an identifier. And ``onmessage`` would receive an object with 2 properties too (the highlighted HTML and the same identifier) and use the identifier to associate it to the right DOM node.
This way, you would keep a single worker, which would handle all messages (you should stop closing the worker at the end of the highlighting of course)

jakoch:
- Ok, i've rebased it against master. It's now only a few lines.
- For the remarks regarding indentation: it's your master branch.
- You might find a standalone example with some randomly picked readme's with code blocks of different languages [here](https://gist.github.com/jakoch/12c6e6ee015de4164ce7f14bb2c96332/archive/92a09803ecbb66220a21bfbe57b2d7c6cd39e61d.zip).
- A problem i found: The automatic language detection is sometimes off, when code blocks are really short.
There is simply not enough semantic information available to infer the language correctly.
I think it's not possible (for any highlighter) to solve this without providing an additional language annotation.
At the same time developers usually don't annotate their code-fences. 
One solution to avoid highlighting in a wrong language would be to exclude really short code blocks (string length / one-liner check). Another would be to use a fixed language, e.g. 'bash', for short code blocks. What's your opinion?

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
stof: btw, the identifier can be the index in the collection, which will make it easy

src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
```diff
@@ -23,6 +24,31 @@
...
+      {
+        onmessage = function(event) {
+          importScripts('https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/highlight.min.js');
+          self.hljs.configure({tabReplace:4});
+          var result = self.hljs.highlightAuto(event.data);
+          postMessage(result.value);
+          close(); // close worker
+        }
+      }
+      $(window).on('load', highlight_code);
```
stof: should be ```$(highlight_code)`` instead

stof:
@jakoch ideally, it would be great that github could apply the class name in their rendered readme when the fenced code block defines the class. Unfortunately, they don't do it AFAIK (or the HTMLPurifier setup used when receiving the readme drops them, which would be an issue on the Packagist side then).
In such case, the page should provide the language in the message sent to the worker (using an object with multiple properties, as I suggested previously, will make it easier to make such change in the message)

Seldaek:
Thanks :)

PR #591, opened 2015-10-03
Show the package type in the package view

### LINKED ISSUE #382: Show package type
It should be at least visible on the detail page of a package which type it is, eg. `library`, `yii2-extension`, ...

Maybe next to the tags in a different color.

venyii:
Fixes #382 

This will show the package type below the `Language:` info.

screen shot 2015-10-03 at 18 42 37


Seldaek:
Thanks

PR #592, opened 2015-10-03
Add a warning in the package view if no license was found for the current version

### LINKED ISSUE #577: The warning about license being unknown should be displayed in a prominent place
Currently, the message about license being unknown is low in the page (even off-screen when you have lots of dependencies).
And even if it is displayed in red, this is not very visible in the orange UI
unknown_license

It would be great to make it more visible. I see 2 solutions:

- move the licensing info higher in the page for everyone
- add a warning at the top of the package page for maintainers in case it has no license, similar to what we do when the package does not have the github hook to auto-update it.

venyii:
Fixes #577 

This PR checks if the `expandedVersion` has a license, and if not, displays a warning.

I'm not sure about the exact wording, any wishes/ideas?

screen shot 2015-10-03 at 19 02 12


Seldaek:
Thanks

PR #593, opened 2015-10-04
Add page to show the API documentation

cloudson:
To solve some questions about the packagist api, we should
have a page with a simple documentation about it.
We can discuss about the layout or improvements on that page.


alcohol:
I would prefer to see something based on swagger for example.

Also you demonstrate `https://packagist.org/packages/list.json?vendor=silex` as one of your first examples, but then proceed to list all `symfony/X` components. I think the result does not quite match the input given there ;-)

cloudson:
I think the result does not quite match the input given there ;-)
You are right, I didnt show all the response json because it's so large hehe. 
I didn't know how to show that informations on the page of a clear way. :(

alcohol:
I would include the root keys only, and use dummy values. No need to show actual responses. These might change over time.

Also I was referring to the following:
```
<pre>
GET https://packagist.org/packages/list.json?vendor=silex
<code>
{
  "packageNames": [
    "symfony/asset",
    "symfony/assetic-bundle",
    "symfony/bootcamp-bundle",
    "symfony/browser-kit",
    "symfony/class-loader",
    "symfony/config",
    "symfony/console",
    "symfony/css-selector",
    "symfony/debug",
    "symfony/debug-bundle",
    "symfony/dependency-injection",
    "symfony/doctrine-bridge",
    "symfony/dom-crawler",
    "symfony/event-dispatcher",
    "symfony/expression-language",
    "symfony/filesystem",
    "symfony/finder",
    "symfony/form",
    "symfony/framework-bundle",
    "symfony/framework-standard-edition",
    "symfony/http-foundation",
    "symfony/http-kernel" 
  ]
}
</code></pre>
```

Silex is not the vendor of any of those symfony packages...

cloudson:
humm, ok @alcohol . I will rewrite this :)

stof:
@cloudson you have a broken HTML at the beginning of the APIdoc template:

```html
<p>Packagist has a public API to consume data. See bellow the most important 
URIs: </>
```

and just after than, you are wrapping a list inside a ``<p>`` tag, which is invalid HTML (browsers will close the ``<p>`` tag before the list, making it empty, and so useless)

you also have a bunch of unescaped ``&`` and ``<`` in your HTML (in code snippets).

and I agree with @alcohol about the fact that showing all keys for all versions in the output of the package JSON view makes the page much harder to read. Especially given that most of these keys are just the data defined by the composer.json schema, and so could be linked to the dedicated composer doc to explain them. This would make the page much easier to read.

Seldaek:
Thanks!

PR #599, opened 2015-10-22
Allow (optional) port number in repository URL

kornrunner:
This allows us to create webhooks for URLs that have port numbers in them (specifically `ssh://` running under custom port).

I've added the test as well, but haven't run them, haven't set it up properly yet.

### DIFF
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -111,7 +111,7 @@ class ApiController extends Controller
         }
 
         if (isset($payload['repository']['url'])) { // github/gitlab/anything hook
-            $urlRegex = '{^(?:ssh://git@|https?://|git://|git@)?(?P<host>[a-z0-9.-]+)[:/](?P<path>[\w.-]+/[\w.-]+?)(?:\.git)?$}i';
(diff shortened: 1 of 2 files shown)


src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -111,7 +111,7 @@ public function updatePackageAction(Request $request)
         }
 
         if (isset($payload['repository']['url'])) { // github/gitlab/anything hook
-            $urlRegex = '{^(?:ssh://git@|https?://|git://|git@)?(?P<host>[a-z0-9.-]+)[:/](?P<path>[\w.-]+/[\w.-]+?)(?:\.git)?$}i';
+            $urlRegex = '{^(?:ssh://git@|https?://|git://|git@)?(?P<host>[a-z0-9.-]+):(?:[0-9]+)?/(?P<path>[\w.-]+/[\w.-]+?)(?:\.git)?$}i';
```
stof: this change breaks all URLs which don't have the port in them

stof:
@Seldaek why don't we have Travis on the repo to run the testsuite ?

Seldaek:
Just never bothered to set it up. But yeah that change is broken, you can run the tests without setting up the whole project though. At least if you use phpunit --filter to only run the tests you modified, this doesn't depend on anything being setup.

kornrunner:
Sorry guys, it's my bad either way. I tried to run It should be ok now.

Travis would be great, I agree.

Seldaek:
Thanks for updating, looks better now :)

kornrunner:
Also, this one is somewhat related (though in composer, actually): https://github.com/composer/composer/pull/4530.
It'd be great if you could check it out. Thank you!

PR #611, opened 2015-12-02
Show deps count on dependents page

Soullivaneuh:
Just added a counter on dependents page.

I also made a quick-fix on CSS for package list that hide the title line.

Before:

selection_553

After:

selection_555


alcohol:
LGTM

Seldaek:
Thanks

PR #616, opened 2015-12-11
Add package aside icons

Soullivaneuh:
This is a little design trick proposal to add icons on the package aside:

selection_559

What do you think?

I also added link for GitHub's open issues, stars and forks.


src/Packagist/WebBundle/Resources/public/css/main.css
```diff
@@ -1003,10 +1003,10 @@ ul.packages .metadata-block:last-child {
 .package-aside .details {
   padding-left: 0;
 }
-.package .package-aside i, .package .readme-expander i {
+.package .package-aside i:not(.fa), .package .readme-expander i:not(.fa) {
```
stof: using ``.glyphicon`` instead of ``i:not(.fa)`` may be better

Soullivaneuh: I'm not using Glyphicon but FontAwesome that provide much icons to feed my needs: http://fontawesome.io/icons/

BTW, using `.glyphicon` on the view with produce the same gray color issue.

stof: well, these CSS rules are explicitly excluding your FA icons, which is why I suggest making it target the other icons explicitly instead

Soullivaneuh: Not sure of what you mean. This?

```css
i:not(.fa):not(.glyphicon)
```

Or maybe you want me to add a `i.fa` rule to rollback to the default color? It's a little bit to tricky IMHO.

Seldaek:
It was actually a conscious decision not to put icons there to keep it lighter. I know it's doable but it doesn't mean it has to be done IMO.

Soullivaneuh:
> It was actually a conscious decision not to put icons there to keep it lighter.

Well, it's just a proposition. If you think it's not necessary or needed, just close this PR. ;-)

Seldaek:
Yes I just didn't want to close it directly without leaving the door open for discussion, but my opinion is that it's not necessary :)

Soullivaneuh:
Yeah, it's just a matter of tastes.

Do you want to keep it open waiting another opinions?

src/Packagist/WebBundle/Resources/public/css/main.css
stof: I meant using ``.glyphicon`` *instead of* ``i:not(.fa)``

alcohol:
Maybe not the icons, but the extra links could be a welcome addition?

src/Packagist/WebBundle/Resources/public/css/main.css
Soullivaneuh: Oh, ok I got it (finally).

That's indeed better. But will not works if one time, some people want to use a gliphycon without the gray color.

Soullivaneuh:
> Maybe not the icons, but the extra links could be a welcome addition?

Yes. If you're OK for that, I'll make another PR.

Soullivaneuh:
@Seldaek I saw you closed the PR. Are you at least interested by extra links like @alcohol suggested?

Seldaek:
Sure why not link I suppose.

Issue #595, opened 2015-10-15
Site Search ( or composer search ) does not search


cedamorim:
When I try to do any research, it is always returned the same result (all records).

rchouinard:
I'm experiencing the same. Looks like the changes to WebContoller in 46bc2d4 might have broken it.

chlabelek:
Same. When typing, the search query is "q" instead of "tags".

alcohol:
Cannot reproduce. Searching works just fine for me :\

chlabelek:
Fixed indeed.

PR #618, opened 2015-12-11
Deps update and fix deprecated stuff

Soullivaneuh:
I locked Symfony version to 2.7 because 2.8 make Packagist website failing.

- [x] Deps update
- [x] Fix not up to date deps (sensio packages and stuff)
- [x] Intefrate fos-user-bundle 2.x to get rid of deprecated warnings


composer.json
```diff
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": ">=5.5",
-        "symfony/symfony": "^2.7",
+        "symfony/symfony": "2.7.*",
```
alcohol: :cry:

src/Packagist/WebBundle/Resources/config/services.yml
```diff
@@ -26,14 +26,12 @@ services:
 
     packagist.user_repository:
         class: Packagist\WebBundle\Entity\UserRepository
-        factory_service: doctrine
-        factory_method: getRepository
+        factory: [@doctrine, getRepository]
```
alcohol: Might as well go ahead and quote all `@<something>` references (in preparation for SF 2.8/3.x).

src/Packagist/WebBundle/Resources/config/services.yml
```diff
@@ -98,8 +95,7 @@ services:
 
     packagist.menu.user:
         class: Knp\Menu\MenuItem
-        factory_service: packagist.menu_builder
-        factory_method: createUserMenu
+        factory: [@packagist.menu_builder, createUserMenu]
```
stof: Please quotes the string starting with ``@`` to be valid Yaml (see the 2.8 deprecations)

composer.json
Soullivaneuh: Just try phpunit with SF 2.8! :smiley: 

It seems some BC breaks are here finally. :-/

src/Packagist/WebBundle/Resources/config/services.yml
Soullivaneuh: Why? This is needed only for string that does not represent a service AFAIK. Am I wrong?

alcohol: No, the `@` is a reserved character now (because -reasons-, don't ask me, I really don't know). So you should always quote the identifiers that use it :)

stof: @alcohol ``@`` is reserved in the YAML spec. This is not new. But the Symfony parser was accepting it, so not following the spec.

@Soullivaneuh the Yaml parser has no concept of services. The YamlFileLoader expects a string starting with ``@`` to reference a service

alcohol: > The “@” (#x40, at) and “`” (#x60, grave accent) are reserved for future use.

Soullivaneuh: Oh ok then, thanks for clarification!

composer.json
stof: which BC breaks ?

Soullivaneuh: I didn't dig into it for now. PHPUnit get 500 errors on web tests, not with 2.7.

I would like to fix deprecated stuff before, then try to look at 2.8.

Soullivaneuh:
I don't know how to fix this one: https://travis-ci.org/composer/packagist/jobs/96286044#L381

I didn't find any `_method` requirements on the project.

composer.json
stof: are they failure because of deprecations or because of something else ?

src/Packagist/WebBundle/EventListener/RegistrationListener.php
```diff
@@ -0,0 +1,46 @@
...
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            FOSUserEvents::REGISTRATION_SUCCESS => 'onRegistrationSuccess'
+        ];
+    }
+
+    public function onRegistrationSuccess(FilterUserResponseEvent $event)
```
stof: wrong event class

Soullivaneuh:
Weird, I have deprecation notices that I don't have on my local.

stof:
@Soullivaneuh clear your cache to force Twig to recompile the templates :smile:

stof:
I discovered that Packagist contains a bunch of useless ZF components. I will wait until this is merged before sending a PR to avoid conflicts on the composer.lock file

Soullivaneuh:
> @Soullivaneuh clear your cache to force Twig to recompile the templates :smile: 

:smile:

stof:
@Soullivaneuh it might also be great to update to NelmioSolariumBundle 2, to use a maintained version of Solarium (but may be done separately)

Soullivaneuh:
Well, this is not the only one:

```bash
8 packages are not up to date:

  - solarium/solarium (2.4.1) latest is 3.4.1
  - nelmio/solarium-bundle (v1.1.0) latest is v2.2.0
  - sensio/distribution-bundle (v2.3.22) latest is v5.0.2
  - sensio/framework-extra-bundle (v2.3.4) latest is v3.0.11
  - sensio/generator-bundle (v2.5.3) latest is v3.0.1
  - zendframework/zend-hydrator (1.0.0) latest is 2.0.0
  - doctrine/common (v2.5.2) latest is v2.6.0
  - symfony/symfony (v2.7.7) latest is v3.0.0
```

But let's start to make it working without deprecation before! :smile:

stof:
@Soullivaneuh for ``_method`` warnings, it is because of the ``@Method`` annotation. Fixing it depends on migrating to SensioFrameworkExtraBundle 3 (the version 2 is not maintained anymore anyway)

Soullivaneuh:
Wow, that's a tricky issue, thanks for the tip.

stof:
well, SFEB 2.x was never updated to use the new API, as it was EOLed before the 2.7 release actually

stof:
@Soullivaneuh next steps: updating the packagist twig extension, and removing usage of ``form_enctype`` in templates. then we'll be almost good

Seldaek:
Please don't touch solarium because AFAIK latest solarium requires a higher version of solr, so that'll block the PR until I can upgrade solr on the server.. which is an entirely different matter. Rather do this separately.

stof:
@Seldaek I totally agree it should be done separately.
Regarding the Solr version, Solarium supports Solr 1.4+, but updating a field to set ``null`` in it works only for Solr 4+. For older versions, the field should be removed instead of set to null but I don't know whether this impacts Packagist

Soullivaneuh:
Well, I'm totally not a Solr expert so I will let this work for someone else. ;-)

Soullivaneuh:
All SF 2.7 and Twig deprecations should be resolved now.

@stof I will allow SF 2.8 right now so you'll be able to see test errors.

Soullivaneuh:
Well, no errors now. Seems to be resolved with SFEB update.

Now, I have more deprecated stuff to fix. :-)

Soullivaneuh:
Not sur this error is related: https://travis-ci.org/composer/packagist/jobs/96566766#L424

Travis issue?

Soullivaneuh:
Deprecated stuff fixed, commits squashed and rebased.

PR ready for review and merge.

app/autoload.php
```diff
@@ -2,6 +2,8 @@
 
 use Doctrine\Common\Annotations\AnnotationRegistry;
 
+error_reporting(error_reporting() & ~E_USER_DEPRECATED);
```
alcohol: Please do not commit this.

src/Packagist/WebBundle/Controller/WebController.php
```diff
@@ -38,16 +39,17 @@ public function indexAction()
...
 
         $filteredOrderBys = $this->getFilteredOrderedBys($req);
         $normalizedOrderBys = $this->getNormalizedOrderBys($filteredOrderBys);
 
         $this->computeSearchQuery($req, $filteredOrderBys);
 
-        if ($req->query->has('search_query')) {
-            $form->submit($req);
-        }
+        $form->handleRequest($req);
```
alcohol: I think the last time we tried to change this we had to specifically revert it because the conditional check was necessary for the search page to function properly.

src/Packagist/WebBundle/Controller/WebController.php
```diff
@@ -107,18 +109,16 @@ public function searchAction(Request $req)
                 $select->addSorts($normalizedOrderBys);
             }
 
-            if ($req->query->has('search_query')) {
-                $form->submit($req);
+            $form->handleRequest($req);
```
alcohol: Same comment as above.

src/Packagist/WebBundle/Resources/config/services.yml
```diff
@@ -1,7 +1,7 @@
 services:
     packagist.twig.extension:
         class: Packagist\WebBundle\Twig\PackagistExtension
-        arguments: [ @doctrine ]
+        arguments: [ '@doctrine' ]
```
alcohol: Not sure why you would mix using single `'` and double `"` quotes. I would prefer to see consistency in this file :-)

web/app.php
```diff
@@ -5,7 +5,11 @@
 use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = include __DIR__.'/../app/bootstrap.php.cache';
+/**
+ * @var \Symfony\Component\ClassLoader\ClassLoader
+ */
+$loader = include __DIR__.'/../app/autoload.php';
```
alcohol: I think these should be `require[_once]`, not `include`.

alcohol:
Is `src/Packagist/WebBundle/EventListener/RegistrationListener.php` autoloaded based on directory naming conventions or something? I'm used to just registering listeners through configuration, so I don't know much about any automagic conventions regarding directory structures.

app/autoload.php
Soullivaneuh: This is a stupid copy paste from Symfony standard edition: https://github.com/symfony/symfony-standard/blob/2.8/app/autoload.php#L6

Do you still want to remove this?

src/Packagist/WebBundle/Controller/WebController.php
Soullivaneuh: What was the issue?

src/Packagist/WebBundle/Resources/config/services.yml
Soullivaneuh: I'm used to use single quotes if double are not necessary, as PHP rules.

Do you want me to put single quote everywhere or to move back to double quotes?

alcohol: I don't really care, just use the same consistently.

app/autoload.php
alcohol: Yea :)

src/Packagist/WebBundle/Resources/config/services.yml
Soullivaneuh: Done at https://github.com/composer/packagist/commit/fe9e8a60015c15b60256de9e8ef73e0da679b42f.

src/Packagist/WebBundle/Controller/WebController.php
Soullivaneuh: Ok found it: https://github.com/composer/packagist/issues/595

Will take a look.

Seldaek: The problem AFAIK is that form->handleRequest isn't very smart and it just accepts the form as submitted and valid even if there is nothing submitted in this case, maybe because our whole form is optional, not sure, but it's not really a big deal to use the form->submit() manually IMO, we don't have to rely on the magic everywhere. It's not deprecated or anything AFAIK.

Soullivaneuh: The fact is `submit` is not deprecated but passing the whole request to `submit` is.

I will take a look.

app/autoload.php
stof: @Soullivaneuh there is a PR on symfony-standard to remove it. This was added in the 2.7.0 release when our deprecation warnings were not silenced.

web/app.php
stof: should not be _once, otherwise you are not guaranteed to get the loader

app/autoload.php
```diff
@@ -2,6 +2,8 @@
...
+
 if (!$loader = @include __DIR__.'/../vendor/autoload.php') {
```
stof: I suggest either changing this to ``require``, as done in Symfony, or at least ensuring that you get a failure exit code (i.e. displayign the message and then using ``exit(1)``) if you want to keep the nicer error message

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -1015,7 +1021,7 @@ private function createAddMaintainerForm(Package $package)
 
         if ($this->isGranted('ROLE_EDIT_PACKAGES') || $package->getMaintainers()->contains($user)) {
             $maintainerRequest = new MaintainerRequest();
-            return $this->createForm(new AddMaintainerRequestType(), $maintainerRequest);
+            return $this->createForm(AddMaintainerRequestType::class, $maintainerRequest, ['method' => 'POST']);
```
stof: POST is the default method, so this can be omitted

src/Packagist/WebBundle/Security/Provider/UserProvider.php
```diff
@@ -29,11 +30,18 @@ class UserProvider implements OAuthAwareUserProviderInterface, UserProviderInter
...
+     * @var EmailUserProvider
+     */
+    private $userProvider;
+
+    /**
      * @param UserManagerInterface $userManager
+     * @param EmailUserProvider    $userProvider
      */
-    public function __construct(UserManagerInterface $userManager)
+    public function __construct(UserManagerInterface $userManager, EmailUserProvider $userProvider)
```
stof: you should typehint the interface here, as you only use the interface

src/Packagist/WebBundle/Security/Provider/UserProvider.php
```diff
@@ -88,7 +96,7 @@ public function loadUserByOAuthUserResponse(UserResponseInterface $response)
      */
     public function loadUserByUsername($usernameOrEmail)
     {
-        $user = $this->userManager->findUserByUsernameOrEmail($usernameOrEmail);
+        $user = $this->userProvider->loadUserByUsername($usernameOrEmail);
 
         if (!$user) {
             throw new UsernameNotFoundException(sprintf('No user with name or email "%s" was found.', $usernameOrEmail));
```
stof: this can be removed now, as the user provider already handles it

app/autoload.php
Soullivaneuh: Ok then.

Seldaek:
Thanks, just had to fix a couple minor things (see commits above), mostly to avoid dev dependencies as they tend to get hard to remove after a while when you have no idea anymore why they were there.

Speaking of which, I am not sure why you upgaded snc redis bundle to dev-master from 1.1.9, I downgraded it back and it seems to work fine so I'll keep it as is for now but let me know if there was a good reason to upgrade.

Seldaek:
Oh I see it breaks the test suite with 1.1.9 because of a deprecated PSR-3 method. I guess I'll live with that for now and nag snc for a new tag because relying on master really sucks.

PR #619, opened 2015-12-11
Clean the code of controllers

stof:
- avoid using deprecation logger methods on failure
- remove dead code (unused Template annotation, unused argument...)
- fix inline phpdoc helping IDE, to actually help them


src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -458,7 +454,7 @@ public function viewPackageDownloadsAction(Request $req, $name)
      * )
      * @Method({"GET"})
      */
-    public function viewPackageVersionAction(Request $req, $versionId)
+    public function viewPackageVersionAction( $versionId)
```
Seldaek: stray space here

stof: fixed

Seldaek:
Thanks

PR #620, opened 2015-12-11
Remove class from ending html tag

Soullivaneuh:
Because, seriously @Seldaek?! :wink:


stof:
:+1:

Seldaek:
That's called fucking up due to over-use of multiple cursors in sublime text :p

Soullivaneuh:
> That's called fucking up due to over-use of multiple cursors in sublime text :p

http://www.excusesdedev.com/ :-p

PR #621, opened 2015-12-11
Avoid creating a useless prototype for the search ordering form

stof:
There is no JS code reading the prototype to add elements in the collection. The UI hides the fields and builds links separately.


Seldaek:
No idea what this option even does (I barely use forms..) so I'm just gonna trust you on this one.

stof:
@Seldaek this option (which is enabled by default) triggers (when ``allow_add`` is also enabled) the creation of a ``data-prototype`` attribute on the collection with the HTML necessary to add new elements in the collections client-side before submitting the form.
this is useful in most cases dealing with a collection form, but the ordering UI works differently here, and the collection is always hidden. So there is no need to generate this prototype

Seldaek:
Ah ok I see, thanks.

PR #625, opened 2015-12-23
Update value of static variable so Redis command is only defined once

SHyx0rmZ:


SHyx0rmZ:
I guess that's what I get for not running the tests locally (not at my development machine right now). It's also a rather ugly problem so I'll close this pull request now and create a new one when I have a better solution at hand.

alcohol:
The tests for Packagist are broken at the moment.

Seldaek:
Thanks for the fix, I merged it and fixed it so the tests pass.

PR #641, opened 2016-02-08
Better monospaced font for box drawing

Gisleburt:
Consider this the lowest of the low in priority. :smile: 

Most of the current fonts used in `<pre>` do not have monospaced support [box drawing characters](https://en.wikipedia.org/wiki/Box-drawing_character)

This can lead to deformities for people who try to be too smart for their own good, like myself.

Example:
[https://packagist.org/packages/foundry/masonry](https://packagist.org/packages/foundry/masonry)

The fonts in play are.

In [`pre, code`](https://github.com/composer/packagist/blob/d8ccd79956b4bc7d7d27de00419341c28c36b64b/src/Packagist/WebBundle/Resources/public/css/main.css#L30-L36):
 * Consolas :x:
 * Courier New :x:
 * Courier :x:
 * monospace :x:

In [`code`](https://github.com/composer/packagist/blob/master/web/libs/bootstrap-3.3.5/css/bootstrap.min.css) (imported from bootstrap):
 * Menlo  :white_check_mark:
 * Monaco :x:
 * Consolas :x:
 * Courier New :x:
 * monospace :x:

This pull request adds Menlo to the front of the `pre, code`, which is a quick fix and to me, hardly noticable in style.

### DIFF
--- a/src/Packagist/WebBundle/Resources/public/css/main.css
+++ b/src/Packagist/WebBundle/Resources/public/css/main.css
@@ -29,7 +29,7 @@ html, body {
 
 pre, code {
     background: #fff;
-    font-family: Consolas, "Courier New", Courier, monospace;
+    font-family: Menlo, Consolas, "Courier New", Courier, monospace;
     border-radius: 0;
     border: none;
     color: #2d2d32;


Gisleburt:
Side note: I personally find [Source Code Pro](https://github.com/adobe-fonts/source-code-pro) to be the best all round monospaced font for programming (comes with PhpStorm though is weridly not the default), and it gets the line width right on the box drawing too, which Menlo doesn't quite.

Gisleburt:
Tests are broken:
```
Remaining deprecation notices (1)

The Symfony\Bridge\Monolog\Logger::err method inherited from the Symfony\Component\HttpKernel\Log\LoggerInterface interface is deprecated since version 2.2 and will be removed in 3.0. Use the error() method instead, which is PSR-3 compatible: 1x

    1x in WebControllerTest::testSearchNoOrderBysAction from Packagist\WebBundle\Tests\Controller
```

Gisleburt:
I couldn't get unit tests to work on my local copy, nor could I find any calls to `Symfony\Bridge\Monolog\Logger::err` so I can't help on this one, it's beyond my expertise.

Seldaek:
I guess I can merge, but on windows/linux this has no effect as Menlo is only on OSX Snow Leopard and up. That's sad but so it goes.

Gisleburt:
@Seldaek Source Code Pro is available as a Google Font, it seems like a better alternative and I'd be happy to make the PR if you think it's worth it. :)

https://www.google.com/fonts#UsePlace:use/Collection:Source+Code+Pro

Can be added with:

```css
@import url(https://fonts.googleapis.com/css?family=Source+Code+Pro);
```

so only the one file needs to change.

Seldaek:
I did the switch..

image

image


As you can see it is still off on windows, problem is chars like `╭` and `→` are not the same width as the others, even though it's a monospace font. I'm not sure what does this, but this breaks the whole rendering :/

FWIW it also fails to render correctly on github.

Gisleburt:
That's disapointing. Front end is hard.  :stuck_out_tongue:

Thank you for spending so much time on this though, it's greatly appreciated. :relaxed:

PR #646, opened 2016-02-25
Add php 5.5 to travis ci

paladox:
Reason since that is the minium supported php version for this repo we should test against php 5.5 too.


Seldaek:
We don't explicitly support 5.5 actually, that was a mistake sorry. Server runs on 5.6 and will move to 7 eventually.

PR #648, opened 2016-03-01
Updated the way numbers are formatted

javiereguiluz:
This PR proposes two changes:

**1)** The thousands separator (`&thinsp;`) is almost invisible. In my opinion this makes large numbers very hard to understand:

| Before | After
| --- | ---
| before | after

**2)** There is no need to do the number formatting per number because Symfony allows to configure this globally. I've only updated a template fragment, but if this change is accepted, I can update the rest of the templates.

---

Note: I haven't tested these changes locally because I can't install Packagist application yet because of its requirements.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -32,6 +32,10 @@ twig:
         - 'PackagistWebBundle::forms.html.twig'
     debug:            %kernel.debug%
     strict_variables: %kernel.debug%
+    number_format:
+        decimals: 0
+        decimal_point: '.'
+        thousands_separator: ','
     globals:
         google_analytics: %google_analytics%
         packagist_host: %packagist_host%
--- a/src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
+++ b/src/Packagist/WebBundle/Resources/views/Package/viewPackage.html.twig
@@ -149,12 +149,12 @@
                         </div>
 
                         <div class="facts col-xs-12 col-sm-6 col-md-12">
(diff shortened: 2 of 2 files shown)


Seldaek:
Just for reference, they show just fine here..

image

stof:
it may depend on the font being used

Seldaek:
It's "Open Sans" from google fonts across the board, so I'd think it should render same cross-platform. Except the world is broken.

alcohol:
This is kind of locale specific I think? For example, we (Dutch) use quite different formats depending on context (to make things easier), as do many other countries: https://en.wikipedia.org/wiki/Decimal_mark#Examples_of_use

Not sure what the "right" approach would be. Maybe just use a normal space instead of `&thinsp;`?

javiereguiluz:
@alcohol yes, you are right. But Packagist is an English-only website, so maybe we should only care for English locale for now.

alcohol:
Well... English is not a locale, it's a language. :p

javiereguiluz:
@alcohol you are right again!

Going back to the issue, maybe it's just a Mac OS problem? In Firefox and Safari is slightly better than in Chrome, but still could be improved:

| Chrome | Firefox | Safari
| --- | --- | ---
| chrome | firefox | safari

javiereguiluz:
I'm closing it because this issue magically disappeared. On Chrome browser I can now see the separation between numbers:

packagist_numbers

Thank you all for the discussion.

PR #650, opened 2016-03-08
Make possible next runs of DumpPackagesCommand after SymlinkDumper's exceptions thrown

fduch:
Now if DumpPackagesCommand fails due to any kind of exception thrown inside SymlinkDumper (it can be caused by filesystem problems or mysql server connectivity issues for example), - the lock aquired before SymlinkDumper is executed stays unreleased. 
This causes impossibility of aquiring this lock during the next DumpPackagesCommand run. 

This patch releases lock regardless of whether dump process was successful or not.

### DIFF
--- a/src/Packagist/WebBundle/Command/DumpPackagesCommand.php
+++ b/src/Packagist/WebBundle/Command/DumpPackagesCommand.php
@@ -86,8 +86,11 @@ class DumpPackagesCommand extends ContainerAwareCommand
             return;
         }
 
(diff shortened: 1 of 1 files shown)


fduch:
Oops, checks failed. But the reason is not related to suggested patch :)

Seldaek:
It's usually not really necessary as the lock gets released when the php process dies, but why not I suppose. It doesn't do any harm.

Issue #588, opened 2015-09-30
[Question] Does packagist have an API?


samayo:
I would like to know, if there is any way to find out if certain package exists using cURL. An example of request would be 
`http://api.packagist.org/package/samayo/bulletproof`
Which would result in a JSON format of the package's details (exists, date, download count .. ) if, not the JSON would return 404. 

I searched for this and could not find anything except for [KnpLabs/packagist-api](https://github.com/KnpLabs/packagist-api) which unfortunately, is a PHP-only solution.

alcohol:
The API methods it has are located in the following file:

https://github.com/composer/packagist/blob/master/src/Packagist/WebBundle/Controller/ApiController.php

Unfortunately, it seems "error" responses, such as a package not found, will still return a '200 OK' header actually. The response body will contain a JSON object which has a status propery that will have the value of "error".

@Seldaek is this where you learned your lesson about how **not** to do it? :P

alcohol:
Aside from that, no, Packagist does not seem to offer any methods that match your use-case.

stof:
https://packagist.org/packages/samayo/bulletproof.json does what you want

@alcohol the file you linked contains webhooks, for which a 200 is indeed returned to avoid the webhook call being retried.

alcohol:
Ah ok, fair enough. I still think returning error codes is better though. Any proper service wouldn't retry an API if it received a 400 or 404 for example.

stof:
Well, you should first check how the different webhook systems (github, bitbucket, gitlab, etc...) are behaving regarding retry.

samayo:
@stofChristopheCoevoet Thanks. That is what I was looking for. One last question, is there a rate limit as to how many times per minute we can make a request to the server?

alcohol:
Hopefully they properly support 503 in that regard :-)

alcohol:
@samayo no limits are artificially imposed. But please be considerate.

laudeon:
Hello,

I would like to know if there is any way to call packagist API without server-side language ? 
Actually, JSONP is not allow and CORS too. And I found only PHP client.

Kind regards!

alcohol:
@Seldaek do you have any desire or intent to allow CORS on the API?

Seldaek:
I don't have any desire or intent but I'm not really against it I guess..
We could add the NelmioCorsBundle and set it up for the few read-only api
urls.

samayo:
[...]

PR #652, opened 2016-03-16
Enabled CORS for API URLs

javiereguiluz:
As discussed in #588.


stof:
you are misisng the update of the AppKernel to rgister the bundle

javiereguiluz:
Updated. Thanks!

Seldaek:
Thanks, note that the regexes don't need to escape `/`s (fixed in commit above) :)

javiereguiluz:
Thanks for merging and sorry for the regexp error.

Seldaek:
Well no biggie they were not errors per se just unneeded/superfluous 
escaping so I thought I'd let you know :)

PR #659, opened 2016-03-31
Add HTML5 autofocus to search box

hobofan:
Adds a HTML5 autofocus attribute to the search box. This allows the autofocus to work without JS enabled on supported browsers.

I recently added the same thing to crates.io, the main Rust package repository: https://github.com/rust-lang/crates.io/pull/293


Seldaek:
We solved it differently in https://github.com/composer/packagist/commit/5f431c77e1f55308f591400abf93f9590ae05c1d  because it was causing issues for some people on mobile (the touch keyboard would pop up in some mobile browsers, which is really not helpful there).

Seldaek:
I am curious why you sent this though, did you not see the autofocus behavior? Maybe a too small window? Or too small due to dev tools being open?

hobofan:
Mostly because I think that HTML5 should be used over javascript when possible, since it usually leads to less complexity, and the behaviour is better defined. I noticed it when accidentally having javascript disabled.

The reasoning with mobile browsers makes sense however. I read somewhere that the autofocus attribute is disabled on mobile webkit, but I wasn't able to test that out.

Seldaek:
Yeah i fully agree and I couldn't reproduce it myself on mobile but someone
complained and this workaround was seemingly harmless enough.

PR #663, opened 2016-04-18
Suggesters information

### LINKED ISSUE #582: Suggestions and dependents page
Could be nice to count suggested package as a dependency for the dependents section.

Concrete example: https://packagist.org/packages/sllh/iso-codes-validator/dependents

You can see 2 package that depends of this one.

But in this package: https://packagist.org/packages/ronanguilloux/isocodes

You can see that sllh/iso-codes-validator is suggested. It should appear on dependents section IMO.

Soullivaneuh:
Closes #582.

Suggestions page content:

image

Package info box:

image

### DIFF
--- a/src/Packagist/WebBundle/Entity/PackageRepository.php
+++ b/src/Packagist/WebBundle/Entity/PackageRepository.php
@@ -357,6 +357,51 @@ class PackageRepository extends EntityRepository
             ->getResult();
     }
 
+    public function getSuggestCount($name)
+    {
+        $apc = extension_loaded('apcu');
+
+        if ($apc) {
+            $count = apcu_fetch('packagist_suggesterCount_'.$name);
+        }
+
+        if (!isset($count) || !is_numeric($count)) {
+            $count = $this->getEntityManager()->getConnection()->fetchColumn(
+                "SELECT COUNT(DISTINCT v.package_id)
+                FROM package_version v
+                LEFT JOIN link_suggest s ON v.id = s.version_id AND s.packageName = :name
+                WHERE v.development AND (s.packageName IS NOT NULL)",
+                ['name' => $name]
+            );
+
+            if ($apc) {
+                apcu_store('packagist_dependentsCount_'.$name, $count, 7*86400);
+            }
+        }
+
+        return (int) $count;
+    }
+
+    public function getSuggests($name, $offset = 0, $limit = 15)
+    {
+        $qb = $this->getEntityManager()->createQueryBuilder();
+        $qb->select('p.id, p.name, p.description, p.language, p.abandoned, p.replacementPackage')
+            ->from('Packagist\WebBundle\Entity\Package', 'p')
+            ->join('p.versions', 'v')
+            ->leftJoin('v.suggest', 's')
+            ->where('v.development = true')
+            ->andWhere('s.packageName = :name')
+            ->groupBy('p.id, p.name, p.description, p.language, p.abandoned, p.replacementPackage')
+            ->orderBy('p.name')
+            ->setParameter('name', $name);
+
+        return $qb->getQuery()
+            ->setMaxResults($limit)
+            ->setFirstResult($offset)
+            ->useResultCache(true, 7*86400, 'suggesters_'.$name.'_'.$offset.'_'.$limit)
+            ->getResult();
+    }
+
     private function addFilters(QueryBuilder $qb, array $filters)
     {
         foreach ($filters as $name => $value) {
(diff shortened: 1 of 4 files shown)


src/Packagist/WebBundle/Entity/PackageRepository.php
```diff
@@ -357,6 +357,51 @@ public function getDependents($name, $offset = 0, $limit = 15)
             ->getResult();
     }
 
+    public function getSuggestCount($name)
+    {
+        $apc = extension_loaded('apcu');
+
+        if ($apc) {
+            $count = apcu_fetch('packagist_dependentsCount_'.$name);
```
Seldaek: This should be suggesterCount

src/Packagist/WebBundle/Entity/PackageRepository.php
```diff
@@ -357,6 +357,51 @@ public function getDependents($name, $offset = 0, $limit = 15)
...
+            ->where('v.development = true')
+            ->andWhere('s.packageName = :name')
+            ->groupBy('p.id, p.name, p.description, p.language, p.abandoned, p.replacementPackage')
+            ->orderBy('p.name')
+            ->setParameter('name', $name);
+
+        return $qb->getQuery()
+            ->setMaxResults($limit)
+            ->setFirstResult($offset)
+            ->useResultCache(true, 7*86400, 'dependents_'.$name.'_'.$offset.'_'.$limit)
```
Seldaek: suggesters_*

Soullivaneuh: Oops, ugly copy/paste spotted! :laughing:

Seldaek:
I think I'd use Suggesters instead of Suggestions

Soullivaneuh:
> I think I'd use Suggesters instead of Suggestions

Good catch! I was not sure of the work to use...

Soullivaneuh:
@Seldaek all done.

src/Packagist/WebBundle/Resources/views/Package/suggesters.html.twig
```diff
@@ -0,0 +1,29 @@
...
+                <h2 class="title">
+                    <a href="{{ path("view_package", {name: name}) }}">{{ name }}</a> suggesters
+                    <small>({{ count }})</small>
+                </h2>
+            </div>
+        </div>
+    </div>
+
+    <section class="row">
+        <section class="col-md-12">
```
stof: nesting sections like that does not make sense.

stof: btw, using the bootstrap grid just to make a single full-width column is useless. You could remove both classes (and this will remove the duplicate tag btw)

Soullivaneuh:
Failing Travis not related IMO.

src/Packagist/WebBundle/Resources/views/Package/suggesters.html.twig
Soullivaneuh: This is a copy of `dependents.html.twig`.

@Seldaek Should I update this one too?

Seldaek: There might be other css than just the bootstrap stuff attached to these classes, I don't remember, so maybe best to keep it as it is.. doesn't really matter and it works.

Seldaek: Unless you want to spend time double checking that it really does not change anything to remove them, then sure go ahead.

Soullivaneuh: I'll kept this as is. I think this is good for another PR.

Soullivaneuh: BTW @stof, I agree for `section` case but not for `row` and `col-md-12`.

There is no parent `row` on the whole page. If we want to refactor, we'll have to refactor the hole templates folder.

Another great things we can do later is to use [bootlint](https://github.com/twbs/bootlint) to check if our templates are good according to Bootstrap conventions.

Soullivaneuh:
@Seldaek Commits squashed and message amended. Are we OK for this PR?

src/Packagist/WebBundle/Resources/views/Package/suggesters.html.twig
stof: @Soullivaneuh this is not an issue. This is what a row does in bootstrap:

- clearfix to ensure floated columns are kept inside it
- negative left and right margins to compensate the column padding

and here is what a column does:

- float left
- set size (``col-*-12`` is full width)
- adding padding to create the gutter

So when using a row with a single -12 column in it, you just create a block talking all the width of the parent (but using 2 divs to achieve it). And this is exactly the behavior of a block element in HTML when no additional styles are applied.

there is nothing in bootstrap forcing to have everything nested inside rows (this is needed only when using a multi-column layout).

Seldaek:
Yeah it looks fine I think, I'll try and deploy some time.

PR #666, opened 2016-04-20
[WIP] Put abandonned package at end of the list

### LINKED ISSUE #664: Hide abandoned packages by default
Although I think hiding not stable package by default is not a good idea (see https://github.com/composer/packagist/issues/300#issuecomment-211368641), I think hiding abandoned ones would be great.

Indeed, abandoned packages mean "Do no use it!", so we showing them on the list by default?

What I suggest:

* Hiding abandoned package by default.
* Add a "show abandoned" button.
* Do not apply this rule on search, the "abandoned" warning on package name will do the job.

Soullivaneuh:
Closes #664.

image

- [x] `UserController` packages
- [x] All packages list
- [x] Dependent packages list
- [ ] Suggester packages list (waiting #663 to be merged first)
- [ ] Search results

@Seldaek I don't know how to proper handle search result. Could you give me some clues?

In a global way, am I missing some sections?

### DIFF
--- a/src/Packagist/WebBundle/Entity/PackageRepository.php
+++ b/src/Packagist/WebBundle/Entity/PackageRepository.php
@@ -268,7 +268,7 @@ class PackageRepository extends EntityRepository
         return $qb->getQuery()->getResult();
     }
 
-    public function getFilteredQueryBuilder(array $filters = array())
+    public function getFilteredQueryBuilder(array $filters = array(), $orderByName = false)
     {
         $qb = $this->getEntityManager()->createQueryBuilder();
         $qb->select('p')
@@ -280,6 +280,11 @@ class PackageRepository extends EntityRepository
         }
 
         $qb->orderBy('p.abandoned');
+        if (true === $orderByName) {
+            $qb->addOrderBy('p.name');
+        } else {
+            $qb->addOrderBy('p.id', 'DESC');
+        }
 
         $this->addFilters($qb, $filters);
 
--- a/src/Packagist/WebBundle/Entity/PackageRepository.php
+++ b/src/Packagist/WebBundle/Entity/PackageRepository.php
@@ -279,7 +279,7 @@ class PackageRepository extends EntityRepository
             $qb->leftJoin('v.tags', 't');
         }
 
-        $qb->orderBy('p.id', 'DESC');
+        $qb->orderBy('p.abandoned');
 
         $this->addFilters($qb, $filters);
 
@@ -347,7 +347,8 @@ class PackageRepository extends EntityRepository
             ->where('v.development = true')
             ->andWhere('(r.packageName = :name OR dr.packageName = :name)')
             ->groupBy('p.id, p.name, p.description, p.language, p.abandoned, p.replacementPackage')
-            ->orderBy('p.name')
+            ->orderBy('p.abandoned')
+            ->addOrderBy('p.name')
             ->setParameter('name', $name);
 
         return $qb->getQuery()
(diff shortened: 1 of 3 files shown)


src/Packagist/WebBundle/Entity/PackageRepository.php
```diff
@@ -279,7 +279,7 @@ public function getFilteredQueryBuilder(array $filters = array())
             $qb->leftJoin('v.tags', 't');
         }
 
-        $qb->orderBy('p.id', 'DESC');
```
stof: the order by id should also be kept, to make the sorting stable (but after the abandoned flag)

Soullivaneuh: @stof I replaced this order here: https://github.com/composer/packagist/pull/666/files#diff-fa57170b5eba2155e6e39953836f12efR61

`getFilteredQueryBuilder` method is used by `UserController` and `PackageController` but not with the same order. I had to do this way.

The order is not changed at all in any case, except for abandoned packages.

stof: Well, the ordering is stable here because always applies an internal sorting last, ensuring stable ordering. But this is a non-standard behavior. SQL does not define anything like that (for instance, in Postgresql, the order of elements returned when there is no order by clause is very likely to change between executions, making it unusable in a paginated query).
And I'm not even sure that MySQL guarantees BC about what the internal sorting is. So keeping an explicit sorting would be better (the sorting based on abandoned is not enoguh to make things stable, as there are thousands of ex-aequo for this field)

Soullivaneuh: But if I write:

```php
$qb->orderBy('id', 'DESC');
$qb->addOrderBy('p.abandoned');
```

And want to change the order by name on user controller, it will erases the `p.abandoned` order.

Am I right? In this case, what would be the solution?

And BTW, the `orderBy` is finally set on each controller, so what is the problem?

stof: you should order first by abandoned, and then by id (to sort packages being ex-aequo for abandoned). sorting by id first would not let any ex-aequo, making subsequent sorting a no-op

And regarding the controllers overriding the ordering, this is an issue, as you would have to sort by abandoned in the controllers too (as they delete any order set here). a better solution would be to pass an argument to the method to tell it to order by name or no.

Soullivaneuh: > a better solution would be to pass an argument to the method to tell it to order by name or no.

Will try this way.

Soullivaneuh: @stof done.

src/Packagist/WebBundle/Entity/PackageRepository.php
```diff
@@ -347,7 +352,8 @@ public function getDependents($name, $offset = 0, $limit = 15)
             ->where('v.development = true')
             ->andWhere('(r.packageName = :name OR dr.packageName = :name)')
             ->groupBy('p.id, p.name, p.description, p.language, p.abandoned, p.replacementPackage')
-            ->orderBy('p.name')
+            ->orderBy('p.abandoned')
```
stof: I suggest using ``addOrderBy`` for both, to avoid having to change 2 lines the next time we want to add something at the beginning

Soullivaneuh:
@stof @Seldaek code upated.

Seldaek:
Thanks

PR #669, opened 2016-04-26
Redirect to package detail from search when only one package was found

finwe:
This is a PR in a "RFC state" - I was't able to test it properly as it would require Solr.

It is also possible that this would be confusing to users when found package is just very similar by name - so some query/found package name matching could be appropriate.


stof:
I'm not sure this is a good idea. As you said, this may be confusing.

Seldaek:
Yeah I don't think this is so good to have different behaviors like that. All you have to do to go to the detail page is to press enter so I don't think that's such a hurdle.

PR #684, opened 2016-06-11
Made the steps to set up the hook more readable

PeeHaa:
I've changed the long paragraph to a list with bullets so that the steps to set up the hook are more readable.

**Before this PR**

packagist-before

**After this PR**

packagist-after


Seldaek:
Thanks

PR #692, opened 2016-06-21
Added type parameter to search

pierredup:
Allow the search to specify the type of package to search for.

Related to composer/composer#5454

PR #693, opened 2016-06-22
Replace type placeholder in the search

pierredup:
This is a follow-up to #692.

When someone is not using the latest version of composer (which includes the type parameter for the search), then searching for a package includes the type parameter placeholder. E.G (`https://packagist.org/search.json?q=symfony&type=%type%`). So the placeholder needs to be removed from the type query string.

### DIFF
--- a/src/Packagist/WebBundle/Controller/WebController.php
+++ b/src/Packagist/WebBundle/Controller/WebController.php
@@ -75,7 +75,7 @@ class WebController extends Controller
 
         $this->computeSearchQuery($req, $filteredOrderBys);
 
-        $typeFilter = $req->query->get('type');
+        $typeFilter = str_replace('%type%', '', $req->query->get('type'));
         $tagsFilter = $req->query->get('tags');
 
         if ($req->query->has('search_query') || $typeFilter || $tagsFilter) {


src/Packagist/WebBundle/Controller/WebController.php
```diff
@@ -75,7 +75,7 @@ public function searchAction(Request $req)
 
         $this->computeSearchQuery($req, $filteredOrderBys);
 
-        $typeFilter = $req->query->get('type');
+        $typeFilter = str_replace('%type%', null, $req->query->get('type'));
```
staabm: Shouldnt it be replaced by a empty string?

pierredup: The result will be exactly the same, so I don't think it makes a difference

staabm: Result is the same, right. str_replace expects a strong though

alcohol:
Why fix this here, and not in Composer instead, which should not be sending a placeholder in the first place?

pierredup:
Composer is not sending the placeholder. Well, not in the latest version at least.
The issue is that older versions of composer (everything that is not from the latest master) doesn't support the type search. So it only sends the search parameter. When using the latest composer, then there is no issue, since it won't send the placeholder.
This change is just to fix the search for everyone using stable versions of composer.

alcohol:
Sorry, I forgot that the search string is actually defined by Packagist, and not part of Composer itself.

Seldaek:
Thanks, deployed and fixed

PR #694, opened 2016-06-23
Add more data in json response of viewPackageAction

psampaz:
This pull request adds the following info in the  json response of viewPackageAction

dependents (count)
suggesters (count)
github_stars
github_watchers
github_forks
github_open_issues
language


Seldaek:
Thanks

PR #699, opened 2016-07-04
Update syntax

indragunawan:
- Update array() to short syntax []
- Apply php-cs-fixer level symfony
- Change deprecated type hint Symfony\Component\HttpKernel\Log\LoggerInterface to Psr\Log\LoggerInterface


app/check.php
```diff
@@ -67,7 +67,7 @@ function_exists('xcache_set')
  */
 function check($boolean, $message, $help = '', $fatal = false)
 {
-    echo $boolean ? "  OK        " : sprintf("\n\n[[%s]] ", $fatal ? ' ERROR ' : 'WARNING');
+    echo $boolean ? '  OK        ' : sprintf("\n\n[[%s]] ", $fatal ? ' ERROR ' : 'WARNING');
     echo sprintf("$message%s\n", $boolean ? '' : ': FAILED');
 
     if (!$boolean) {
```
stof: This file is updated automatically from SensioDistributionBundle (by a composer hook), so please don't edit it by hand (otherwise, running ``composer install`` will leave the project in a modified state)

web/app.php
```diff
@@ -5,7 +5,7 @@
 use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
-/**
+/*
```
stof: this is wrong. The right fix would be to fix the content to say ``@var $loader ...``

Seldaek:
I cherry-picked your last commit (https://github.com/composer/packagist/commit/d841698a8bc42d737f3dced79491e369d7e4a462) but for the rest, running php-cs-fixer is my job :) I won't review such huge changes it's a waste of my time.

indragunawan:
how about changing the array syntax? 
btw thanks for advice

Seldaek:
It creates so many changes that I have to review to see if you didn't change anything else.. If I do it myself I can trust that php-cs-fixer didn't change anything else. So I can do it, and I might eventually, but it's not a pressing matter.

indragunawan:
ok. i know your responsibility. 
change array syntax not using php-cs-fixer btw. 
thanks for merging :)

PR #700, opened 2016-07-05
exclude Tests from classmap

indragunawan:
by excluding Tests from autoload classmap, it reduce about 371 lines.

and when i run composer validate it show warning
name : The property name is required
The property _comment is not defined and the definition does not allow additional properties

and i think its @Seldaek will solve the warning.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
     },
     "autoload": {
         "psr-4": { "Packagist\\": "src/Packagist/" },
-        "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
+        "classmap": [ "app/AppKernel.php", "app/AppCache.php" ],
+        "exclude-from-classmap": [ "**/Tests/" ]
     },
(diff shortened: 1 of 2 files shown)


indragunawan:
at my local, the phpunit result is OK and i dont know why it can by FAILED

Seldaek:
Thanks, I fixed it in a slightly different way for clarity, but regarding the warning that's just because you don't have the 1.2.0-rc version of composer yet I guess. It'll fix itself in time.

indragunawan:
when only add our project tests nameapace, how about vendor tests nampace, its also generate at autoload classmap. 
did we still need vendor tests? whats your opinion?

Seldaek:
Well vendor tests should be excluded by the vendors.. If you exclude
**/tests you mind end up excluding a valid class that's inside a tests
namespace. Chances are low but it's technically not the right thing to do.
Every package should be responsible for its own autoloading rules.

PR #703, opened 2016-07-18
Y-axis had incorrect/duplicate labels, fixes #696

Dumkaaa:
fixes #696

## **Before: incorrect/duplicate labels**
before

## **After: correct labels**
after

PR #705, opened 2016-07-19
fix Travis CI build failed

### LINKED ISSUE #704: The Travis CI build failed
From @stof comment on  https://github.com/composer/packagist/pull/703#issuecomment-233587844:

> @Seldaek it would be great to fix Packagist\WebBundle\Tests\Package\SymlinkDumperTest::testGetTargetListingBlocks so that Travis is green again.

Dumkaaa:
**fix #704**

as @stof say in https://github.com/composer/packagist/pull/703#issuecomment-233587844, The Travis CI build failed in `Packagist\WebBundle\Tests\Package\SymlinkDumperTest::testGetTargetListingBlocks`

I think this happened because the user who developed this functionality(tests) from California (@lavoiesl) and his timezone is 'America/California'. Travis timezone is 'UTC'.

I set the default timezone to 'UTC' in SymlinkDumperTest::testGetTargetListingBlocks

So Travis  turns green again, I hope :pray: 

**fix #704**


Dumkaaa:
And yes! :sunglasses: 
From the second attempt it was succeeded to solve a problem,  so that Travis is green again! :green_apple:

stof:
A simpler fix to avoid having to force the timezone (which breaks the isolation of tests btw, as you don't reset the global state) would be to simply change the reference date. the issue is that ``2015-05-31`` is a Sunday, and ``monday last week`` has a different meaning depending on the timezone in this case (as the week starts either on Monday or Sunday). Using ``2015-05-30`` would make it work fine for any timezone.

Dumkaaa:
Yeah, right. It was possible to do so. This is another solution.
@stof, you think it would be better to do that?

Seldaek:
IMO the timezone should be forced in the test bootstrap and not in the test itself, because the environment and packagist code expects a UTC timezone (it's forced in app.php and app/console).

Dumkaaa:
@Seldaek please check this time.
I set the timezone in the test bootstrap
It looks like a workable solution

PR #708, opened 2016-07-29
changes i.e. to e.g.

bshaffer:
This should be e.g. ("for example") rather than i.e. ("in other words"). See [this very serious grammar reference](http://theoatmeal.com/comics/ie) for more information.

Context: https://packagist.org/packages/submit


Seldaek:
Thanks

PR #709, opened 2016-08-03
Fixed repoUrl which links to stargazers, watchers and forks

ondrejmirtes:
I found out that the links on the package detail page sometimes lead to 404s:

https://packagist.org/packages/phpstan/phpstan:
- https://github.com/phpstan/phpstan.git/stargazers
- https://github.com/phpstan/phpstan.git/watchers

But sometimes the URL is right:

https://packagist.org/packages/composer/composer:
- https://github.com/composer/composer/stargazers
- https://github.com/composer/composer/watchers

I really don't know what's the underlying issue, if the URL reported by GitHub is inconsistent or what, so I tried to fix it with a conditional trimming of the `.git` from the end of the URL.

### DIFF
--- a/src/Packagist/WebBundle/Entity/Package.php
+++ b/src/Packagist/WebBundle/Entity/Package.php
@@ -552,6 +552,10 @@ class Package
         }
 
         $repoUrl = preg_replace('{^git@github.com:}i', 'https://github.com/', $repoUrl);
+        $dotGitLength = -strlen('.git');
+        if (substr($repoUrl, -$dotGitLength, $dotGitLength) === '.git') {
+            $repoUrl = substr($repoUrl, 0, -$dotGitLength);
+        }
         $this->repository = $repoUrl;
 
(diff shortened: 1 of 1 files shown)


Seldaek:
That's odd.. I fixed the whole DB a while back and made sure that .git URLs didn't make their way in anymore.. but it looks like that didn't quite work :) (see https://github.com/composer/packagist/commit/b0ab24bd4df1a467b6cf6fff4bb2dbc3b6c89af6, the key is https://github.com/composer/packagist/blob/b0ab24bd4df1a467b6cf6fff4bb2dbc3b6c89af6/src/Packagist/WebBundle/Entity/Package.php#L580)

I'll keep this open as a reminder but stripping .git isn't a solution because for non-github URLs the .git might well be needed.

Seldaek:
As a fix for your package though, just change the URL to be the https://github.com/foo/bar URL and all will be fine.

PR #724, opened 2016-09-13
A simple change - pre border using #f28d1a

thiagotoledo:
Using the color #f28d1a plus look better , easier to read the code (that is in black) .


Seldaek:
Thanks for the visual tweak, but I am not sure I understand how it makes the code easier to read :) Anyway it looks better I agree.

PR #725, opened 2016-09-13
Standardizing color of links

thiagotoledo:
Standardizing the color of links and facilitating navigation creating the state visited with a small visual changes .


Seldaek:
Sorry but I don't quite agree here, it just means you have less contrast for the links vs their backgrounds, and making underlined for visited doesn't quite make sense to me.

PR #727, opened 2016-09-30
Change github menu item name in packagist "GitHub Service Hook HowTo"

legionth:
The menu item "Webhooks & Services" does not longer exist. It is now named "Integrations & services".

github-settings-menu


stof:
:+1:

Seldaek:
Thanks

Issue #65, opened 2011-12-09
Add Doctrine fixture file


datiecher:
It would be easier to attract contributors to the project if Packagist had a fixture file that would load some packages, authors, etc. 

I could have a go at doing it myself but I'm not comfortable enough with the codebase yet.

stof:
@Seldaek I suggest tagging this as easy-pick

PR #730, opened 2016-10-04
Add missing services for UserManipulator

mhujer:
I started working on https://github.com/composer/packagist/issues/65 but I found out, that the `UserManipulator` is broken since [upgrade to dev version of FOSUserBundle](https://github.com/composer/packagist/commit/b6588d10de4bba3f10263c0e38cba9efded09d18#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R41) (it is not possible to call `php app\console fos:user:create`).

This PR adds the missing dependencies.


Seldaek:
Thanks

PR #731, opened 2016-10-12
add package name to composer.json

danslo:
I'm working on a little sideproject that adds features to Packagist but those features don't belong in this repository. In this case I actually want to be able to install this repository as a composer dependency and then use [bundle inheritance](http://symfony.com/doc/current/bundles/inheritance.html) to make my changes.

Problem is, I can't install it because it doesn't have a name, so here's the proposed 1-line change. A tag will be nice too but I suppose I could manage without it :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,5 @@
 {
+    "name": "composer/packagist",
     "description": "Package Repository Website",
     "keywords": ["package", "composer"],
     "homepage": "http://packagist.org/",


Seldaek:
I'd really rather not publish it on packagist because it's not made to be reused by others much, and if we put it up there it kinda sends the wrong signal :/

danslo:
Fair enough, I suppose it wouldn't be too difficult to maintain a fork that I pull into once in a while. Thanks for the quick reply.

PR #737, opened 2016-10-18
Logged in user is redirected to homepage when accessing registration page

### LINKED ISSUE #712: Logged in users probably shouldn't be able to register
If a user is logged in they shouldn't be able to register: https://packagist.org/register/

mhujer:
Fixes #712


Seldaek:
Thanks

Issue #744, opened 2016-10-29
Use `scrollIntoView(false)` to focus the selected search result


rudiedirkx:
There's a block of code to focus/position the newly selected LI:

```
var elTop = nextSelected.position().top,
    elHeight = nextSelected.height(),
    windowTop = $(window).scrollTop(),
    windowHeight = $(window).height();

if (elTop < windowTop) {
    $(window).scrollTop(elTop);
} else if (elTop + elHeight > windowTop + windowHeight) {
    $(window).scrollTop(elTop + elHeight + 20 - windowHeight);
}
```

I don't know why, but very often that doesn't work. It trails a few items behind, or it doesn't do anything. (And sometimes it does work.) I don't even want to fix why it doesn't work, because there's a perfect native replacement: [`Element.scrollIntoView(false)`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView). The entire above block would be replaced by:

```
nextSelected.get(0).scrollIntoView(false);
```

The `false` is to align the item to the bottom of the viewport, because there's always room there, and it scrolls the page to the top by default. (The top has the fixed bar.)

Support is allround, so this would be a good replacement IMO. I've only tested by hacking the live JS, but that works perfectly.

Problem exists on stable Firefox and Chrome on Win. Solution works for both too. If it's PR worthy, I will.

alcohol:
I'm no JS expert for sure, but this seems like an elegant solution. @stof can you comment on this proposed change?

PR #745, opened 2016-10-29
#744 Use scrollIntoView instead of calculating scrollTop

rudiedirkx:
See #744


stof:
👍

Seldaek:
Thanks, not sure at which point this broke but indeed it's not quite working. The scrollIntoView(false) isn't perfect when going back up I guess but at least going down it should work a lot better.

Issue #556, opened 2015-08-05
README integration with Gitlab


ragboyjr:
I've submitted several packages to `packagist` and none of them show the README at the bottom of the package. Here is an [example package](https://packagist.org/packages/krak/array) that doesn't have the README. In fact, all of them from the [Krak](https://packagist.org/packages/krak) package aren't showing a README. I've compared my `composer.json` to many other projects, and I don't see anything wrong with them. The only thing different is that my projects are from gitlab where as the others are from github. 

I don't know if this is a bug or just a missing feature, but I wanted to report it here in case I was missing something.

Thanks!

alcohol:
As you can see here, this functionality seems currently implemented for GitHub only:
https://github.com/composer/packagist/blob/master/src/Packagist/WebBundle/Package/Updater.php#L393

If Gitlab has a similar API, then you could write a method similar to the above and call it after this block:
https://github.com/composer/packagist/blob/master/src/Packagist/WebBundle/Package/Updater.php#L170-L172

We are open to pull requests ;-)

ragboyjr:
Thanks! It looks like Gitlab doesn't support this feature as of yet. I added a feature request [here](http://feedback.gitlab.com/forums/176466-general/suggestions/9177020-add-repo-contents-api-like-github), but it's unknown if it will get implemented any time soon. So I don't know exactly what you guys want to do with this

stof:
Well, nothing will be done with this until an API is available.

Garbee:
Would the [repository file](https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/repository_files.md#get-file-from-repository) API handle this case?

alcohol:
@Garbee technically, yes I reckon it would. But it isn't as straight forward as the Github API.

stof:
@Garbee  no it would not, as this would not return a rendered readme, forcing Packagist to implement rendering of all text formats supported by gitlab for their readmes (and would require guessing the name of the readme file, requiring several API calls to gitlab).

Vote on https://gitlab.com/gitlab-org/gitlab-ce/issues/24124 to support the feature request on gitlab

tflori:
hm why not add a key `readme` in composer.json which is a relative path to the readme? packagist can then update the readme from this file...

alcohol:
Because adding yet another composer.json config option for something as insignificant as a readme is really not a viable solution.

stof:
[...]

PR #750, opened 2016-11-12
add readme support for non github repos 

tflori:
solves #556 

The tests will fail till https://github.com/composer/composer/pull/5862 is merged.

### DIFF
--- a/src/Packagist/WebBundle/Package/Updater.php
+++ b/src/Packagist/WebBundle/Package/Updater.php
@@ -615,14 +615,16 @@ class Updater
             }
         }
 
-        // convert relative to absolute images
-        $images = $dom->getElementsByTagName('img');
-        foreach ($images as $img) {
-            if ($isGithub && false === strpos($img->getAttribute('src'), '//')) {
-                $img->setAttribute(
-                    'src',
-                    'https://raw.github.com/'.$owner.'/'.$repo.'/HEAD/'.$img->getAttribute('src')
-                );
+        if ($isGithub) {
+            // convert relative to absolute images
+            $images = $dom->getElementsByTagName('img');
+            foreach ($images as $img) {
+                if (false === strpos($img->getAttribute('src'), '//')) {
+                    $img->setAttribute(
+                        'src',
+                        'https://raw.github.com/'.$owner.'/'.$repo.'/HEAD/'.$img->getAttribute('src')
+                    );
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


tflori:
> The tests will fail till composer/composer#5862 is merged.

Sorry, that is wrong. The tests will not fail because there is not tests for it. I don't know the test setup. To create an error we need to update a package that has `{ "readme": "README.md" }` in the `composer.json`. Then it will throw a fatal error because `VcsDriver::getFileContent()` does not exist on the current composer version.

src/Packagist/WebBundle/Package/Updater.php
```diff
@@ -566,6 +551,97 @@ private function updateGitHubInfo(RemoteFilesystem $rfs, Package $package, $owne
...
+                    'href',
+                    'https://github.com/'.$owner.'/'.$repo.'/blob/HEAD/'.$link->getAttribute('href')
+                );
+            }
+        }
+
+        // convert relative to absolute images
+        $images = $dom->getElementsByTagName('img');
+        foreach ($images as $img) {
+            if ($isGithub && false === strpos($img->getAttribute('src'), '//')) {
```
stof: if it is not github, you don't even need to look for images, instead of searching them and doing nothing inside the loop

tflori:
@Seldaek now that composer/feature-getFileContent is merged this works too. I tried it in a local instance. I really would like to have a test but I'm not sure how to test it.

tflori:
I found solution to test it, but it is more or less ugly. But anyway: there is a warning that getFileContent does not exist. I guess that it is using the latest stable composer 1.2.2 and not dev-master. So we have to wait for next composer release.

tflori:
no warnings in tests - from my site: it works :smile:

Seldaek:
Thanks

src/Packagist/WebBundle/Package/Updater.php
```diff
@@ -456,6 +459,56 @@ private function updateInformation(Package $package, PackageInterface $data, $fl
...
+                $ext = substr($readmeFile, strrpos($readmeFile, '.'));
+
+                if ($ext === $readmeFile) {
+                    $ext = '.txt';
+                }
+
+                switch ($ext) {
+                    case '.txt':
+                        $source = $driver->getFileContent($readmeFile, $driver->getRootIdentifier());
+                        $package->setReadme('<pre>' . htmlspecialchars($source) . '</pre>');
```
stof: this is missign the ``!empty`` check

Seldaek: True, will fix

Issue #754, opened 2016-12-16
add maintainer avatar at package data API


IndraGunawan:
when get package data using API the `maintainer` object only provide name. i think it would be good if API of package data can provide gravatar url as maintainer avatar

https://packagist.org/packages/monolog/monolog.json

```
{
    "package": {
        "name": "monolog/monolog",
        "maintainers": [
            {
                "name": "Seldaek",
                "avatar": "//www.gravatar.com/avatar/48b79d17cd8a911327cbd88c122b1efb",
            }
        ],
        ...
    }
}
```

Seldaek:
Sounds fine to me, if you wanna send a PR..

PR #755, opened 2016-12-18
add maintainer avatar_url to package detail JSON

IndraGunawan:
#754


Seldaek:
Thanks, it's now live on the API

PR #770, opened 2017-02-24
Improve Updater to not remove useful README titles

### LINKED ISSUE #764: No stripping of first h1/h2 of README when it is not the first content of the README
Concerning [\Packagist\WebBundle\Package\Updater:632](https://github.com/composer/packagist/blob/master/src/Packagist/WebBundle/Package/Updater.php#L632)

In this code the first `h1`, or `h2` if no `h1` is found, is stripped from the README. This allows for stripping of a title in the README and increases readability. But I think an extra check should be included to check whether this header is actually the first line of the README.

I personally don't usually start the README of a package with the title of the package, since I think this is redundant. Probably for the same reason Packagist strips the first header :) That lead to the first header to be somewhere after three paragraphs and it is subsequently removed from the README display in Packagist. 

I think adding an extra check to only hide the first header if it is *not* preceeded by text is the simplest solution for unexplained removal of content of the README.

javiereguiluz:
This tries to fix #764 by only removing the first `<h1>` element if it's the first actual element of the README file.

### DIFF
--- a/src/Packagist/WebBundle/Package/Updater.php
+++ b/src/Packagist/WebBundle/Package/Updater.php
@@ -629,12 +629,10 @@ class Updater
             }
         }
 
-        // remove first title as it's usually the project name which we don't need
-        if ($dom->getElementsByTagName('h1')->length) {
+        // remove first page element if it's a h1, because it's usually the project
+        // name or the `README` string which we don't need
+        if ('h1' === $dom->childNodes->item(0)->nodeName) {
             $first = $dom->getElementsByTagName('h1')->item(0);
-            $first->parentNode->removeChild($first);
-        } elseif ($dom->getElementsByTagName('h2')->length) {
-            $first = $dom->getElementsByTagName('h2')->item(0);
             $first->parentNode->removeChild($first);
         }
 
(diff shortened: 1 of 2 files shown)


stof:
We should keep removing h2 if it is the first element too. Some projects use a h2 rather than a h1 to take less space in the github rendering

src/Packagist/WebBundle/Package/Updater.php
```diff
@@ -629,11 +629,12 @@ private function prepareReadme($readme, $isGithub = false, $owner = null, $repo
             }
         }
 
-        // remove first title as it's usually the project name which we don't need
-        if ($dom->getElementsByTagName('h1')->length) {
+        // remove first page element if it's a <h1> or <h2>, because it's usually
+        // the project name or the `README` string which we don't need
+        if ('h1' === $dom->childNodes->item(0)->nodeName) {
             $first = $dom->getElementsByTagName('h1')->item(0);
```
stof: could be simplified:

```php
$first = $dom->childNodes->item(0);

if ($first && ('h1' === $first->nodeName || 'h2' === $first->nodeName)) {
    $first->parentNode->removeChild($first);
}
```

My snippet also account for the case of empty readme, where there could be no first node.

javiereguiluz: Much better! Thank you.

Seldaek:
Thanks!

PR #784, opened 2017-04-23
Allow nested project paths

sadortun:
Gitlab now support sub-groups in repositories. Updated regex.

https://docs.gitlab.com/ce/user/group/subgroups/

Fixes error : 
> Hook executed successfully but returned HTTP 406 {"status"=>"error", "message"=>"Could not parse payload repository URL"}

### DIFF
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -110,11 +110,10 @@ class ApiController extends Controller
         }
 
         if (isset($payload['repository']['url'])) { // github/anything hook
-            $urlRegex = '{^(?:ssh://git@|https?://|git://|git@)?(?P<host>[a-z0-9.-]+)(?::[0-9]+/|[:/])(?P<path>[\w.-]+/[\w.-]+?)(?:\.git|/)?$}i';
-            $url = $payload['repository']['url'];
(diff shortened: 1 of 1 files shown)


stof:
Composer does not yet support gitlab subgroups. Implementation is in progress in https://github.com/composer/composer/pull/6373
As Packagist uses Composer internally, it needs support in Composer first (it will *also* need a change in the packagist webhook though)

sadortun:
Hi @stof

The repository path and package name are two different things. This only fixes the repository path issue.

Example : 
- composer package name : **example/my-package**
- repository : **HTTPS://git.exanple.com/company/software/example/my-projrct.git

stof:
I never talked about package names

sadortun:
> Composer does not yet support gitlab subgroups. 
@stof We are using composer currently with sub-groups, and it working very well ! Composer have no need to be aware of the sub-groups if the package name does not mention them.

sadortun:
Created a test project if you want to validate : https://Sadortun@gitlab.com/issue-packagist-784/my_sub_group/test_project.git

You can add the project, then trigger the webhook

stof:
@sadortun how are you configuring your repository ?

Seldaek:
Thanks but as far as I can tell there was an issue with the regex tweak you did as it only allowed pairs of foo/bar to repeat, and not just /bar to repeat.. Anyway I think it's fixed now, would be great if you can try it and confirm.

PR #790, opened 2017-05-31
allow divs in readmes

luka-zitnik:
Can we enable divs at packagist.org, like this?

Basically, I want to split a series of images with a block-level container, and GitHub will allow me to do that with rst's `container`/`compound` directive.


luka-zitnik:
Sorry, forget about this. Found another way to do it.

PR #794, opened 2017-06-20
Set version feed Date header to latest release date

brandonkelly:
This will make it possible for third party clients to determine the latest release date from a simple HEAD request, rather than downloading the whole body and checking the `<pubDate>` / `<updated>` value.


Seldaek:
Thanks

Issue #805, opened 2017-08-15
From Solr to Algolia


julienbourdeau:
### Context

@maxiloc met @naderman at Laracon US and they discussed the possibility of using Algolia to power the search on packagist.org. Nils said we should make a PR.

Ideally, I think it would be great to completely get rid of Solr but I'd like to do it step by step.

I'm opening this issue so we can discuss the general idea. **I will open pull request as we move forward and we can keep track of what's going on here.**

### Algolia's Advantages

Here is a list of advantages of switching to Algolia, some are inherent to Algolia, some are just related to the PRs we'll open and could probably be achieved with Solr.

* ⌛️ Faster that the current search
* 👍 More relevant (see below)
* ✏️ Typo-tolerant (_laarvel_ query will find Laravel packages, see below)
* ⛱ No Solr setup to maintain and updates
* 💰 No server cost

For reference, Algolia powers [yarnpkg.com](https://yarnpkg.com).

## Tasks list

- [x] Index packages in Algolia
- [x] Replace main search on the frontend
- [ ] Power RSS (browse page)
- [ ] Next step TBD

## Help required

Packagist is not an easy project and some aspect can be hard to grasp. As a first time contributor, working on something pretty deep, I would require some help to full understand what's going on.

More importantly, I would need an extract (even just a chunk) of the databases (MySQL and Redis). I totally understand that this might feel a little weird, but in order to test the indexation properly, that would really help. I'm willing to sign anything you want if needed.

### Relevancy examples 

**Before**
![packagist](https://user-images.githubusercontent.com/1525636/29296724-057b50d6-8111-11e7-9b2b-d5eddb122948.png)

**After**
<img width="980" alt="packagist" src="https://user-images.githubusercontent.com/1525636/29296812-86b97ed4-8111-11e7-8c9f-9f707e398bdd.png">

### Typo-tolerance

**Before**
<img width="727" alt="packagist" src="https://user-images.githubusercontent.com/1525636/29296749-28bad21a-8111-11e7-9b5f-7ea12b59db3b.png">

**After**

<img width="957" alt="packagist" src="https://user-images.githubusercontent.com/1525636/29296835-a83b8480-8111-11e7-8127-8a043ed8d511.png">

alcohol:
> No server cost

This is a bit of a non-issue, since solr just runs on the same machine as packagist.org itself. @Seldaek can give you more specifics about the resources it consumes, but I am guessing they are negligible. 

[...]

PR #806, opened 2017-08-15
Introducing Algolia

julienbourdeau:
## Context

@maxiloc met @naderman at Laracon US and they discussed the possibility of using Algolia to power the search on packagist.org. Nils said we should make a PR.

Ideally, I think it would be great to completely get rid of Solr but I'd like to do it step by step. This PR is the very first step.

## ETA

I open this issue #805 to keep track of what's left to do and give more details about the work

This PR is not entirely finished but there are a few things missing.

I would really need your help to get the database. So far we used the API to create a index in Algolia and use it in the frontend, but in order make the backend works, I need to reproduce the prod environment.

Then I will polish this PR and explain in details what I did, if necessary.

Would you mind providing the database to me?


composer.json
```diff
@@ -55,7 +55,8 @@
         "knplabs/knp-menu-bundle": "^2.1",
         "ezyang/htmlpurifier": "^4.6",
         "nelmio/cors-bundle": "^1.4",
-        "cebe/markdown": "^1.1"
+        "cebe/markdown": "^1.1",
+        "algolia/algoliasearch-client-php": "^1.18"
```
stof: having a new dependency here without any change in the composer.lock looks suspicious to me.

src/Packagist/WebBundle/Resources/public/js/search.js
```diff
@@ -1,160 +1,79 @@
...
+search.addWidget(
+    instantsearch.widgets.searchBox({
+        container: '#search_query_query',
+        magnifier: false,
+        reset: false,
+        wrapInput: false
+    })
+);
+
+var itemTemplate = `
```
stof: this is leaking global variables

src/Packagist/WebBundle/Resources/public/js/search.js
```diff
@@ -1,160 +1,79 @@
...
-            enter: 13,
-            left: 37,
-            up: 38,
-            right: 39,
-            down: 40
-        };
-
-        if (keymap.up !== event.which && keymap.down !== event.which && keymap.enter !== event.which) {
-            return;
+const search = instantsearch({
```
stof: AFAIK, Packagist does not use ES6 currently

src/Packagist/WebBundle/Resources/public/js/search.js
```diff
@@ -1,160 +1,79 @@
...
+    appId: '6B7GQC987O',
+    apiKey: 'a4b2a9870a8575c9e5c3d2dd00d8de95',
```
stof: should it really be hardcoded here ? Shouldn't it get these from the backend configuration ?

src/Packagist/WebBundle/Resources/views/Web/searchForm.html.twig
```diff
@@ -5,28 +5,7 @@
...
-            {% if hasActiveOrderBy %}
-                <a title="Clear order bys" href="?q={{ searchForm.vars.value.query }}" class="clear active">
-                    <i class="glyphicon glyphicon-remove-circle"></i>
-                </a>
-            {% endif %}
-        </div>
-        {% endspaceless %}
-        <div
-            class="hidden">
-            {{ form_widget(searchForm.orderBys) }}
```
stof: if you remove the rendering of these fields, you should also remove them from the form in the PHP code

composer.json
julienbourdeau: Yes I didn't commit it because I wll probably have to rebase a few times. I will add it to the correct commit when I polish the PR 👌

src/Packagist/WebBundle/Resources/public/js/search.js
julienbourdeau: I will rewrite it in ES5.

src/Packagist/WebBundle/Resources/views/Web/searchForm.html.twig
julienbourdeau: Will do 👍

src/Packagist/WebBundle/Resources/public/js/search.js
julienbourdeau: Indeed 👍

src/Packagist/WebBundle/Resources/views/Web/searchForm.html.twig
julienbourdeau: To be honest I'm trying to remove as little as possible right now, but this can be remove savely I think.

julienbourdeau:
Thank you @stof for your review. I'll update everything quickly.
In the mean time, do you have any idea how I could get the access to the data?

Seldaek:
Merged, finished up the work and deployed it. The indexing is in progress. Let's continue this in #805 - thanks for getting the ball rolling here :)

julienbourdeau:
Thank you for merging! 🎉 I do my best to keep improving it.

PR #809, opened 2017-09-11
[WIP] Add Docker-compose based dev environment

### LINKED ISSUE #808: CI build is not running all tests
I was trying to contribute to this project, and I started to build my local dev env. During this process I discovered that:

 * the Travis build is not running all tests ([only 5!](https://travis-ci.org/composer/packagist/builds/265946063#L455))
 * the Travis build uses PHPUnit 6.3, the project requires 5 (it invokes `phpunit` instead of `vendor/bin/phpunit`)
 * a lot of tests are broken due to the fact that by default in MySQL we have `ONLY_FULL_GROUP_BY` enabled, that triggers some errors when filtering results in queries
 * I'm unable to set up Solr correctly

If it could be helpful, I can open a WIP PR with a basic docker-compose setup that I'm currently using, to help diagnose the issue.

Jean85:
This is to try and fix #808. STILL WIP

[EDIT] Issuing a `docker-compose run --rm php bash` command gives you a CLI inside the PHP container, where it's possible to run composer and phpunit commands. More than one test is still failing.


docker/php/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM php:5.6-cli
```
stof: Travis runs on PHP 7.0, so I think that the prod runs on PHP 7 too, not on 5.6

docker/solr/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM openjdk:7-jre-slim
+
+MAINTAINER Alessandro Lai <alessandro.lai85@gmail.com>
+
+RUN apt-get update \
+   && apt-get install -y curl \
+   && apt-get clean
+
+ADD
```
stof: adding nothing ? Why ?

docker/solr/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN apt-get update \
+   && apt-get install -y curl \
+   && apt-get clean
+
+ADD 
+COPY entrypoint.sh /usr/local/bin/solr-entrypoint.sh
+COPY *.xml /etc/solr/
+
+RUN chmod +x /usr/local/bin/solr-entrypoint.sh
```
stof: is it needed ? couldn't you make the original file executable and have it preserved when copying ?

docker/solr/entrypoint.sh
```diff
@@ -0,0 +1,214 @@
...
+}
+
+run_solr5() {
+    dir_name=$1
+    solr_port=$2
+    ./$dir_name/bin/solr -p $solr_port -c
+    echo "Started"
+}
+
+download_and_run() {
```
stof: Downloading Solr could be done at build-time rather than at start time (and we don't need the support of multiple Solr versions, as Packagist wants to use only 3.6, and if it migrates to Solr 6 in the future, it should also migrate to the official image)

docker/solr/schema.xml
```diff
@@ -1,282 +1,282 @@
-<?xml version="1.0" encoding="UTF-8" ?>
-<schema name="packagist" version="1.4">
```
stof: Why are you changing all lines in these files ? Are you changing line endings ?

docker/php/Dockerfile
Jean85: I'm using that since the composer.json says `"php": ">=5.6"`. Since this is OSS, it would mean a BC otherwise...

docker/solr/schema.xml
Jean85: It was my IDE, it changed line separators.

docker/solr/entrypoint.sh
Jean85: Done!

Jean85:
Thanks @stof for the review, I've addressed all your concerns, apart from the PHP version one: in that case, we should decide if we need to bump the version on `composer.json` or not...

docker/solr/schema.xml
stof: if the file has CRLF in it currently, normalizing to LF is fine (we should have LF)

docker/solr/entrypoint.sh
```diff
@@ -0,0 +1,213 @@
...
+    then
+        echo "SOLR_DOCS not defined, skipping initial indexing"
+    else
+        echo "Indexing $solr_docs"
+        echo "./$dir_name/bin/post -c $collection $solr_docs -p$solr_port"
+        ./$dir_name/bin/post -c $collection $solr_docs -p $solr_port
+    fi
+}
+
+download_and_run 3.6.0
```
stof: you don't need `download_and_run` anymore, as you download at build time now. You only need `run` here.

docker/solr/entrypoint.sh
```diff
@@ -0,0 +1,213 @@
...
+        if [ -z "${SOLR_COLLECTION_CONF}" ]
+        then
+            run_solr5_example $dir_name $SOLR_PORT
+        else
+            run_solr5 $dir_name $SOLR_PORT
+            create_collection $dir_name $SOLR_COLLECTION $SOLR_COLLECTION_CONF $SOLR_PORT
+        fi
+        if [ -z "${SOLR_DOCS}" ]
+        then
+            echo "SOLR_DOCS not defined, skipping initial indexing"
```
stof: IMO, you don't need this `SOLR_DOCS` stuff at all, as Packagist does not use it.

Jean85:
Ok I've simplified the entrypoint script to avoid any unuseful cruft.

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    image: mysql:5.7
+    environment:
+      MYSQL_ROOT_PASSWORD: packagist
+  redis:
+    image: redis:3.2
+  solr:
+    build:
+      context: docker/solr
+    environment:
+      SOLR_CONFS: '/etc/solr'
```
stof: This should be set in the Dockerfile IMO, as this path actually depends on the Dockerfile itself (and so overriding it does not make sense).
We should keep this knowledge in a single place.

docker/solr/entrypoint.sh
```diff
@@ -0,0 +1,94 @@
...
+            cp $file $dir_name/example/multicore/$solr_core/conf
+            echo "Copied $file into solr conf directory."
+        else
+            echo "${file} is not valid";
+            exit 1
+        fi
+      done
+    fi
+
+    # enable custom core
```
stof: This docker-compose setyp does not use custom core AFAICT, so this is also unused stuff

docker/solr/entrypoint.sh
```diff
@@ -0,0 +1,94 @@
...
+    wait_for_solr
+    cd ../../
+    echo "Started"
+}
+
+initialize_and_run() {
+	version='3.6.0'
+    dir_name="apache-solr-${version}"
+    dir_conf="conf/"
+    add_core $dir_name $dir_conf $SOLR_CORE "$SOLR_CONFS"
```
stof: this initialization could also be done at build time, as it only depends on the config files, which are copied in the image at build time.

Jean85:
Ok I've centralized anything else remained, it should be ok now.
Now we can focus on making the tests green.

Seldaek:
Please stop right here.. IMO this is a waste of time. I don't even know what you are trying to solve here, but I'm sure introducing docker is not the solution. More later when I have time, sorry for the bluntness but I don't want you to waste more time on it than you already did.

Jean85:
I'm trying to work on #808. Basically the CI is not running on all tests, and there are multiple ones that are broken, as they are right now.

Docker is just a reproducible way to run tests locally.

Seldaek:
@Jean85 I saw all that but sprinkling docker onto something is not the ultimate solution to all problems :p I have personally no interest in working with it and won't maintain this so I rather not have it in as it'll get outdated and end up confusing people more than anything.

Moving on to #808.

PR #810, opened 2017-09-13
Update to PHPUnit 6 & other dependencies (and mark PHP 7 as minimum required)

Jean85:
This sparked from https://github.com/composer/packagist/issues/808#issuecomment-328877332: this forces Travis to use the PHPUnit binary installed through Composer, and it updates it to version 6.

This means that it will work only under PHP 7+, and in the CI it's already like that. That's why I've updated the language requirement in the `composer.json`. I understand that that may be a BC from an OSS point of view, so let's discuss it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "exclude-from-classmap": [ "src/Packagist/WebBundle/Tests/" ]
     },
     "require": {
-        "php": ">=5.6",
+        "php": "^7.0",
         "symfony/symfony": "^2.8",
         "doctrine/orm": "^2.4",
         "doctrine/doctrine-bundle": "^1.2",
@@ -60,7 +60,7 @@
     "_comment": ["fos user bundle 2.0.0 tag needed"],
     "require-dev": {
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
-        "phpunit/phpunit": "^5.0"
+        "phpunit/phpunit": "^5.4.3|^6.0"
     },
     "scripts": {
         "post-install-cmd": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "d66a4039537323b09a560472fe93c122",
+    "content-hash": "0f14ccaf66d5a4125a1ff3e50c218312",
     "packages": [
         {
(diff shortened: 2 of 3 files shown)


Jean85:
I've had to update the Symfony packages to make it work. Now we're using the right PHPUnit binary and executing all the tests in Travis!

stof:
this needs to be rebased, because the composer.lock was modified in master

Jean85:
Done. I'm no longer seeing more tests executed here than on master, something as changed? Maybe 6 tests were deleted due to Solr being ditched?

Still, this PR at least updates a few packages and bumps minimum PHP version, is it still worth to merge?

stof:
Well, tests are running fine in master now, because Travis is forced to use a version of PHPunit compatible with the testsuite (was done in https://github.com/composer/packagist/commit/b1268f255172360e2da431d2817936048c3b09d7)
So this Pr is now about updating dependencies. Not sure @Seldaek wants to accept it from external contributors (the testsuite coverage is clearly not enough to be confident that this does not break things)

Jean85:
To be fair, I didn't change any constraint while upgrading, so if some package breaks something, we either have to fix it or to enforce a stricter constraint (and someone broke semver :-1: ).

Also: since I've noticed that 9ec4323d3779e51f2c3241e56b6e9fb0352e5d91 broke PHP 5.6 compat, I think he is :+1: on upgrading the language constraint :smile:

Seldaek:
Yeah I rather do updates myself. It doesn't work with current deps due to symfony lack of support for phpunit6 though so I'll hold on to this for a bit, but we should update eventually.

stof:
@Seldaek Symfony supports PHPUnit if you update it. Packagist uses 2.8.11, from 2016-09-07T02:04:25+00:00
And September 2016 is indeed before the PHPUnit 6 support. But latest 2.8 release is 2.8.27.
You would need to update the PHPUnit bridge to 3.3 in the dev deps though (support for PHPUnit 6 in the bridge has not been backported as it was a bigger change and you can use a different version anyway)

Jean85:
To be clear, in this PR I updated (using `--with-dependencies`) PHPUnit and all the Symfony-related packages:
 - symfony/symfony
 - symfony/swiftmailer-bundle
 - symfony/monolog-bundle
 - sensio/distribution-bundle
 - sensio/framework-extra-bundle
 - sensio/generator-bundle

Seldaek:
Thanks for the PR but as it was kinda out of date by now I just reapplied the changes myself over the latest lock file.

PR #812, opened 2017-09-13
Add active filter widget

julienbourdeau:
If you have a filter activated and you search for something else, you may not realize a filter is active.
I added the active filter widget at the top of the sidebar.

packagist-active-filters


Haroenv:
Great!

Seldaek:
Is it possible to show "type: " or "tag: " before the filter name? If not no biggie but would be a bit more clear.

julienbourdeau:
Yes good idea. I updated the PR accordingly.

packagist

PR #813, opened 2017-09-13
Improve Algolia relevancy

julienbourdeau:
I tweaked the ranking formula to improve relevancy. You can try it on the `packagist_prod_replica` index. (no reindexing required).

`popularity` is between 0 and 10 so it create [a lot of ties](https://www.algolia.com/doc/guides/ranking/ranking-formula/), that's why we can put it so high in the ranking formula.

We search first in the package name and then it tags. It helps finding templating engine when you search _"templating"_ for instance.

Then if every criteria are the same, we promote a package that is not deprecated compare to a deprecated one. We could even add the trendiness after that for instance.

Let me know what you think :)

---

Live testable version available here: https://piano-tuner-rhinoceros-60112.netlify.com

### DIFF
--- a/src/Packagist/WebBundle/Resources/config/algolia_settings.yml
+++ b/src/Packagist/WebBundle/Resources/config/algolia_settings.yml
@@ -1,22 +1,22 @@
 searchableAttributes:
-    - "name"
-    - "package_name"
-    - "package_organisation"
-    - "description"
-    - "tags"
+  - "unordered(package_name)"
+  - "unordered(tags)"
+  - "unordered(package_organisation)"
+  - "unordered(description)"
 
 ranking:
     - "typo"
     - "geo"
     - "words"
     - "filters"
+    - "desc(popularity)"
     - "attribute"
     - "proximity"
     - "exact"
     - "custom"
 
 customRanking:
-    - "desc(popularity)"
+    - "asc(abandoned)"
 
 attributesToHighlight:
     - "name"


Seldaek:
trendiness is basically very similar to popularity (`downloads in last 7 days` vs `downloads of last month + stars`), probably should drop it, the rest sounds good. I'll have a look a bit later. Thanks!

julienbourdeau:
Ok great! I didn't use trendiness so it's all good :)

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
```diff
@@ -1,22 +1,22 @@
 searchableAttributes:
-    - "name"
-    - "package_name"
-    - "package_organisation"
-    - "description"
-    - "tags"
+  - "unordered(package_name)"
+  - "unordered(tags)"
+  - "unordered(package_organisation)"
+  - "description"
```
redox: `description` should be `unordered` as well, no?

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
```diff
@@ -1,22 +1,22 @@
...
 
 ranking:
     - "typo"
     - "geo"
     - "words"
     - "filters"
+    - "desc(popularity)"
```
redox: Not sure about that one, making it above `attributes` will put very popular libraries with a match in the `description` to be above a "perfect" match in the `package_name` for instance.

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
julienbourdeau: Description is the whole `README.md` so I though ordered would make more sense. I will test and compare 👌

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
julienbourdeau: We wanted to do this to highlight package that uses a techno but aren't named after it. For example if you look for "websocket", we want Ratchet lib to come before any random package with _websocket_ in the name. Another example would be if you look for "templating", you'll find twig, smarty and mustache, which are probably what you're looking for.

Also popularity is only from 0 to 10 so it should create a lot of ties.

Maybe I can put a webpage up with these settings so everyone can try easily 💡

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
Seldaek: @julienbourdeau description only contains the "description" field from composer.json. The readme content is not indexed at the moment. Maybe we should add it though.

julienbourdeau: Oh my bad! I tested with ordered and unordered, I didn't find much differences on the tested queries. Although, since it's the composer.json description, I agree that unordered make more sense. I'll update.

redox: I think you didn't see any differences because of the `desc(popularity)` ranking criterion hijacking the ranking formula before `attributes`.

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
redox: > Also popularity is only from 0 to 10 so it should create a lot of ties.

That's a good thing, but I still think we want to "boost" popular objects only if they have a 9 or 10.

What about:
 - keeping `desc(popularity)` in the `customRanking`
 - creating a new `most_popular` boolean attribute in each record  with value `1` if popularity is 10 or 9, `0` otherwise
 - replacing this `desc(popularity)` middle-ranking criterion by `desc(most_popular)`
 - eventually moving this criterion at the very top of the ranking formula (before `typo`)

src/Packagist/WebBundle/Resources/config/algolia_settings.yml
```diff
@@ -1,22 +1,22 @@
 searchableAttributes:
-    - "name"
-    - "package_name"
-    - "package_organisation"
-    - "description"
-    - "tags"
+  - "unordered(package_name)"
+  - "unordered(tags)"
+  - "unordered(package_organisation)"
+  - "unordered(description)"
```
Seldaek: I had to append `unordered(name)` here otherwise a search for an exact package name doesn't even match, which sounds insane to me.. say a search for e.g. symfony/monolog-bundle should return that one package at the very least.

PR #814, opened 2017-09-14
Add HiDPI version of maintainer gravatars

ocean90:
Currently avatars of maintainers are only available in 48x48 pixels which makes them blurry on HiDPI screens. We can fix that by adding the `srcset` attribute.

Before/After:
before-after


Seldaek:
Yup looks fine, thanks!

Issue #815, opened 2017-09-19
Packagist does not show my package in search results


mbajda:
I've recently (yesterday) added my package to packagist. The package name is:

> mbajda/yahoo-weather

I can use a 'require' key in composer.json to add a requirement for this package
```
"require": {
    "mbajda/yahoo-weather": "dev-master"
}
```
and the downloading works properly. What does not work, is that when I try to search for that package in the Packagist search box, the package cannot be found - I don't see it in the search results.

localheinz:
Can confirm this issue!

🤓

mrcnpdlk:
I also have this problem. My new package is not searchable after adding to the packegist.

localheinz:
Actually, I have to revise my comment - not only can I confirm that the package `mbajda.yahoo-weather` cannot be found, but also that a package I submitted myself yesterday cannot be found.

Shows up on https://packagist.org/users/localheinz/packages/:

![screen shot 2017-09-20 at 10 51 34](https://user-images.githubusercontent.com/605483/30635121-bf569cd4-9df1-11e7-967c-b1429f318f76.png)

Does not show up in search results at https://packagist.org/?q=factory%20muffin%20definition.

@mbajda 

Your package also shows up under the list of your packages, though: https://packagist.org/users/mbajda/packages/

localheinz:
Possibly related to #805.

localheinz:
Not sure if it is related, but if I search for `localheinz/test-util`, https://packagist.org/?q=localheinz%20test%20util shows **0** downloads>

![screen shot 2017-09-20 at 10 57 56](https://user-images.githubusercontent.com/605483/30635403-96948698-9df2-11e7-86d5-65914be04ebc.png)

However, I can confirm that I have downloaded it a couple of times, which can also be seen at https://packagist.org/packages/localheinz/test-util:

![screen shot 2017-09-20 at 10 58 54](https://user-images.githubusercontent.com/605483/30635457-bb9966ac-9df2-11e7-998b-9b99a19a312c.png)

Maybe it's just the indexing that is behind or something?

stof:
I suspect that the indexing was failing due to the issue fixed in https://github.com/composer/packagist/pull/820#issuecomment-330901506

Seldaek:
Fixed now, sorry about that.

localheinz:
Thanks a ton, @Seldaek!

PR #820, opened 2017-09-20
Update IndexPackagesCommand.php

browseman:
Just spot one typo issue:
./app/console packagist:index --no-debug --env=prod

[Symfony\Component\Console\Exception\CommandNotFoundException]
Command "packagist:index" is not defined.
Did you mean one of these?
packcagist:index
packagist:dump
packagist:update
packagist:tokens:generate
packagist:stats:compile
packagist:clear:versions

### DIFF
--- a/src/Packagist/WebBundle/Command/IndexPackagesCommand.php
+++ b/src/Packagist/WebBundle/Command/IndexPackagesCommand.php
@@ -31,7 +31,7 @@ class IndexPackagesCommand extends ContainerAwareCommand
     protected function configure()
     {
         $this
-            ->setName('packcagist:index')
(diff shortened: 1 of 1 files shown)


alcohol:
Thanks :-)

stof:
btw, this might be the cause for https://github.com/composer/packagist/issues/815 (if the cron fails since the migration to Algolia due to this type, it is logical that new packages are not visible in the search)

alcohol:
@Seldaek can you confirm what @stof suggested?

Seldaek:
Yeah that sounds like that was the reason, oops :) I'll try and merge some of these search changes later and deploy latest state.

PR #828, opened 2017-09-27
Add tags link backward compat

julienbourdeau:
I added backward compatibility for link like https://packagist.org/search/tags=templating

In the meantime, I switch tags to checkbox link so they can be combined and I added search for tags.

sep-27-2017 18-43-18

### DIFF
--- a/src/Packagist/WebBundle/Resources/views/layout.html.twig
+++ b/src/Packagist/WebBundle/Resources/views/layout.html.twig
@@ -204,6 +204,7 @@
 
         <script>
             var algoliaConfig = {{ algolia|json_encode|raw }};
+            algoliaConfig['tags'] = '{{ tags|default(false) }}';
         </script>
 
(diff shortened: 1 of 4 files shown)


src/Packagist/WebBundle/Resources/views/layout.html.twig
```diff
@@ -204,6 +204,7 @@
 
         <script>
             var algoliaConfig = {{ algolia|json_encode|raw }};
+            algoliaConfig['tags'] = '{{ tags|default(false) }}';
```
stof: why not setting `tags` in the `algolia` structure before encoding it ?

julienbourdeau: I agree it would be better but I'm not sure how to do it. Right now algolia is defined in the [config.yml](https://github.com/composer/packagist/blob/master/app/config/config.yml#L38-L41) as a twig global. Any advice?

Seldaek:
Thanks, fixed a couple things and deployed :)

PR #831, opened 2017-10-09
Fix webhook source identification

iammattcoleman:
GitHub and GitLab both send `$payload['repository']['url']`, but only GitLab's payload contains `$payload['project']`.

This ensures that the proper URLs are used from webhook payloads when GitLab is configured with nonstandard URLs (for example, when developing integrations in a local instance of gitlab-development-kit).


iammattcoleman:
@Seldaek Hey, I noticed you're the last person to modify this file. Can you give it a review? Thanks!

Issue #849, opened 2017-12-21
Don't focus search bar on detail pages


lornajane:
I have difficulty using the packagist website because it focuses the search bar on every page.  I need accessibility tools to navigate a page or click on anything, and they are keyboard-based.  As soon as I find the page I need and try to do anything, ooops I've typed into a search box in the header and navigated to another set of search results and now need to start again.

Some interesting references and discussion on the issue of autofocussing fields for accessibility here https://stackoverflow.com/questions/2180645/is-automatically-assigning-focus-bad-for-accessibility/2180679#2180679

I'm happy to try to fix it if this seems like something that might be accepted.

Seldaek:
Yes I can see that on the detail pages it probably doesn't make lots of sense. This is done in search.js the algolia init contains `autofocus: true`. I guess this needs to be made dependent on a global variable defined in the main layout page or something. I can look at that but not for a couple weeks I guess, but feel free to have a go at it if you have time :)

PR #854, opened 2017-12-29
Focus the search box on the main page only

lornajane:
This is a suggested solution for #849 where the focus on the search bar on pages where you probably want to do something other than search makes the site confusing/hard particularly for those with accessibility issues.  If there are other pages where the search bar _should_ be focussed, perhaps we can extend this patch to include those but I hope this works as a starting point.


Seldaek:
Thanks

PR #867, opened 2018-02-02
phpunit up to 6, add php 7.1

andreybolonin:


.travis.yml
```diff
@@ -15,6 +15,7 @@ services:
 
 php:
   - 7.0
+  - 7.1
```
stof: Having multiple jobs on Travis does not make sense. We only need to run tests for the version used by packagist.org

app/autoload.php
```diff
@@ -2,6 +2,19 @@
...
+if (!class_exists('\PHPUnit\Framework\TestCase', true)) {
+    class_alias('\PHPUnit_Framework_TestCase', '\PHPUnit\Framework\TestCase');
+} elseif (!class_exists('\PHPUnit_Framework_TestCase', true)) {
+    class_alias('\PHPUnit\Framework\TestCase', '\PHPUnit_Framework_TestCase');
+}
+
+if (!class_exists('\PHPUnit\Util\Test', true)) {
+    class_alias('\PHPUnit_Util_Test', '\PHPUnit\Util\Test');
+} elseif (!class_exists('\PHPUnit_Util_Test', true)) {
+    class_alias('\PHPUnit\Util\Test', '\PHPUnit_Util_Test');
```
stof: Creating such class aliases is a bad idea IMO. We should just make out tests use the right version (and being sure we use a `symfony/phpunit-bridge` version compatible with PHPUnit 6)

PR #873, opened 2018-02-07
Removed unused variable

sadortun:
>>>
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
  You have requested a non-existent service "solarium.client".
>>>

### DIFF
--- a/src/Packagist/WebBundle/Command/CompileStatsCommand.php
+++ b/src/Packagist/WebBundle/Command/CompileStatsCommand.php
@@ -117,7 +117,7 @@ class CompileStatsCommand extends ContainerAwareCommand
             $ids[] = $row['id'];
         }
 
-         if ($verbose) {
+        if ($verbose) {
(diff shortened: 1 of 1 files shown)


src/Packagist/WebBundle/Command/CompileStatsCommand.php
```diff
@@ -117,10 +117,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $ids[] = $row['id'];
         }
 
-        // add downloads from the last 7 days to the solr index
-        $solarium = $this->getContainer()->get('solarium.client');
-
-        if ($verbose) {
+         if ($verbose) {
```
stof: changing this indentation looks wrong to me

sadortun: @stof You are right, that was a mistake ! Fixed.

PR #885, opened 2018-02-27
Allow all github user content domains for images in CSP

### LINKED ISSUE #879: resources not loaded due to content-security policy
It's not a big issue, but images linked in GitHub readme files don't load giving a console error 

```
rfc1867:1 Refused to load the image 'https://raw.github.com/Lewiscowles1986/rfc1867/HEAD/diagram/class-relationship.png?raw=true' because it violates the following Content Security Policy directive: "img-src 'self' https://www.gravatar.com/ https://camo.githubusercontent.com/ https://ssl.google-analytics.com/ http://www.google-analytics.com/".
```

stof:
Github does not only use the camo subdomain. Images which are already pointing to a githubusercontent subdomain are not rewritten in readme files.

Closes #879

### DIFF
--- a/app/config/config_prod.yml
+++ b/app/config/config_prod.yml
@@ -86,7 +86,7 @@ nelmio_security:
             img-src:
                 - 'self'
                 - 'https://www.gravatar.com/'
-                - 'https://camo.githubusercontent.com/'
+                - 'https://*.githubusercontent.com/'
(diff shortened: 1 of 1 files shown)


stof:
Closing this in favor of #891 as the raw github link does not use a githubusercontent domain in the rendering.

PR #890, opened 2018-03-07
Fix the rendering of the API token

stof:
Before:
packagist_token_before

after:
packagist_token_after

Note that the design was already fine after revealing the token. (but it stays fine).


Seldaek:
Thanks for all the fixes!

PR #892, opened 2018-03-08
Update requirements in the readme

stof:
The project currently uses Doctrine ORM 2.5, which generates paginated queries incompatible with the strict mode of MySQL. Compatibility with strict mode would require migrating to the ORM 2.6 first.

Documenting this in the readme will avoid WTF moments for contributors running MySQL locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This project is not meant for re-use. It is open source to make it easy to contr
 Requirements
 ------------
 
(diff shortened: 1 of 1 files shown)


stof:
Note that the offending mode is `ONLY_FULL_GROUP_BY`

Seldaek:
We are now on doctrine/orm 2.6

PR #896, opened 2018-03-16
Fix: daily stats calculation

glaubinix:
`$average` value is always lower case.

The calculation still was correct but it never executed the optimized code branch where all days get fetch with one redis call.


Seldaek:
Thanks

PR #925, opened 2018-07-12
Fix time documentation

nadar:
Actually i am not sure what time is returned, but its definitely no the time of the last release (which would be very helpful indeed, but its not). So i assume its the date when the repo is registered on packagist.


nadar:
The proof from the example in the docs: 

https://packagist.org/packages/monolog/monolog.json

package time is: "2011-09-27T00:35:19+00:00"
time of latest release (1.23.0) is: "2017-06-19T01:22:40+00:00"

Seldaek:
It is indeed the creation time, thanks

stof:
yeah, the time of releases is found inside `versions` (as each version has its own time)

PR #942, opened 2018-08-30
Update allowed images in CSP

### LINKED ISSUE #941: Allow images from shields.io, gitlab.style.ci and poser.pugx.org in CSP
Hey can we add following urls to the CSP?
- https://img.shields.io
- https://gitlab.styleci.io
- https://poser.pugx.org

Currently none of my badges are displayed.

image

emanuelmutschlechner:
Closes #941

### DIFF
--- a/app/config/config_prod.yml
+++ b/app/config/config_prod.yml
@@ -102,6 +102,9 @@ nelmio_security:
                 - 'https://gitlab.com/'
                 - 'https://www.google-analytics.com/'
                 - 'http://www.google-analytics.com/'
+                - 'https://img.shields.io/'
+                - 'https://gitlab.styleci.io/'
+                - 'https://poser.pugx.org/'
             style-src:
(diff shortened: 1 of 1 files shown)


stof:
I think a better solution should be found here.

For Github repositories, almost all images are proxies through their camo proxy (allowing to ensure HTTPS and maybe other stuff) and so Packagist could whitelist it.
Gitlab does not perform a similar proxying, so images could come from anywhere for Gitlab repos.

@Seldaek should the img-src allow any https URL on Packagist for that ?

emanuelmutschlechner:
@stof There's a two year old ticket over at GitLab. I don't think they'll add camo proxyies in foreseeable future.

https://gitlab.com/gitlab-org/gitlab-ce/issues/13166

Seldaek:
I agree with @stof allowing everything probably makes more sense here..

PR #956, opened 2018-10-16
Add headers to improve maintainers visibility

wheatleyjj:
A number of my colleagues were unable to find how to add a maintainer after scouring aggressively as the button isn't the most obvious in its purpose.

It's mostly because when a package has one maintainer (the owner), it looks like the avatar of the package owner, not a collection of maintainers.

This PR updates the view from this:

screen shot 2018-10-16 at 11 18 41 

To this:

screen shot 2018-10-16 at 11 18 31


Seldaek:
Thanks :) Definitely quite a few people not finding this.

PR #968, opened 2018-11-13
do not use outdated PHP version for example

OskarStark:
This should be changed in january to sth. like this:

```diff
- "php": "^5.6.37 || ^7.0",
+ "php": "^7.0.31 || ^7.1",
```


Seldaek:
Sounds good, thanks

OskarStark:
Shall I provide another PR which could be merged when 5.6 has EOL?

Seldaek:
Sure, although I'd say we can also just make it `"^7.1"` at that point ;)

PR #979, opened 2018-12-14
Add PHP test server router script support

yookoala:
Detects asset requiest uri, when called, and return false. Allows `web/app_dev.php` to function as a PHP test server router script.

With this merged, you may run Packagist on PHP Test Server with this command:
```
php -S localhost:8000 -t ./web ./web/app_dev.php
```

An improvement for developers who wants to improve Packagist.

Reference:
http://php.net/manual/en/features.commandline.webserver.php (Please see **Example 3**)

### DIFF
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -10,6 +10,15 @@ use Symfony\Component\Debug\Debug;
 // read http://symfony.com/doc/current/book/installation.html#configuration-and-setup for more information
 //umask(0000);
 
+// When used directly as a php router script,
+// allow all assets to be served normally.
+if (preg_match(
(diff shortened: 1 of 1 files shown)


yookoala:
Updated the extension whitelist with the actual assets that is found in the `/web` folder.

Seldaek:
Thanks but I believe using https://symfony.com/doc/current/setup/built_in_web_server.html would probably make more sense (once symfony is updated to 3.x or 4.x at least). I don't really want to merge this as I am not sure it won't have any side effects on prod.

yookoala:
Understood.

PR #980, opened 2018-12-16
Replaced rawgit.com urls with combinatronics.com urls.

c0mbinatr0nics:
Combinatronics.com is a drop in replacement for rawgit.com which is EOL.


julienbourdeau:
I don't see any change in your PR, as if the commit was empty. Is it only me? 🤔

PR #996, opened 2019-03-12
Fix monthly stats averages

BenMorel:
This PR fixes the calculations of daily downloads when averaging monthly.

Before| Now
------------ | -------------
The total monthly downloads were divided by `30.41` | The total monthly downloads are divided by the actual number of days in the month, `28` to `31`
The whole month of the `from` date was taken into account | The `from` date is actually respected
**Example**: `from=2018-01-29` yielded an interval `2018-01-01` → `2018-01-31` for the first month | `from=2018-01-29` yields `2018-01-29` → `2018-01-31`
The whole month of the `to` date was taken into account | The `to` date is actually respected
**Example**: `to=2019-03-12` yielded an interval `2019-03-01` → `2019-03-31` for the last month | `to=2019-03-12` yields `2019-03-01` → `2019-03-12`
The total number of downloads for the start and end month was divided by `30.41` | The total number of downloads **for the given interval** is divided by **the actual number of days in the interval**, `1` to `31`

### Why this change?

It will better reflect the actual daily downloads of the package, and will avoid sharp drops in the chart at the beginning of each month:

Monthly stat drop

### DIFF
--- a/src/Packagist/WebBundle/Controller/PackageController.php
+++ b/src/Packagist/WebBundle/Controller/PackageController.php
@@ -984,7 +984,7 @@ class PackageController extends Controller
             foreach ($values as $valueKey) {
                 $value += $dlData[$valueKey] ?? 0;
             }
-            $datePoints[$label] = $value;
+            $datePoints[$label] = ceil($value / count($datePoints[$label]));
         }
 
         $datePoints = array(
@@ -993,17 +993,6 @@ class PackageController extends Controller
         );
 
         $datePoints['average'] = $average;
-
-        if ($average !== 'daily') {
-            $dividers = [
-                'monthly' => 30.41,
-                'weekly' => 7,
-            ];
-            $divider = $dividers[$average];
-            $datePoints['values'] = array_map(function ($val) use ($divider) {
-                return ceil($val / $divider);
-            }, $datePoints['values']);
-        }
 
         if (empty($datePoints['labels']) && empty($datePoints['values'])) {
             $datePoints['labels'][] = date('Y-m-d');
@@ -1101,13 +1090,15 @@ class PackageController extends Controller
         $dateKey = 'Ymd';
         $dateFormat = $average === 'monthly' ? 'Y-m' : 'Y-m-d';
         $dateJump = '+1day';
-        if ($average === 'monthly') {
-            $from = new DateTimeImmutable('first day of ' . $from->format('Y-m'));
-            $to = new DateTimeImmutable('last day of ' . $to->format('Y-m'));
-        }
 
         $nextDataPointLabel = $from->format($dateFormat);
-        $nextDataPoint = $from->modify($interval);
+
+        if ($average === 'monthly') {
+            $nextDataPoint = new DateTimeImmutable('first day of ' . $from->format('Y-m'));
+            $nextDataPoint = $nextDataPoint->modify($interval);
+        } else {
+            $nextDataPoint = $from->modify($interval);
+        }
 
         $datePoints = [];
         while ($from <= $to) {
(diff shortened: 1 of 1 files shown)


src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -984,7 +984,7 @@ public function overallStatsAction(Request $req, Package $package, Version $vers
             foreach ($values as $valueKey) {
                 $value += $dlData[$valueKey] ?? 0;
             }
-            $datePoints[$label] = $value;
+            $datePoints[$label] = ceil($value / count($datePoints[$label]));
```
stof: I would use `count($values)` here

BenMorel: Good catch! Fixed 👍

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -1101,13 +1090,15 @@ private function createDatePoints(DateTimeImmutable $from, DateTimeImmutable $to
         $dateKey = 'Ymd';
         $dateFormat = $average === 'monthly' ? 'Y-m' : 'Y-m-d';
         $dateJump = '+1day';
-        if ($average === 'monthly') {
-            $from = new DateTimeImmutable('first day of ' . $from->format('Y-m'));
-            $to = new DateTimeImmutable('last day of ' . $to->format('Y-m'));
-        }
 
         $nextDataPointLabel = $from->format($dateFormat);
```
BenMorel: BTW, I think this should be named `$currentDataPointLabel`.

BenMorel:
Any feedback here?

BenMorel:
Bump.

BenMorel:
Thank you! 👍

alcohol:
No, thank you :-)

BenMorel:
BTW, how long does it take before changes are deployed to the Packagist website?

alcohol:
I cannot say for sure, since that is a manual process and I am not capable of this myself.

Seldaek:
Thanks, looking good, it's deployed now.

BenMorel:
Great thank you, looking good indeed now:

image

no more sharp drops on current month 👍

Seldaek:
@BenMorel looks like this might have a regression for packages where the selected period is weekly and not monthly https://twitter.com/AlessandroLai/status/1152197439442567169 - would you mind looking into it?

BenMorel:
Crap, I'll try to look into it today.

BenMorel:
At first glance it looks like it's working as expected :/
Let's take [sentry/sentry](https://packagist.org/packages/sentry/sentry/stats) as an example.

What's causing a spike is that the average for `2019-07-17` is calculated on just 1 day (Packagist shows stats with a 2 days lag, so the last day in the stats is 17 July today). So the number 35.1k you see in the chart is the right number, as you can see in the daily stats:

https://packagist.org/packages/sentry/sentry/stats/all.json?from=2019-07-10&average=daily

Previous weeks however, are all calculated on 7 days, **including saturdays & sundays** where the number of downloads drops sharply (~5k vs 30k), so the average is much less than when a partial week is calculated on week days only.

To summarize:

- it's working as it should
- maybe this is not what we want?

Some suggestions:

- leave it as it is (even though it looks weird, it *is* the actual number of daily downloads in the current, **partial** week)
- apply some kind of smoothing factor calculated on a number of previous weeks, depending on the week days involved in the current week
- shift the start-of-week day to ensure that the week always ends on the same day as the last requested day in the stats, so that we're always dividing by 7

I'm torn between the first and third suggestions myself.

naderman:
Another suggestion: make the last part of the line which is an extrapolation dotted, so it's clear it's still going to change?

Seldaek:
@BenMorel if you can easily patch it to do the third option, it sounds like the best way to me..  but leaving it as is is also reasonable IMO.

@naderman's option sounds ok too but on  some long periods like that graph of sentry you linked, a single datapoint has very little visual space, so I don't think the dotted line will be very visible.

Issue #998, opened 2019-03-29
Popular section pagination, returns the same results


0mars:
https://packagist.org/explore/popular?page=2

is same as page 1, the pagination links are okay, but the results are the same for every page

Seldaek:
Fixed in https://github.com/composer/packagist/commit/105f8faeaec14d3134756676955199c6debdb2d5

PR #999, opened 2019-04-02
Fix popular section

pgrimaud:
This will fix the issue : https://github.com/composer/packagist/issues/998

```php
echo min(1, 2);
1
// min will return the lowest value
```
Actual : ?page=2 will show data from ?page=1

but with `max()`

```php
echo max(1, 2);
2
// max will return the highest value
```
With this fix : ?page=2 will show data from ?page=2


Seldaek:
Sorry missed this and did the fix in https://github.com/composer/packagist/commit/105f8faeaec14d3134756676955199c6debdb2d5 yesterday, thanks for the PR though :)

PR #1015, opened 2019-06-13
Fix invalid index error

sadortun:


alcohol:
This normally results in a notice. I think it all depends on how your error reporting is setup and whether or not you are converting notices and such into errors/exceptions or not.

sadortun:
Notice or not, you are still accessing an invalid index, which is bad !

stof:
the question is, in which case is this index not set ?

stof:
OK, `app/config/defaults.yml` sets the parameter to an empty array. So a local setup would indeed not have these keys.

sadortun:
@stof, I would say in the default install case, since this parameter is, to my knowledge, not described anywhere in the install docs, or `parameters.yaml.dist` ...

src/Packagist/WebBundle/HealthCheck/MetadataDirCheck.php
```diff
@@ -20,7 +20,7 @@ public function __construct(array $awsMetadata)
 
     public static function isMetadataStoreMounted(array $awsMeta): bool
     {
-        if ($awsMeta['primary'] && $awsMeta['has_instance_store']) {
+        if (isset($awsMeta['primary']) && isset($awsMeta['has_instance_store'])) {
```
stof: this is wrong. It won't check the boolean values of these params, which will break the prod (a secondary node would get handled as primary, as `isset($awsMeta['primary'])` would return `true` for `['primary' => false]`

sadortun: You are right !

stof: what you should do, to be consistent with other places using `$awsMeta` in the codebase is to change that to `!empty($awsMeta) && $awsMeta['primary'] && ...`

sadortun:
Thanks @Seldaek, sorry for the delay !

donaldinou:
Sorry to come back on this issue but I think that the function "isMetadataStoreMounted" should have the `!empty($awsMeta)` verification because it is call on the `dump` method from the `SymlinkDumper` and so used with the associated command.

Seldaek:
@donaldinou done in 8720c3f

PR #1016, opened 2019-06-13
Add missing step

sadortun:


alcohol:
Not really a missing step as there are several ways this could be setup (via webhooks for example as well).

stof:
Well, the fact that this should run is indeed missing.

PR #1017, opened 2019-06-22
Improve steps for setup of a dev environment

jackbentley:
I wanted to contribute to the code base, but found the readme lacking a little on a getting started guide.

I thought it would be helpful to update the readme with a bit more detailed instructions on getting setup for development.

Additionally, I've done a `composer update` and I've added the `incenteev/composer-parameter-handler` package that Symfony Standard Edition (v3.4) ships with to aid in a simplier setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,24 +3,43 @@ Packagist
 
 Package Repository Website for Composer, see the [about page](https://packagist.org/about) on [packagist.org](https://packagist.org/) for more.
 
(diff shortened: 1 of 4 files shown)


alcohol:
While more detailed instructions are generally welcome (though we really do not want to encourage users to run their own instance), adding the parameter handler is a significant change which I am not sure we want or need.

colinodell:
I personally found the instructions in here to be very helpful in setting up my local environment.  I mean sure, it's not hard to figure out if you've worked on Symfony projects before, but this still helps developers who want to spin up a local dev environment in order to contribute to the project.

Seldaek:
Thanks

PR #1021, opened 2019-07-03
add mirror operation by Alibaba Cloud holdings.

xiaohuilam:


xiaohuilam:
@GrahamCampbell
Because the cross border internet control ACT and GFW of China, not only the access non-mandarin site will be slow from Mainland China, but also access mandarin site from countries excepts China.

I’m sure it works, and it is presented by largest commercial shopping company of China: Alibaba, its uptime should be stable.

And by the way, if not Chinese user, use this composer mirror might slow down the build speed.

PR #1022, opened 2019-07-10
Added endpoint to fetch total downloads number.

frankdejonge:
Context: https://twitter.com/seldaek/status/1148880016043728900

I've added an endpoint to fetch statistics, for now I've only implemented the one for total downloads ever, which I need. More could be added later if needed.


Seldaek:
Thanks, deployed now

PR #1023, opened 2019-07-14
Return failed searchAction calls as early as possible.

hopeseekr:
These two error paths can (and thus should) be called at the very beginning of the method, thus saving precious computation time.


Seldaek:
As usual, simple things are never as simple as they appear ;)

Got a few reports of issues with search (e.g. https://twitter.com/RageZBla/status/1152856680536788992) - fixed by https://github.com/composer/packagist/commit/a0abfaba097c10bfad59d83b67ec5fd42bf0a8c1

Issue #1020, opened 2019-07-01
[RFC] Two-Factor Authentication


colinodell:
Would you consider implementing two-factor authentication to help protect accounts of package publishers?  I don't have any specific implementation in mind - I just know that other package manager sites (like npm) allow 2FA and it seems like a good security measure.

hopeseekr:
I think I'll go ahead and implement the PR for this, at least using The Time-based One-Time Password algorithm (TOTP), RFC 6238, basically the same thing that's compatible with Google 2FA.

Once I begin in earnest, I'll update this issue. Of course, it'll be in my fork without any guarantee this project will actually merge it.

alcohol:
I think any reasonable PR which introduces 2FA would be welcome. Anything that is compatible with apps such as Google Authenticator (or Authy, preferably) would be good. Definitely not mobile text message based or such as we do not have the resources to work with that.

stof:
a TOTP implementation would support both Google Authenticator and Authy automatically, as they are both TOTP apps (and would also support other TOTP apps)

colinodell:
I've begun working on a TOTP implementation and hope to have that ready for review soon.

Would there be any interest in also allowing email-based 2FA and/or backup codes?

stof:
IMO, backup codes are indeed useful, to prevent locking someone off their account entirely if their phone with the TOTP app gets stolen (and they don't use a TOTP app supporting backup of the config like Authy does).

alcohol:
I don't see why not. QR code should not be leading functionally speaking, it should be an additional feature. As you said, not all apps support QR code as input.

colinodell:
Good call!  I have updated my PR to expose the key: https://github.com/composer/packagist/pull/1031#issuecomment-540543058

TheArKaID:
Umm guys, I lost my 2FA and also my backup code so that I cannot pass the 2FA screen. Is there any other way to log in ? Something like `Reset Password` that sent to the email ?

Seldaek:
@TheArKaID I disabled it for you

PR #1031, opened 2019-10-02
TOTP implementation for 2FA

colinodell:
Implements #1020 using TOTP codes

## Overview

A new section has been added to the user area:

image

Clicking the button link takes you to this new page:

image

Using a compatible app, the user must scan the QR code and enter in a valid TOTP code to prove they've got it working.  Once they submit a valid code 2FA will be active on their account:

image

The user is also given a single-use backup code - more on that later.

When a user with a 2FA-enabled account logs in, we redirect them to this new page:

image

They must enter a valid TOTP code from their authenticator in order to continue and be fully logged-in.  They can also request that the current browser be remembered for 30 days so they aren't prompted to constantly re-enter TOTP codes from the current computer.

If the user loses access to their authenticator, they can instead use that backup code to gain access.  Using this code will cause 2FA to become disabled until they set it up again:

image

Users can also disable 2FA manually if they wish.  From the main 2FA status in their account area, they simply click that giant red button which takes them to a confirmation page:

image

Clicking the red button will disable 2FA:

image

## Notes

A few notes about this implementation:

 - This 2FA functionality works with OAuth accounts too!
 - I've added very basic rate-limiting with Redis - you basically get locked out for 15 minutes after entering 5 incorrect codes
 - Users are sent an email notification whenever 2FA is enabled or disabled
 - I added new `mobile` icon to represent 2FA in the menu
 - The bundle I'm using does offer the ability to use email-based codes, but I have not implemented that

I'll add some additional notes inline with the code

Happy #Hacktoberfest!


app/config/config.yml
```diff
@@ -135,6 +135,25 @@ hwi_oauth:
...
+
+    backup_codes:
+        enabled: true
+        manager: Packagist\WebBundle\Security\TwoFactorAuthManager
+
+    totp:
+        enabled: true
+        server_name: '%packagist_host%'
+        issuer: Packagist
+        window: 1
```
colinodell: A new code is generated every 30 seconds.  A `window` of `1` will allow three codes to be accepted:

 - The current code
 - The code generated 30 seconds before it
 - The next code that will be generated

This allows some flexibility in case the server clock and mobile clock are slightly out-of-sync, or the user needs a few extra seconds to enter in their code.

src/Packagist/WebBundle/Controller/UserController.php
```diff
@@ -270,6 +274,92 @@ public function deleteFavoriteAction(User $user, Package $package)
...
+     * @Route("/users/{name}/2fa/", name="user_2fa_configure", methods={"GET"})
+     * @ParamConverter("user", options={"mapping": {"name": "username"}})
+     */
+    public function configureTwoFactorAuthAction(User $user)
+    {
+        if (!($this->isGranted('ROLE_DISABLE_2FA') || $user->getId() === $this->getUser()->getId())) {
+            throw new AccessDeniedException('You cannot change this user\'s two-factor authentication settings');
+        }
+
+        if ($backupCode = $this->get('session')->get('backup_code')) {
```
colinodell: I briefly considered using a flash message for this, but I wanted special formatting and therefore opted to just stash the code in normal session storage and manually remove it once we read it.

src/Packagist/WebBundle/Controller/UserController.php
```diff
@@ -270,6 +274,92 @@ public function deleteFavoriteAction(User $user, Package $package)
...
+        // Temporarily store this code on the user, as we'll need it there to generate the
+        // QR code and to check the confirmation code.  We won't actually save this change
+        // until we've confirmed the code
+        $user->setTotpSecret($enableRequest->getSecret());
+
+        if ($form->isSubmitted()) {
+            // Validate the code using the secret that was submitted in the form
+            if (!$authenticator->checkCode($user, $enableRequest->getCode())) {
+                $form->get('code')->addError(new FormError('Invalid authenticator code'));
+            }
```
colinodell: These three lines were significantly less code than trying to build a custom validator, so I hope you don't mind me taking this shortcut.

src/Packagist/WebBundle/Entity/User.php
```diff
@@ -318,4 +334,44 @@ public function getGravatarUrl()
...
+    {
+        return new TotpConfiguration($this->totpSecret, TotpConfiguration::ALGORITHM_SHA1, 30, 6);
+    }
+
+    public function isBackupCode(string $code): bool
+    {
+        return $this->backupCode !== null && hash_equals($this->backupCode, $code);
+    }
+
+    public function invalidateBackupCode(string $code): void
```
colinodell: The interface assumes the user may have multiple codes, but we're only using a single code here.

src/Packagist/WebBundle/Resources/public/font/config.json
```diff
@@ -0,0 +1,178 @@
+{
```
colinodell: I had to manually look up each icon that was used to create the custom font, which took a while, so I figured why not commit this so it's easier to make future changes?

src/Packagist/WebBundle/Security/TwoFactorAuthRateLimiter.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function onAuthFailure(TwoFactorAuthenticationEvent $event)
+    {
+        $key = '2fa-failures:'.$event->getToken()->getUsername();
+
+        $this->redis->multi();
+        $this->redis->incr($key);
+        $this->redis->expire($key, self::RATE_LIMIT_DURATION * 60);
+        $this->redis->exec();
```
colinodell: We could use something more advanced like a sliding window but I figured this was good enough for now.

src/Packagist/WebBundle/Resources/views/user/enable_two_factor_auth.html.twig
```diff
@@ -0,0 +1,66 @@
...
+            <p>
+                To enable two-factor authentication, you'll need a mobile app that supports TOTP such as
+                <a href="https://authy.com/download/">Authy</a>
+                or
+                <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2">Google Authenticator</a>.
+            </p>
+
+            <p><small>(This is not an exhaustive list of compatible apps. Packagist does not endorse or recommend one application over another.)</small></p>
+
+            <img src="{{ qr_code_data_uri(provisioningUri|raw) }}" height="200" class="pull-right" />
```
stof: the `raw` filter seems to be in the wrong place. It won't have any effect on the function argument as that's not a place applying auto-escaping. I suggest removing it.

src/Packagist/WebBundle/Resources/views/user/enable_two_factor_auth.html.twig
```diff
@@ -0,0 +1,66 @@
+{% extends "PackagistWebBundle:user:packages.html.twig" %}
+
+{% import "PackagistWebBundle::macros.html.twig" as macros %}
+
+{% set isActualUser = app.user and app.user.username is same as(user.username) %}
+
+{% block content %}
+    <h2 class="title">
+        {{ user.username }}
+        <small>
```
stof: the `<small>` tag should be inside the `if` to avoid rendering an empty tag.

src/Packagist/WebBundle/Resources/views/user/enable_two_factor_auth.html.twig
```diff
@@ -0,0 +1,66 @@
+{% extends "PackagistWebBundle:user:packages.html.twig" %}
+
+{% import "PackagistWebBundle::macros.html.twig" as macros %}
+
+{% set isActualUser = app.user and app.user.username is same as(user.username) %}
```
stof: I suggest moving this inside the block (as you don't need to share it between blocks). This way, you don't need to hope that the parent template does not override this variable name before rendering the block.

src/Packagist/WebBundle/Controller/UserController.php
stof: the retrieval (and cleaning) of the backup code should be done only when looking at the page for yourselves IMO.

src/Packagist/WebBundle/Resources/views/user/enable_two_factor_auth.html.twig
colinodell: Whoops, good catch!  This stray filter has been removed.

src/Packagist/WebBundle/Resources/views/user/enable_two_factor_auth.html.twig
colinodell: This code was copied as-is from existing templates in that same folder.  I've fixed the instances in my templates and will submit a separate PR to address the others :)

src/Packagist/WebBundle/Resources/views/user/enable_two_factor_auth.html.twig
colinodell: This code was copied as-is from existing templates in that same folder.  I've fixed the instances in my templates and will submit a separate PR to address the others :)

src/Packagist/WebBundle/Controller/UserController.php
colinodell: Fixed and simplified to:

```php
if ($user->getId() === $this->getUser()->getId()) {
    $backupCode = $this->get('session')->remove('backup_code');
}
```

See commit https://github.com/composer/packagist/pull/1031/commits/0b4272099dc205665164eda71f4828482b5cc82fhttps://github.com/composer/packagist/pull/1031/commits/0b4272099dc205665164eda71f4828482b5cc82f

colinodell:
Thanks for the review, @stof!  I have made the suggested edits.  For the sake of tracking them, I have not squashed these into my earlier commits yet but can certainly do so to clean up the history before merging.

colinodell:
I have added commit 00ca834 which exposes the secret key to the user during setup. This will allow people without QR code scanners to set up 2FA:

image

Seldaek:
Just a quick flow review based on the screenshots (thanks for that, really helpful), it looks great overall, I'd just think having a proper intermediate page to show the confirmation that it's been enabled and show the backup code there prominently would be good. Having it only in the flash message confirmation can be very easily glanced over IMO.

colinodell:
I found time this evening to implement it :smiley:

When the user is setting up 2FA and they submit the confirmation code from their authenticator, they will be taken to this new intermediate confirmation screen:

image

2FA is fully enabled at this point (as per the green success message).  The green button is simply a link to the main page showing 2FA status, which again confirms that 2FA is enabled:

image

(You don't actually have to click the big green button for 2FA to become enabled)

Hopefully this makes it more obvious that the user should save their backup code in a safe place.

colinodell:
I don't like how that giant "disable two-factor auth" button detracts from the green success message, so I've turned that into a link instead:

image

(That link still takes the user to the confirmation page which does still have a giant red button on it.)

Seldaek:
Thanks, looking all good! It's live.

Slamdunk:
Awesome :+1: 

Minor GUI flaw: in the login the checkbox for "Trust this computer" is unaligned and overlaps the input field

immagine

Agent: Firefox 70 on Xubuntu 19.10

css

PR #1033, opened 2019-10-09
Template tweaks

colinodell:
This PR makes two minor changes to the `user` templates:

 - Only set `isActualUser` in the block where it's needed - see https://github.com/composer/packagist/pull/1031#discussion_r332588385
 - Only render the <small> tag when we have content for it - see https://github.com/composer/packagist/pull/1031#discussion_r332588127


Seldaek:
Thanks

PR #1035, opened 2019-10-25
Spelling is hard

colinodell:


Seldaek:
Hehe it happens ;)

Seldaek:
If you can have a look at https://github.com/composer/packagist/pull/1031#issuecomment-546258205 tho that'd be nice, but no rush.

colinodell:
#1036 should fix that styling bug

PR #1036, opened 2019-10-25
Add missing form-group wrapper

colinodell:
Fixes https://github.com/composer/packagist/pull/1031#issuecomment-546258205

### DIFF
--- a/app/Resources/SchebTwoFactorBundle/views/Authentication/form.html.twig
+++ b/app/Resources/SchebTwoFactorBundle/views/Authentication/form.html.twig
@@ -22,7 +22,9 @@
                 </div>
 
                 {% if displayTrustedOption %}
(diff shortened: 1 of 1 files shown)


app/Resources/SchebTwoFactorBundle/views/Authentication/form.html.twig
```diff
@@ -22,7 +22,9 @@
                 </div>
 
                 {% if displayTrustedOption %}
-                    <div class="checkbox"><label for="_trusted"><input id="_trusted" type="checkbox" name="{{ trustedParameterName }}" /> Trust this computer for 30 days</label></div>
+                    <div class="input-group">
```
stof: isn't it a missing `form-group` instead ? `input-group` is meant to group the input and the addon.

colinodell: Yep, you're right, I copypasted the wrong class here.  This is what I get for rushing a fix before I've had my coffee... Fixed!

Seldaek:
Thanks

PR #1044, opened 2019-11-11
Added more mirrors

Webysther:
Based on recents finds: https://github.com/Webysther/packagist-mirror#-packagist-public-metadata-mirrors-around-the-world


Seldaek:
Thanks

PR #1046, opened 2019-11-23
Security Advisories

glaubinix:
This pull request integrates security advisories into packagist.org. It is built in a way to additional sources can be integrated.

**Users are able to:**
* see how many advisories are available for every package
* see which versions of a package are affected by one or more advisory
* see detailed information about advisories for a package
* query the data via an API

**Adjustments to the package view page:**
* number of security advisories
* red icon in the version list to see which versions are affected
Screenshot 2019-11-23 at 13 22 04

**New package security advisory page**
* list all advisories for a package
* via the query param `version` matching advisories will be highlighted
Screenshot 2019-11-23 at 13 24 56

### DIFF
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -258,7 +258,7 @@ class ApiController extends Controller
     public function securityAdvisoryAction(Request $request): JsonResponse
     {
         $packageNames = array_filter((array) $request->get('packages'));
-        if (!$packageNames) {
+        if ((!$request->query->has('updatedSince') && !$request->get('packages')) || (!$packageNames && $request->get('packages'))) {
             return new JsonResponse(['status' => 'error', 'message' => 'Missing array of package names as the "packages" parameter'], 400);
         }
 
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -267,9 +267,9 @@ class ApiController extends Controller
         /** @var array[] $advisories */
         $advisories = $this->getDoctrine()->getRepository(SecurityAdvisory::class)->searchSecurityAdvisories($packageNames, $updatedSince);
 
-        $response = [];
+        $response = ['advisories' => []];
         foreach ($advisories as $advisory) {
-            $response[$advisory['packageName']][] = $advisory;
+            $response['advisories'][$advisory['packageName']][] = $advisory;
         }
 
         return new JsonResponse($response, 200);
--- a/src/Packagist/WebBundle/Controller/ApiController.php
+++ b/src/Packagist/WebBundle/Controller/ApiController.php
@@ -257,9 +257,11 @@ class ApiController extends Controller
      */
     public function securityAdvisoryAction(Request $request): JsonResponse
     {
-        $packageNames = array_values(array_filter(array_map(function (string $packageName) {
-            return trim($packageName);
-        }, explode(',', $request->get('packages')))));
+        $packageNames = array_filter((array) $request->get('packages'));
+        if (!$packageNames) {
+            return new JsonResponse(['status' => 'error', 'message' => 'Missing array of package names as the "packages" parameter'], 400);
+        }
+
         $updatedSince = $request->query->getInt('updatedSince', 0);
 
         /** @var array[] $advisories */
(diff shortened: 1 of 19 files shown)


src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -1120,6 +1138,45 @@ public function versionStatsAction(Request $req, Package $package, $version)
...
+            $version = $this->getDoctrine()->getRepository(Version::class)->findOneBy([
+                'name' => $name,
+                'id' => $versionId,
+            ]);
+            if ($version) {
+                foreach ($securityAdvisories as $advisory) {
+                    $versionParser = new VersionParser();
+                    $affectedVersionConstraint = $versionParser->parseConstraints($advisory['affectedVersions']);
+                    if (!isset($data['hasVersionSecurityAdvisories'][$version->getId()]) && $affectedVersionConstraint->matches(new Constraint('=', $version->getNormalizedVersion()))) {
+                        $data['matchingAdvisories'] = $advisory['id'];
```
Seldaek: This should be an array of ids matching I believe?

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -1120,6 +1138,45 @@ public function versionStatsAction(Request $req, Package $package, $version)
...
+
+        $data['matchingAdvisories'] = [];
+        if ($versionId = $request->query->getInt('version')) {
+            $version = $this->getDoctrine()->getRepository(Version::class)->findOneBy([
+                'name' => $name,
+                'id' => $versionId,
+            ]);
+            if ($version) {
+                foreach ($securityAdvisories as $advisory) {
+                    $versionParser = new VersionParser();
```
Seldaek: No need to create one per iteration?

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -1120,6 +1138,45 @@ public function versionStatsAction(Request $req, Package $package, $version)
...
+                    $affectedVersionConstraint = $versionParser->parseConstraints($advisory['affectedVersions']);
+                    if (!isset($data['hasVersionSecurityAdvisories'][$version->getId()]) && $affectedVersionConstraint->matches(new Constraint('=', $version->getNormalizedVersion()))) {
```
Seldaek: The isset here seems off. Seems it was copied over from viewPackageAction

src/Packagist/WebBundle/Controller/PackageController.php
```diff
@@ -1120,6 +1138,45 @@ public function versionStatsAction(Request $req, Package $package, $version)
...
+                        $data['matchingAdvisories'] = $advisory['id'];
+                    }
+                }
+            }
+        }
```
Seldaek: Generally speaking once this block executes, the paginator will be in a bad state I assume, as this filters the list of securityAdvisories which are displayed? Might have missed something, but it sounds to me like the filtering should happen before paginator and count etc are assigned to $data?

src/Packagist/WebBundle/Controller/ApiController.php
```diff
@@ -246,6 +247,32 @@ public function trackDownloadsAction(Request $request)
...
+            return trim($packageName);
+        }, explode(',', $request->get('packages')))));
+        $updatedSince = $request->query->getInt('updatedSince', 0);
+
+        /** @var array[] $advisories */
+        $advisories = $this->getDoctrine()->getRepository(SecurityAdvisory::class)->searchSecurityAdvisories($packageNames, $updatedSince);
+
+        $response = [];
+        foreach ($advisories as $advisory) {
+            $response[$advisory['packageName']][] = $advisory;
```
Seldaek: Should be returned nested as `{"advisories": {...}}`

src/Packagist/WebBundle/Controller/PackageController.php
glaubinix: The pagination is currently unused. Removed it completely.

Seldaek:
Thanks!

scheb:
Really like this feature! Did you consider making use of this and let Composer warn about security advisories when you're using an affected version?

Recently, I had to deal with a security issue and I found it hard to let people know, because I don't know all of the users and there is no direct communication channel to all of them.

I know there are solutions out there to check your project for security issues, but these require addtional software to be added to your project/build - not everyone is doing this. Composer is the one piece of software that everyone uses to install their dependencies and every once in a while you do a composer install/update. I think it would be really nice if it would warn you when there is a security risk in your current dependencies. Would give these issues more visibility and hopefully encourage people to upgrade.

Issue #985, opened 2019-01-11
Allow anyone to trigger a package update if updatedAt > 24h


Seldaek:
Might help to allow users to help themselves for packages that aren't synced automatically or that failed to sync for some reason.

PR #1047, opened 2019-11-24
Package: allow updates to be triggered by every logged in user if last updatedAt was > 24 hours ago

glaubinix:
Implements: https://github.com/composer/packagist/issues/985


Seldaek:
Thanks

PR #1049, opened 2019-11-25
Add api_url to popular endpoint in json response payload

kwidua:
```
{
    "packages": [
        {
            "api_url": "http://127.0.0.1:8000/app_dev.php/packages/composer/composer.json",
            "description": "",
            "downloads": 0,
            "favers": 0,
            "name": "composer/composer",
            "url": "http://127.0.0.1:8000/app_dev.php/packages/composer/composer"
        }
    ],
    "total": 1
}

```

This helps clients to avoid generating the api url on their side.


beberlei:
We just realized from @naderman input that setting the _format is also possible via Accept header, so this change might not be necessary:

```
curl -H "Accept: application/json" https://packagist.org/explore/popular.json|python -mjson.tool
```

naderman:
Indeed, thank you anyway for trying to improve the API! Definitely appreciated. But I think in this case it'd be simpler to stick to a single URL in the document which works for either use case, so closing this.

PR #1052, opened 2019-11-28
Display funding information for packages and expose funding info to Composer

naderman:
Requires funding support (https://github.com/composer/composer/pull/8453) to be merged into Composer and the composer/composer dep to be updated before this can be merged.

Displaying of funding info in the UI is pretty basic for now. It simply lists all links in the details section. Maybe someone can come up with a nicer way to display this? Maybe a button in that section that is a bit more obvious and then a modal with all the options?

Screenshot_20191129_001636

*Requires the addition of a funding text column to the version table prior to deployment.*

### DIFF
--- a/src/Packagist/WebBundle/Entity/Version.php
+++ b/src/Packagist/WebBundle/Entity/Version.php
@@ -180,7 +180,7 @@ class Version
     private $support;
 
     /**
-     * @ORM\Column(type="text", nullable=true)
+     * @ORM\Column(type="json", nullable=true)
      */
     private $funding;
 
@@ -270,7 +270,7 @@ class Version
         if ($serializeForApi && $this->getSupport()) {
             $data['support'] = $this->getSupport();
         }
-        if ($serializeForApi && $this->getFunding()) {
+        if ($this->getFunding()) {
             $data['funding'] = $this->getFunding();
         }
         if ($this->getReleasedAt()) {
@@ -329,7 +329,6 @@ class Version
             $array['support'] = $this->getSupport();
             ksort($array['support']);
         }
-        // TODO do we need to do something for funding here? sort by type/url?
 
         return $array;
     }
@@ -632,7 +631,17 @@ class Version
      */
     public function setFunding($funding)
     {
-        $this->funding = $funding ? json_encode($funding) : null;
+        // sort records when storing so to help the V2 metadata compression algo
+        if ($funding) {
+            usort($funding, function ($a, $b) {
+                $keyA = ($a['type'] ?? '') . ($a['url'] ?? '');
+                $keyB = ($b['type'] ?? '') . ($b['url'] ?? '');
+                
+                return $keyA <=> $keyB;
+            });
+        }
+
+        $this->funding = $funding;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


naderman:
@Seldaek there is one open question in the code as a TODO, in how far the funding array needs to be sorted/modified for the v2 format. It also wasn't quite clear to me what the up/downside of handling the data like the `$support` (text column) or the `$authors` (json column) property is. Maybe it should use the authors mechanism instead?

naderman:
Tests fail because it requires the Composer PR linked in the description.

Seldaek:
@naderman should definitely be a json column, the text one is historical because lack of json support in mysql back then

src/Packagist/WebBundle/Entity/Version.php
```diff
@@ -265,6 +270,9 @@ public function toArray(array $versionData, bool $serializeForApi = false)
         if ($serializeForApi && $this->getSupport()) {
             $data['support'] = $this->getSupport();
         }
+        if ($serializeForApi && $this->getFunding()) {
```
stof: `$serializeForApi` should be removed here. We want to export the funding metadata in the composer repository too. Otherwise, `composer fund` would not have the funding data for packages installed from packagist (we already have this issue for `support`, but missing funding info will be a bigger issue as it makes the whole funding feature broken)

src/Packagist/WebBundle/Resources/views/package/view_package.html.twig
```diff
@@ -174,6 +174,16 @@
                             {% endif %}
                         </div>
 
+                        <div class="facts col-xs-12 col-sm-6 col-md-12">
+                            {% if version and version.funding is defined and version.funding|length > 0 %}
```
stof: How can `version.funding` be undefined ? Will `version` be something different than the `Version` entity ?

Note that `is defined` is **not** a null check.

src/Packagist/WebBundle/Resources/views/package/view_package.html.twig
```diff
@@ -174,6 +174,16 @@
...
+                                <p>
+                                    Fund package maintenance!
+                                    {% for fundingOption in version.funding %}
+                                        <br>&raquo; <a rel="nofollow noopener external noindex" href="{{ fundingOption.url }}">{{ fundingOption.type == 'custom' ? fundingOption.url : fundingOption.type }}</a>
```
stof: what if we have multiple funding of the same type (for instance multiple github funding configs, for different people) ?

src/Packagist/WebBundle/Entity/Version.php
naderman: Thank you, missed that!

src/Packagist/WebBundle/Resources/views/package/view_package.html.twig
naderman: @stof that should indeed be displayed differently for the links then. As per the PR description I was hoping maybe someone else would help make better UI for this or I'll have to give it a shot myself. Definitely not leaving it like this.

src/Packagist/WebBundle/Entity/Version.php
```diff
@@ -179,6 +179,11 @@ class Version
      */
     private $support;
 
+    /**
+     * @ORM\Column(type="text", nullable=true)
```
Seldaek: As per discussion this should be type json, then the set/getter can be simplified

Seldaek:
Merging as is, will look at UI later

PR #1058, opened 2019-12-16
Security Advisories: store reported date

glaubinix:
This allows security advisories to be displayed sorted by reported date to make it easier to see the newest advisories for a package first.

Note about the schema changes:
* I changed `affectedVersions` to `text` from `string` because some symfony/symfony advisories go past the 255 char limit and are being truncated
* Easiest way to apply the `reportedAt` column is to truncate the table and trigger a new update job for the advisory repo. As an alternative apply with DEFAULT NULL. Run a job and drop the default.

### DIFF
--- a/src/Packagist/WebBundle/Entity/SecurityAdvisory.php
+++ b/src/Packagist/WebBundle/Entity/SecurityAdvisory.php
@@ -51,7 +51,7 @@ class SecurityAdvisory
     private $cve;
 
     /**
-     * @ORM\Column(type="string")
+     * @ORM\Column(type="text")
      */
     private $affectedVersions;
 
@@ -59,6 +59,11 @@ class SecurityAdvisory
      * @ORM\Column(type="string")
      */
     private $source;
+
+    /**
+     * @ORM\Column(type="datetime")
+     */
+    private $reportedAt;
 
     /**
      * @ORM\Column(type="datetime")
@@ -79,9 +84,11 @@ class SecurityAdvisory
             $this->title !== $advisory->getTitle() ||
             $this->link !== $advisory->getLink() ||
             $this->cve !== $advisory->getCve() ||
-            $this->affectedVersions !== $advisory->getAffectedVersions()
+            $this->affectedVersions !== $advisory->getAffectedVersions() ||
+            $this->reportedAt !== $advisory->getDate()
         ) {
             $this->updatedAt = new \DateTime();
+            $this->reportedAt = $advisory->getDate();
         }
 
         $this->remoteId = $advisory->getId();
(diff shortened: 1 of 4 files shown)


src/Packagist/WebBundle/Entity/SecurityAdvisory.php
```diff
@@ -79,9 +84,11 @@ public function updateAdvisory(RemoteSecurityAdvisory $advisory): void
             $this->title !== $advisory->getTitle() ||
             $this->link !== $advisory->getLink() ||
             $this->cve !== $advisory->getCve() ||
-            $this->affectedVersions !== $advisory->getAffectedVersions()
+            $this->affectedVersions !== $advisory->getAffectedVersions() ||
+            $this->reportedAt !== $advisory->getDate()
```
stof: using strict comparison on a DateTime instead is wrong. It will always report them being different as they won't be the same object. You should use `!=` to compare them by value.

src/Packagist/WebBundle/Entity/SecurityAdvisoryRepository.php
```diff
@@ -18,15 +18,15 @@ public function getPackageSecurityAdvisories(string $name): array
         $sql = 'SELECT s.*
             FROM security_advisory s
             WHERE s.packageName = :name
-            ORDER BY s.id DESC';
+            ORDER BY s.reportedAt DESC';
```
stof: I would still add the sorting by id to ensure that we have a stable sorting.

stof:
Btw, reviewing this PR, I identified another issue related to advisories: the FriendsOfPhp repository allows to submit advisories for packages which are not distributed through packagist.org but through a custom repository.
Advisories which have `composer-repository: false` should be excluded (see the magento advisories)
And advisories which have it set to a string other than `https://packagist.org` (which is the implicit default value, so it does not actually appear in YAML files) should probably also be excluded (even though there is none currently, as the drupal-contrib advisories have been reverted due to their ecosystem reusing the same name+version in both the Drupal 7 and Drupal 8 repositories while being different versions and so different vulnerabilities)

glaubinix:
I updated the pull request to also store the composer repository value.

Seldaek:
Thanks

stof:
@glaubinix the repository is now stored, but not taken into account when matching package names to security advisories, so it is actually still broken.

glaubinix:
@stof I take it you are referring to to the advisories listed on for instance the page here https://packagist.org/packages/symfony/symfony/advisories and the security count on the package page? I'll have a look in the next days to adjust it.

PR #1061, opened 2020-01-13
Adding new mirror from India.

vrkansagara:
Adding new mirror from India.


Seldaek:
Already did this change here as I saw your tweet, but thanks

Issue #1006, opened 2019-04-24
Installation stats per major version


stof:
Current, the installation stats are providing a total number, and a number per release (with evolution charts for both of them).

It would be great to have a similar stats per major version (allowing to answer the question "is Twig 1.x used more or less often than Twig 2.x ?" for instance, which is exactly my current use case). Comparing only the latest patch release of each branch is not enough (most used versions of Twig are not actually the latest patch releases right now, due to being recent)

stof:
even better would be to allow drawing the different major versions on the same chart to compare them...

nicekiwi:
Are stats collected in the DB for this?

stof:
Well, I'm quite sure that the current code does not collect such stats in Redis. But we could do it, as the request reports the version (which is used today for exact version graphs), and so we could create an aggregate key per major versions too.

Seldaek:
Stats are in the DB yes, should be easy enough to do a grouped query merging all versions by their major version I think.

acornforth:
I just arrived to this page to request/suggest a very similar thing.
For any given composer package, I would like to see, in the version history for the package (bottom right of page on desktop site, where the warning triangle appears for versions with known security issues) the number of downloads for that version. (another nice to have would be the ability to toggle the view between Total downloads, (since forever) and downloads in the last month/6months etc.

Seldaek:
https://packagist.org/packages/twig/twig/stats 🎉

stof:
@Seldaek @glaubinix thanks a lot for that. :tada:

PR #1065, opened 2020-02-03
Installs per package major version

glaubinix:
This implements package installs per major version as mentioned in https://github.com/composer/packagist/issues/1006

There is a toggle at the top of the version list which lets you switch between versions and major versions. Toggling will automatically select the fist version/major version in the list.

Screenshot 2020-02-03 at 21 29 12

### DIFF
--- a/src/Packagist/WebBundle/Controller/PackageController.php
+++ b/src/Packagist/WebBundle/Controller/PackageController.php
@@ -982,6 +982,7 @@ class PackageController extends Controller
      */
     public function statsAction(Request $req, Package $package)
     {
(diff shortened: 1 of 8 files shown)


stof:
@Seldaek should we do the same for minor versions or no ?

glaubinix:
@stof this was added after the PR got merged. See for instance: https://packagist.org/packages/twig/twig/stats#major/2

stof:
indeed, I missed it because I only toggle between `all` and `3`, and Twig 3 has only 1 minor version for now.

PR #1076, opened 2020-04-08
Fix funding links overflowing the sidebar

colinodell:
Currently, long funding links can overflow the sidebar:

image

Applying the same styling as the `.details` section fixes this:

image


Seldaek:
Thanks

PR #1078, opened 2020-04-09
Retain original funding sort order on website (fixes #1077)

### LINKED ISSUE #1077: Funding links out-of-order
Commit 54fcb5016afa8b2035d8cd1537cc0fea2cab11ba changes the order of funding links to something other than the author's preferred sort order (as provided in either their `composer.json` or `FUNDING.yml` file).

As a result, my funding sources display in this order:

image

Whereas my FUNDING.yml file uses this order:

```yml
github: colinodell
patreon: colinodell
tidelift: packagist/league%2fcommonmark
custom: ["https://www.colinodell.com/sponsor", "https://www.paypal.me/colinpodell/10.00", "https://enjoy.gitstore.app/repositories/thephpleague/commonmark"]
```

Because GitHub is offering to match sponsorships, I would prefer to have that link show first.  Having Packagist respect my preferred order would be the easiest way to accomplish this.

GitHub (of course) puts their Sponsor program first, but at least it respects my `custom` options:

image

## Potential Solutions

Would any of these potential solutions be feasible?

1. Revert 54fcb5016afa8b2035d8cd1537cc0fea2cab11ba if the impact isn't too significant
2. Only perform the sorting when generating metadata (instead of pre-sorting it and losing that sort order)
3. Store the original sort order in the database and use that on the Packagist.org frontend (perhaps store a third `sort_order` key in the JSON, strip it out when generating metadata, or re-sort by this when using on the frontend, or just store the sorted data in an additional field)

colinodell:
Instead of sorting the funding sources when saving, we store them with the original sort order and only perform sorting when generating metadata. This allows the V2 algo to retain the same performance boost while ensuring the Packagist website presents the sources in the author's preferred order.

Fixes #1077

### DIFF
--- a/src/Packagist/WebBundle/Entity/Version.php
+++ b/src/Packagist/WebBundle/Entity/Version.php
@@ -271,7 +271,7 @@ class Version
             $data['support'] = $this->getSupport();
         }
         if ($this->getFunding()) {
-            $data['funding'] = $this->getFunding();
+            $data['funding'] = $this->getFundingSorted();
         }
         if ($this->getReleasedAt()) {
             $data['time'] = $this->getReleasedAt()->format('Y-m-d\TH:i:sP');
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, deployed now but note you'll need a new commit on your package's default branch for the funding info to update (and un-sort itself).

colinodell:
Makes sense.  Thank you so much!

PR #1087, opened 2020-05-21
Guard: fix 2FA implementation for username+password

glaubinix:
Regression which was introduced by https://github.com/composer/packagist/pull/1082


glaubinix:
Fixed with https://github.com/composer/packagist/commit/e96f425804048b51e9fb230b4570c5650e68712e

PR #1104, opened 2020-07-21
Fix: Use https instead of http

localheinz:
This PR

* [x] uses `https` instead of `http` in URLs pointing to author homepages


localheinz:
Thank you, @Seldaek!

PR #1127, opened 2020-10-07
Add little hand to summary element

jawira:
This adds a little hand to the `<summary>` element.

Before:

ksnip_20201007-160423

After:

ksnip_20201007-160232

This improves UX and also is the default behavior in GitHub:

<details>
<summary>Summary has a little hand</summary>
Here goes the details.
</details>

### DIFF
--- a/src/Packagist/WebBundle/Resources/public/css/main.css
+++ b/src/Packagist/WebBundle/Resources/public/css/main.css
@@ -1734,7 +1734,3 @@ svg.chart {
 .icon-lock:before { content: '🔒'; } /* '\1f512' */
 .icon-lock-open:before { content: '🔓'; } /* '\1f513' */
 .icon-rocket:before { content: '🚀'; } /* '\1f680' */
-
-.markdown-body details summary {
-    cursor: pointer;
-}
--- a/src/Packagist/WebBundle/Resources/public/css/main.css
+++ b/src/Packagist/WebBundle/Resources/public/css/main.css
@@ -1734,3 +1734,7 @@ svg.chart {
 .icon-lock:before { content: '🔒'; } /* '\1f512' */
 .icon-lock-open:before { content: '🔓'; } /* '\1f513' */
 .icon-rocket:before { content: '🚀'; } /* '\1f680' */
+
+.markdown-body details summary {
+    cursor: pointer;
+}
--- a/web/css/github/markdown.css
+++ b/web/css/github/markdown.css
@@ -676,7 +676,3 @@
 .markdown-body hr {
   border-bottom-color: #eee;
 }
-
-.markdown-body details summary {
-  cursor: pointer;
-}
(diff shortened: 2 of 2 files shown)


web/css/github/markdown.css
```diff
@@ -676,3 +676,7 @@
 .markdown-body hr {
   border-bottom-color: #eee;
 }
+
+.markdown-body details summary {
```
stof, line 680: This file is a vendor one (see the comment on top). Modifying it would make maintenance harder (if the file gets updated to a newer version, manual changes will be lost).
Project styles should be added in src/Packagist/WebBundle/Resources/public/css/main.css instead

jawira: Thank you for answering so quickly. I moved the style to `main.css`.

src/Packagist/WebBundle/Resources/public/css/main.css
```diff
@@ -1734,3 +1734,7 @@ svg.chart {
 .icon-lock:before { content: '🔒'; } /* '\1f512' */
 .icon-lock-open:before { content: '🔓'; } /* '\1f513' */
 .icon-rocket:before { content: '🚀'; } /* '\1f680' */
+
+.markdown-body details summary {
```
stof, line 1738: I even suggest to use only `details summary` (or maybe even `summary` if that's enough specificity). I don't think the extra specificity to restrict it to `.markdown-body` is necessary.

jawira: I think it's a good idea to leave `.markdown-body`, this can give the context about why the rule is there. Do you agree?

Also, I have reviewed `web/css/github/markdown.css` and this have have not been updated since 2016, although the latest version from this css file is from February 2020. Don't you think the best thing to do is to update this file? I can do in this PR or in a new one.

stof: updating it would indeed be a good idea (and would solve this issue as the new version include this rule)

web/css/github/markdown.css
```diff
@@ -1,124 +1,83 @@
 /* from https://github.com/sindresorhus/github-markdown-css/blob/gh-pages/github-markdown.css */
 
-.markdown-body {
-/*REMOVED  -ms-text-size-adjust: 100%;
```
jawira, line 4: I updated `markdown.css`, but I noticed that previous code was sometimes commented with `/*REMOVED`.
Should I also comment the code (when possible)?

stof: I would say yes, to preserve the customizations done by @Seldaek when starting to use that stylesheet

jawira: I commented the code when required.

jawira: I will resolve this conversation, please let me know if you need any other change.

stof (approved):

Seldaek:
Thanks

PR #1133, opened 2020-10-29
Add French mirror

baptistepillot:


Seldaek:
Why though? What's the point? We already have tons of bandwidth very close in europe.. And this mirror script is not good and does not support composer 2 so it actually slows things down when using composer 2.

baptistepillot:
> Why though? What's the point? We already have tons of bandwidth very close in europe.. And this mirror script is not good and does not support composer 2 so it actually slows things down when using composer 2.

@Seldaek I had problems on `composer update` commands since 10/20/2020 : it became very slow ! From my office, from a production server (I did not tried all of my production servers, perhaps I should !)... when it does not work from several places, for me it does not work.
Re-tried almost every day : it takes me systematically 5-7 minutes to complete each composer command from packagist.org (strangely had the same slowdown with some others mirrors in Germany and Finland).
With this new packagist.fr repo I mirrored a few days agro : updates are done in 15 seconds. Reasonable for my project.

Things are perhaps better today : I just retried it and packagist.org seems fast again from my office (packagist.org : 1'06 ; packagist.fr : 1'07 - yep, slow VPN here)... but not from a production server (packagist.org : 7'32, with an error ! ; packagist.fr : 0'16).

I don't know why it became so slow for me. Another strange thing I did not think to check is what IP packagist.org resolves from different systems... Until now my accesses where slow from everywhere, but know results seem different. DNS resolution :

- packagist.org => 54.38.136.239 (eu) from my office (Bouygues Telecom Business) : works well
- packagist.org => 2001:41d0:801:1000::27d (and others IPv6 stuff like that, changing each time I ping) from my production server (OVH Gravelines, FR). Strange : it does not respond to pings, and updates are very slow here today.
- packagist.org => 54.37.131.18 (eu) from another production server (Online Paris, FR) : works well

Well, I don't understand why... My nameserver is 1.1.1.1/1.0.0.1 on all these sites. And last days it was very slow on my office too. I did not check the DNS until now, this is probably the explanation why I had big slowdowns.

Today it looks like a local DNS problem.
I am not a specialist in networks, today it gives me headache ! I don't undestand what is happening when from my server I've got :

```bash
# ping packagist.org
PING packagist.org(2001:41d0:801:1000::895 (2001:41d0:801:1000::895)) 56 data bytes
# traceroute packagist.org
traceroute to packagist.org (51.38.227.34), 30 hops max, 60 byte packets
```

Why both commands do not resolve the same IP ? This is getting me crazy.

So. There is perhaps no point with packagist.org. My offer for a french mirror of packagist is still on, I will maintain it, even if it perhaps solves a local problem, I will still maintain packagist.fr for my internal use, whenever it is referenced as an official mirror or not. I just wanted to share it if it may be of any need.

Now about composer 2 : do you have a better script to create mirrors, that will work well with the new version ? When will it be released ?

Seldaek:
So it looks like your server has IPv6 enabled.. why that would end up being slow though I can not imagine, as our own servers are at OVH so the network between should be pretty good.

Could you try Composer 2 updates from your slow server? I wonder if that works any faster.

stof:
> Why both commands do not resolve the same IP ? This is getting me crazy.

That's because `traceroute` forces using IPv4. Use `traceroute6` for the equivalent on IPv6.

stof:
> Now about composer 2 : do you have a better script to create mirrors, that will work well with the new version ? When will it be released ?

See https://github.com/composer/mirror for the official mirroring script.

baptistepillot:
Ok : you win. With composer 2, it works very-very well from my server which pings an IPv6 packagist.org :
- packagist.org : 2 seconds,
- packagist.fr : 5 seconds (with your composer 2 warning).

Well, it seems that you don't need a brand new composer 1 french mirror xD.
And it seems that I need to upgrade to composer 2 everywhere, to have this faster. You changed your network infrastructure for composer 2, right ?

So, you will disable third-party mirrors ? There will not be any third-party mirrors with composer 2 ?

Seldaek:
We can't disable them but my current recommendation would be to avoid them for v2 until it's proven that they bring anything. Probably in China due to their firewall it'll still be needed, but there they have a correct mirror impl running and not the websyther one. For other places due to the parallelization of requests in Composer v2, the latency is much less of an issue, and bandwidth usage is reduced too, so I think it should be quite good for users in most locations.

stof:
> And it seems that I need to upgrade to composer 2 everywhere, to have this faster and more reliable. You changed your network infrastructure for composer 2, right ?

No. What changed is the format of metadata in v2-enabled repositories, to allow a much more efficient transfer of data (transferring less bytes, which then go faster)
> So, you will disable third-party mirrors ? There will not be any third-party mirrors with composer 2 ?

Third-party mirrors are not forbidden AFAIK. But the main issue is that most third-party mirrors available today are based on the same mirroring tool which does not support mirroring the v2 metadata.
And thanks to the huge improvement done in the v2 format, using a close mirror with v1 metadata will still be slower than using the official servers with v2 metadata.
So third-party mirrors won't be relevant anymore unless they also mirror v2 metadata (and even then, the impact of a closer mirror should be measured, as it might not be worth the investment)

Seldaek:
@stof composer 2 uses curl + http/2 so it's more than just the repo format which changed. That also means we use https for every request, which can change routing in some cases.

baptistepillot:
So, all good news. Thanks a lot for all these explanations.

Great job for those speed improvements in composer 2. The only reproach I had against composer was its slowness... Now no problem anymore, it's going to be ancient history. Let's work at the speed of light !

Thanks again.

webysther:
Hi @stof and @Seldaek I will update soon the script to support the new script and keep the same interface to the current mirrors.

Seldaek:
@webysther have you tried composer 2 though? I've heard from people in south africa that the mirror wasn't needed anymore, curious to hear from south america :)

webysther:
I will attend to PHP Conference Brazil this year to know better how broad is the current usage, for my personal job the new composer fixed the problem with the load times, but I keep using the mirror for better performance and reliability. Some private companies will keep the mirror internally.

I send the statistics to you, is there no way to publish how much % of users are using the mirror in the world today?

webysther:
The usage is getting high and high every month, with lots os usage on bandwith in Brazil, but this probally change in the next year, I will create some benchmark to measure the advantages (or not) to using the mirror.

As **APT** or **NPM**, the mirror is important for HA around the world and help to keep the community working in situations with anomalies, wich sometimes occur to the servers of packagist.

Seldaek:
@webysther ok thanks, I can see your point about reliability when talking about global internet outages, long distances don't help there.

It's hard to know what % of users are using the mirror, because I don't see how much traffic you get vs how much we get..

You could tell me how many requests for /packages.json you get in one day for example, that would give us an idea :)

PR #1136, opened 2020-11-14
Replace use of the deprecated LockHandler class

mbabker:
`Symfony\Component\Filesystem\LockHandler` is deprecated and removed in Symfony 4, this replaces the use of the deprecated class with the `LockableTrait` that was added to the Console component for this exact use case.


Seldaek:
Thanks

PR #1137, opened 2020-11-14
Replace deprecated Twig uses

mbabker:
Some of the Twig integration was still using older class names, and the templates were using the deprecated filter tag.  These have been updated.

### DIFF
--- a/src/Packagist/WebBundle/Resources/views/forms.html.twig
+++ b/src/Packagist/WebBundle/Resources/views/forms.html.twig
@@ -1,45 +1,38 @@
 {% extends 'form_div_layout.html.twig' %}
 
-{% block textarea_widget %}
-{% set attr = attr|merge({'class': attr.class|default('') ~ ' form-control'}) %}
-{{ parent() }}
-{% endblock textarea_widget %}
+{% block textarea_widget -%}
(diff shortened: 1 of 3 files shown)


src/Packagist/WebBundle/Resources/views/forms.html.twig
```diff
@@ -6,16 +6,16 @@
 {% endblock textarea_widget %}
 
 {% block form_widget_simple %}
-{% filter spaceless %}
+{% apply spaceless %}
```
stof, line 9: To match what is done in Symfony core since years (for better performance and better result regarding the intended output), the form theme should use whitespace control rather than the `spaceless` filter (better performance is because whitespace controll is a compile-time feature while spaceless is a runtime one, and rendering a form renders lots of these blocks)

mbabker: Done.

mbabker:
On a related note, it really seems like the form theme could just extend `bootstrap_3_layout.html.twig` at this point, but that also might be a bigger undertaking since looking at the FOSUserBundle template overrides there are a lot of overridden form bits.

Seldaek:
Thanks - I am not so concerned about the form theme tbh, we have bigger legacy issues to fix here like ripping out the entire FOSUserBundle and migrating to a newer Symfony, etc.

PR #1138, opened 2020-11-14
Upgrade Pagerfanta

mbabker:
`WhiteOctoberPagerfantaBundle` is now abandoned and my `BabDevPagerfantaBundle` is the replacement for it.  The bundle is replaced with this pull request, and deprecated uses of the Pagerfanta API replaced as well.


Seldaek:
Thanks

PR #1139, opened 2020-11-23
Do not use outdated PHP version

OskarStark:
follows #976


Seldaek:
Thanks, but we now recommend the `>=7.4` format on composer docs, so I think we can use that here too.

PR #1143, opened 2020-11-30
Symfony 4.4

mbabker:
This is a minimum effort PR to get Packagist running on Symfony 4.4 (so nothing like Flex was added or filesystem structure changed).  As best as I could test this on a local clone, it seems like it's working fine.


app/AppKernel.php
```diff
@@ -41,4 +41,24 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @note Overridden path is to retain the pre-existing path structure and can be removed if moving to Symfony defaults
+     */
+    public function getCacheDir()
+    {
+        return $this->getProjectDir().'/app/cache/'.$this->environment;
```
mbabker, line 52: Just as the docblock notes, I added overrides for `getCacheDir()` and `getLogDir()` to keep those inside the `<project>/app` folder.  This can be dropped if you're good with moving these to the default `<project>/var` folder.

composer.json
```diff
@@ -67,17 +66,20 @@
         "phpunit/phpunit": "^8.0"
     },
     "scripts": {
+        "auto-scripts": [
+            "@php app/console cache:clear",
+            "@php app/console assets:install --symlink --relative web"
```
mbabker, line 71: Since the distribution bundle isn't compatible with Symfony 4, I moved the commands it was executing out to be triggered directly.  `app/bootstrap.cache.php` isn't generated anymore either.

src/Packagist/WebBundle/Controller/Controller.php
```diff
@@ -13,7 +13,7 @@
 namespace Packagist\WebBundle\Controller;
 
 use Doctrine\Persistence\ManagerRegistry;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller as BaseController;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController as BaseController;
```
mbabker, line 16: `Controller` was deprecated, so made this swap.

src/Packagist/WebBundle/Entity/AuthorRepository.php
```diff
@@ -13,14 +13,14 @@
 namespace Packagist\WebBundle\Entity;
 
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
-use Symfony\Bridge\Doctrine\RegistryInterface;
+use Doctrine\Persistence\ManagerRegistry;
```
mbabker, line 16: `Symfony\Bridge\Doctrine\RegistryInterface` was deprecated, so changed to the base registry.

composer.json
stof: why not keeping these in `symfony-scripts` though ?

Using the `auto-scripts` name to match Flex looks like a bad idea to me, because Flex also manages a custom structure for the `auto-scripts` key, while you use the standard Composer feature. So this might cause incompatibility with a Flex plugin.

web/app.php
```diff
@@ -3,7 +3,7 @@
 use Symfony\Component\HttpFoundation\Request;
 
 /**
- * @var \Symfony\Component\ClassLoader\ClassLoader
+ * @var \Composer\Autoload\ClassLoader
  */
 $loader = require __DIR__.'/../app/autoload.php';
```
stof, line 8: `$loader` even looks unused, so the variable could be skipped entirely.

composer.json
mbabker: Originally I tried flat out copying an `auto-scripts` from another project not realizing the syntax for that comes from Flex, so when changed it to straight up `@php` calls I didn't think twice about moving it back.

web/app.php
mbabker: `$loader` assignments removed

composer.json
mbabker: The scripts are moved back under `symfony-scripts`

mbabker:
Also, CI is going to fail here because of the way the GitHub API keys are getting pushed into the `parameters.yml` file.

`symfony/yaml` as of 4.0 throws an exception on duplicate keys, and right now the Travis script is just appending new rows to the `parameters.yml` file which is causing duplicate keys.  My sed skills are a little rough, but in theory, this diff should fix it (hard to test without spamming this repo since the environment variables aren't going to be present in my fork):

```diff
diff --git a/.travis.yml b/.travis.yml
index a7d2dd95..fc5c7c54 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,8 @@ matrix:
 
 before_install:
     - cp app/config/parameters.yml.dist app/config/parameters.yml
-    - 'echo "    github.client_id: $GH_CLIENT_ID" >> app/config/parameters.yml'
-    - 'echo "    github.client_secret: $GH_CLIENT_SECRET" >> app/config/parameters.yml'
+    - "sed -i s/github.client_id: CHANGE_ME_IN_PROD/github.client_id: $GH_CLIENT_ID/g app/config/parameters.yml"
+    - "sed -i s/github.client_secret: CHANGE_ME_IN_PROD/github.client_secret: $GH_CLIENT_SECRET/g app/config/parameters.yml"
     - composer self-update --2
 
 install:
```

Seldaek:
Thanks I guess but please ask before doing huge amounts of work :/ I started this already last Friday doing a more comprehensive migration so I didn't get it completed yet, see https://github.com/composer/packagist/pull/1144 - I guess I'll take a look here though it might contain a few more things I forgot to do, probably easier that way than you looking at the huge diff in my PR.

mbabker:
It’s all good, I purposefully went low effort because at worst I was really only out a couple hours over a holiday weekend, and I ended up fixing a busted Redis config on my system while working on this so there’s still some positives from it all.

PR #1144, opened 2020-12-01
Upgrade to Symfony 4.4

Seldaek:
Not quite finished yet.


composer.json
```diff
@@ -19,78 +19,132 @@
...
+        "paragonie/random_compat": "2.*",
+        "symfony/polyfill-ctype": "*",
+        "symfony/polyfill-iconv": "*",
+        "symfony/polyfill-php71": "*",
+        "symfony/polyfill-php70": "*",
+        "symfony/polyfill-php56": "*"
     },
     "scripts": {
-        "symfony-scripts": [
-            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
```
stof, line 71: is it intended to remove that script while you kept the `incenteev/composer-parameter-handler` dependency ?

config/packages/framework.yaml
```diff
@@ -0,0 +1,25 @@
...
+        handler_id: snc_redis.session.handler
+
+    #esi: true
+    fragments: true
+    php_errors:
+        log: true
+    form:            true
+    assets:
+        version: 'v=%env(default::ASSETS_VERSION)%'
+    templating:
```
stof, line 24: should be removed, as you don't depend on `symfony/templating`

config/services.yaml
```diff
@@ -0,0 +1,201 @@
...
+
+    # makes classes in src/ available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    App\:
+        resource: '../src/'
+        exclude:
+            - '../src/DependencyInjection/'
+            - '../src/Entity/'
+            - '../src/Kernel.php'
+            - '../src/Tests/'
```
stof, line 48: I suggest moving tests from `src/Tests` to `tests` instead, to match the recommended structure.

config/services.yaml
```diff
@@ -0,0 +1,201 @@
...
+#        resource: '../../Command/*'
+#        tags: ['console.command']
+
+    App\Entity\:
+        resource: '../src/Entity/*Repository.php'
+
+ #   App\Service\:
+ #       resource: '../../Service/*'
+
+#    App\SecurityAdvisory\:
```
stof, line 70: why adding these as commented ?

config/services.yaml
```diff
@@ -0,0 +1,201 @@
...
+#        resource: '../../SecurityAdvisory/*.php'
+
+    App\Twig\PackagistExtension:
+        tags:
+            - { name: twig.extension }
```
stof, line 75: not necessary. The directory loading line 42 already defines such service, and the autoconfiguration is already adding the tag (same for several other services defined below btw)

config/services.yaml
```diff
@@ -0,0 +1,201 @@
...
+        tags:
+            - { name: kernel.event_subscriber }
+
+    App\Security\BruteForceLoginFormAuthenticator:
+        bind:
+            $recaptchaVerifier: "@beelab_recaptcha2.verifier"
+
+    App\Security\RecaptchaHelper:
+        bind:
+            $recaptchaEnabled: "%env(bool:APP_RECAPTCHA_ENABLED)%"
```
stof, line 193: this could be added to the file-level default bindings, avoiding the need to define the services explicitly.

src/Controller/UserController.php
```diff
@@ -178,7 +178,7 @@ public function viewProfileAction(Request $req)
         }
 
         return $this->container->get('templating')->renderResponse(
```
stof, line 180: should use `$this->render()`, which calls Twig without the `templating` wrapper which is deprecated

config/services.yaml
```diff
@@ -0,0 +1,201 @@
...
+services:
+    # default configuration for services in *this* file
+    _defaults:
+        autowire: true      # Automatically injects dependencies in your services.
+        autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
+        bind:
+            # services
+            $redis: '@snc_redis.default_client'
+            $redisCache: '@snc_redis.cache_client'
+            $instantMailer: '@swiftmailer.mailer.instant'
```
stof, line 28: this looks broken as swiftmailer has been removed

config/packages/framework.yaml
Seldaek: It seems to be needed for FOSUserBundle

config/packages/dev/mailer.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+    mailer:
+        envelope:
+            # TODO this does not seem to work
+            recipients: ['%env(APP_DEV_EMAIL_RECIPIENT)%']
```
Seldaek, line 5: Anyone has a clue here why this would fail?  I have the env var set in .env.local, and yet emails seem to go wherever they want.

Seldaek:
Thanks for the review @stof - fixed everything and the couple other TODOs I had left. Happy to get another round of review now I think it's all done. 

Only problem left that I am aware of is the comment I wrote above about mailer.yaml.

config/packages/dev/mailer.yaml
mbabker: This goes far beyond my understanding of the Mailer component, but...

With Mailtrap set up on SMTP, inside the `SmtpTransport` class when I dump [the recipients](https://github.com/symfony/mailer/blob/v4.4.17/Transport/Smtp/SmtpTransport.php#L198-L200) I get the right address (the one from this config), but when I dump [the message stream](https://github.com/symfony/mailer/blob/v4.4.17/Transport/Smtp/SmtpTransport.php#L203-L205) the body it builds has the original email address added to the email.  That's about as far as I can get before I'm over my head.

config/packages/twig.yaml
```diff
@@ -0,0 +1,15 @@
...
+    default_path: '%kernel.project_dir%/templates'
+    debug: '%kernel.debug%'
+    strict_variables: '%kernel.debug%'
+    exception_controller: null
+    form_themes:
+        - 'forms.html.twig'
+    globals:
+        google_analytics: '%google_analytics%'
+        packagist_host: '%env(APP_HOSTNAME)%'
+        # TODO recaptcha_site_key: '%recaptcha_site_key%'
```
mbabker, line 11:
```suggestion
        recaptcha_site_key: '%env(APP_RECAPTCHA_SITE_KEY)%'
```

config/packages/dev/mailer.yaml
stof: Emails have 2 different concepts: the recipients in the mime messages (the `To` and `Cc` headers) and the recipients in the SMTP protocol (which is where the email gets routed by SMTP).
When using this `recipients` feature, the Mailer component overrides things only at the level of the SMTP envelope, not in the mime headers. So you will still see the original recipients in the headers in your email client, even though emails should be delivered to your own mailbox.

Seldaek: Ah that may explain the problem I saw then, as I tested it with a foo@ alias which lands in my inbox anyway.. Will test overriding with an other domain entirely.

stof: you could look at the `Delivered-To` headers in the received email (which is how these SMTP-level features are surfaced)

PR #1145, opened 2020-12-02
4.4 Upgrade Tweaks

mbabker:
- Added `symfony/lock` component for the lockable trait to work on the worker command
- Checked in `symfony.lock` for Flex
- Set PHPUnit default version at 8.5 (matches current master)
- Replaced `$this->get()` use in controllers with DI (for both params and services)
- Removed local `csrf_token()` Twig function in favor of the one from the framework

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -13,7 +13,7 @@
 namespace App\Controller;
 
 use Doctrine\Persistence\ManagerRegistry;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller as BaseController;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController as BaseController;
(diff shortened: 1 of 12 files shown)


src/Controller/Controller.php
```diff
@@ -13,7 +13,7 @@
 namespace App\Controller;
 
 use Doctrine\Persistence\ManagerRegistry;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller as BaseController;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController as BaseController;
```
apfelbox, line 16: We can remove the alias now, as it doesn't conflict with the `Controller` class below anymore.

Seldaek:
Thanks!

PR #1152, opened 2021-01-13
Add phpstan-doctrine

Seldaek:
Blocked by https://github.com/phpstan/phpstan-doctrine/issues/169

### DIFF
--- a/tests/object-manager.php
+++ b/tests/object-manager.php
@@ -1,0 +1,6 @@
+<?php
+
+require dirname(__DIR__).'/config/bootstrap.php';
+$kernel = new AppKernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
+$kernel->boot();
+return $kernel->getContainer()->get('doctrine')->getManager();
--- a/composer.json
+++ b/composer.json
@@ -88,6 +88,7 @@
     },
     "require-dev": {
         "phpstan/phpstan": "^0.12.66",
+        "phpstan/phpstan-doctrine": "^0.12.27",
         "symfony/browser-kit": "^4.4",
         "symfony/css-selector": "^4.4",
         "symfony/debug-bundle": "^4.4",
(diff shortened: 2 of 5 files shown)


tests/object-manager.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+require dirname(__DIR__).'/config/bootstrap.php';
+$kernel = new AppKernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
```
ondrejmirtes, line 4: There's a typo - needs `App\Kernel` :)

Seldaek:
Thanks again @ondrejmirtes - works fine now except for the `Call to an undefined method Doctrine\Persistence\ObjectManager::getConnection()` nagging me :) There's no way to fix it to see that ManagerRegistry::getManager() returns an EntityManager by any chance?

ondrejmirtes:
@Seldaek Normally you would be able to override it with a custom stub file (https://phpstan.org/user-guide/stub-files) but ManagerRegistry already has a stub in phpstan-doctrine :(

Seldaek:
Ok, not possible to get specialized stubs for EntityManager or the document version for mongo-style ObjectManagers? I don't know enough about internals but I am thinking if it has an instance of the manager via object-manager.php it may be able to load the matching stub corresponding to the concrete type in use in the project?

ondrejmirtes:
Can you point me to the line of code in this repo where PHPStan complains about this?

There's been an issue for supporting multiple EntityManagers through ManagerRegistry for a long time (https://github.com/phpstan/phpstan-doctrine/issues/53) but I don't think we have a solution for that yet. There are some workaround posted as comments... (I'm also not sure it's the same issue you're experiencing.)

Seldaek:
If I up the phpstan level to 2, I get this:

```
  Line   Command/ClearVersionsCommand.php
 ------ -------------------------------------------------------------------------------------------------------
  95     Call to an undefined method Doctrine\Persistence\ObjectManager::getConnection()
```

https://github.com/composer/packagist/blob/main/src/Command/ClearVersionsCommand.php#L95

Seldaek:
The issue you linked is about supporting multiple managers, which is an even more complex setup, but in my case I have just one, and getManager is always going to return an EntityManager and not an ObjectManager

PR #1153, opened 2021-01-18
Package: fix update job failure email

glaubinix:
Package update jobs with a broken branch or failure currently return instead of the actual error and also do not email the maintainers because of `{"status":"errored","message":"An unexpected failure occurred","exception":[],"exceptionMsg":"An address can be an instance of Address or a string (\\\"array\\\") given).\",\"exceptionClass\":\"Symfony\\\\Component\\\\Mime\\\\Exception\\\\InvalidArgumentException\"}"`


Seldaek:
Oh yeah totally missed that one, thanks

PR #1158, opened 2021-02-10
Fix support for replacements as links in search results

### LINKED ISSUE #1157: Support for URLs in package replacements is inconsistent
When passing a URL rather than a package name for the replacement, the package page makes a working link. But the search result page insert the URL as a package name inside the packagist URL.
See https://packagist.org/packages/sensiolabs/security-checker vs https://packagist.org/?query=security-checker the later is linking to `https://packagist.org/packages/https://github.com/fabpot/local-php-security-checker`

stof:
Closes #1157


Seldaek:
Thanks

Issue #1159, opened 2021-02-13
Package fixtures?


BenMorel:
Hi, I'm trying to familiarize myself with the codebase to submit a PR for a new feature. I ran the instructions in the README, so far so good.

But now I'm left with an empty database, and can't seem to find any fixtures. **What's the recommended way to get package / author / ... sample data for a reasonable number of packages?**

The feature I'll be working on is a daily export of all packages, together with their download stats, as proposed in an earlier email.

Thanks!

Seldaek:
We don't really have that sorry, you can create an account and submit a bunch of packages yourself, make sure to run the `bin/console pack:run` command so that their metadata syncs up.

BenMorel:
Thanks, I was afraid that was the case. **Would you be interested in extracting a subset of the database for development purposes?** Like the top 1000 packages, together with their relationships.

I have a tool that can export referentially intact subsets: https://github.com/BenMorel/smartdump
I can provide you with the right command to run to extract just what's needed.

That would be much faster than attempting to create 1000 packages by hand! Committing the SQL output, and updating it from time to time, could help contributions. Password hashes would need to be filtered out in some way, though.

stof:
@BenMorel the source of truth of package metadata is not on packagist but in the VCS repositories being registered. If you register them and run the command fetching metadata from github, you will have them.

BenMorel:
@stof Do you mean?

- create a user account locally
- register existing packages
- run `bin/console pack:run`

stof:
@BenMorel running `bin/console packagist:update` as third step to fetch the metadata from github, before dumping them.

@Seldaek isn't it `packagist:dump` rather than `pack:run` ? I don't find any `pack:run` command in the Packagist codebase (maybe that's a Private Packagist thing ?)

stof:
ah, found it. that's `packagist:run-workers`

BenMorel:
Thank you, I'll look into this! I'll first have to figure out why console commands output a whole load of `[cache] Failed to save key "%5B%5BC%5DApp%5CEntity%5CJob%24packageId%5D%5B1%5D" of type integer.` even though I have Redis running (and can see some keys successfully set in the CLI, like `sess:*`).

Seldaek:
The cache warnings i see as well, i suspect it's a doctrine cache misconfiguration since the 4.4 upgrade but haven't had a chance to dig further. If you find a solution feel free to PR that separately but you can also ignore it otherwise, things run fine anyway.

[...]

PR #1160, opened 2021-02-15
Fixtures

BenMorel:
This is a proposal to add some fixtures, to help new contributors get some data to work with. See #1159 for some discussion.

- [X] Creates a maintainer user
- [X] Creates packages from the top 100 packages of all times, to date (hardcoded)
- [X] Updates the packages
- [X] Creates fake download stats

To be discussed:

- <s>Use HTTP requests to the controllers to create users / packages, instead of duplicating some of the logic in the fixtures? As most of the logic is in controllers and not in a reusable service at the moment.</s> **Update**: I don't think this is a good idea after all, as we need to tap into the domain logic anyway to fake download stats etc.

Please note that this is a first draft to open discussion and see if you'd welcome fixtures.


src/DataFixtures/UserFixtures.php
```diff
@@ -0,0 +1,49 @@
...
+        TokenGeneratorInterface $tokenGenerator
+    ) {
+        $this->userManager = $userManager;
+        $this->tokenGenerator = $tokenGenerator;
+    }
+
+    public function load(ObjectManager $manager)
+    {
+        /** @var User $user */
+        $user = $this->userManager->createUser();
```
stof, line 34: I suggest using `new User()` here for simplicity. There is no reason to account for the customizable User class in project-level code.

src/DataFixtures/PackageFixtures.php
```diff
@@ -0,0 +1,178 @@
...
+            $package->addMaintainer($maintainer);
+            $package->setRepository($url);
+
+            $manager->persist($package);
+            $manager->flush();
+
+            $this->providerManager->insertPackage($package);
+
+            $progressBar->advance();
+            $progressBar->setMessage($url);
```
stof, line 61: shouldn't this message be set when starting to process the URL rather than when we're done with it ?

src/DataFixtures/PackageFixtures.php
```diff
@@ -0,0 +1,178 @@
...
+        $maintainer = $this->getReference(UserFixtures::PACKAGE_MAINTAINER_REFERENCE);
+
+        foreach ($urls as $url) {
+            $package = new Package;
+
+            $package->addMaintainer($maintainer);
+            $package->setRepository($url);
+
+            $manager->persist($package);
+            $manager->flush();
```
stof, line 56: should we really flush in a loop ?

src/DataFixtures/UserFixtures.php
BenMorel: Fixed.

src/DataFixtures/PackageFixtures.php
BenMorel: Good point. I had to manually call `->display()` after `setMessage()` though, or the message would not display before `advance()` anyway!

src/DataFixtures/PackageFixtures.php
BenMorel: Indeed, in this case there's no point.

BenMorel:
I'm reasonably satisfied with the state of this PR now. The fixtures are quite heavy (lots to sync with GitHub, lots of download data to fake & migrate to the db), but it gives a reasonable set of data, complete with download stats.

I'll happily take feedback.

README.md
```diff
@@ -44,3 +44,13 @@ These steps are provided for development purposes only.
...
+
+### Fixtures
+
+You can get test data by running the fixtures:
+
+```bash
+bin/console doctrine:fixtures:load
+ ```
+
+This will create 100 packages from packagist.org, update them from GitHub, populate them with fake download stats, and assign a user named `dev` as their maintainer.
```
Seldaek, line 56:
```suggestion
This will create 100 packages from packagist.org, update them from GitHub, populate them with fake download stats, and assign a user named `dev` (password: `dev`) as their maintainer.
```

src/DataFixtures/DownloadFixtures.php
```diff
@@ -0,0 +1,144 @@
...
+     */
+    private function getLatestPackageVersion(EntityManagerInterface $manager, Package $package): Version
+    {
+        return $manager->createQueryBuilder()
+            ->select('v')
+            ->from(Version::class, 'v')
+            ->where('v.package = :package')
+            ->setParameter('package', $package)
+            ->andWhere('v.normalizedVersion NOT LIKE :dev')
+            ->setParameter('dev', 'dev-%')
```
Seldaek, line 93: You can check for `v.development = false` which is faster as it's backed by an index, and less error prone as this may return a version like 5.x-dev which I assume is not what you wanted.

src/DataFixtures/DownloadFixtures.php
```diff
@@ -0,0 +1,144 @@
...
+            ->getQuery()
+            ->getSingleResult();
+    }
+
+    /**
+     * Populates Redis with pseudo-random daily download stats starting at the package creation date, and ending today.
+     * The algorithm attempts to mimic a typical download stats curve.
+     * The Redis keys set mimic the keys set by the DownloadManager.
+     */
+    private function populateDownloads(Package $package, Version $version): void
```
Seldaek, line 105: This only populates redis, then I guess the expectation is that one runs the `packagist:migrate-download-counts` command to move this to the DB? Possibly should be documented.

Seldaek (changes requested):
Looks good overall yes, thanks a bunch for this. A few comments but nothing major.

Seldaek:
BTW let's please first wrap this up and get it merged, more fixtures might be nice to have but it can come in another PR.

src/DataFixtures/DownloadFixtures.php
BenMorel: Thank you, updated.

src/DataFixtures/DownloadFixtures.php
BenMorel: No, it actually runs the command from within the fixtures after populating redis:

https://github.com/BenMorel/packagist/blob/82bb6a141b6e8a14e7a10f7ff091fa485327e298/src/DataFixtures/DownloadFixtures.php#L78-L79

BenMorel:
Please let me know if you need more changes! :+1:

README.md
BenMorel: Applied, thanks.

Seldaek:
Thanks!

PR #1162, opened 2021-02-15
Short array syntax

BenMorel:
Quite an indigestible PR, but I think it's worth it!

Conversion done using [this script](https://gist.github.com/BenMorel/6994483), if you prefer to run it by yourself.


Seldaek:
Thanks yeah doesn't hurt :)

PR #1167, opened 2021-02-18
Typed properties

BenMorel:
- added missing typed properties
- added a couple missing constructor type hints

I've focused mostly on properties set in constructors (the most obvious ones) in this PR.
I'll propose more typed properties later on, in particular in entities, but these will need careful review to prevent uninitialized property errors wherever a `null` default value is assumed.

### DIFF
--- a/src/EventListener/RegistrationListener.php
+++ b/src/EventListener/RegistrationListener.php
@@ -4,7 +4,7 @@ namespace App\EventListener;
 
 use FOS\UserBundle\Event\FormEvent;
 use FOS\UserBundle\FOSUserEvents;
-use FOS\UserBundle\Util\TokenGenerator;
+use FOS\UserBundle\Util\TokenGeneratorInterface;
 use App\Entity\User;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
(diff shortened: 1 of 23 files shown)


BenMorel:
Hmm, this triggered a lot of phpstan errors. All legitimate, most notably polymorphic calls that were hidden by untyped properties so far. Not sure how you want to handle these?

src/EventListener/RegistrationListener.php
```diff
@@ -13,10 +13,7 @@
  */
 class RegistrationListener implements EventSubscriberInterface
 {
-    /**
-     * @var TokenGenerator
-     */
-    private $tokenGenerator;
+    private TokenGenerator $tokenGenerator;
```
stof, line 16: This should rather use the TokenGeneratorInterface for consistency with other classes

BenMorel: `TokenGenerator` is the type used by the constructor. Should the constructor be changed?

stof: I would say yes. Typehinting the interface is better

BenMorel: Done!

Seldaek:
Would you mind rebasing this? Then the PHPStan errors should be fixed (I can help with that if needed), but for sure moving to typed props is a goal so it's great to get some more of it done :)

BenMorel:
@Seldaek I just rebased, I'll definitely use some help to fix PHPStan errors!

Seldaek:
Thanks :)

BenMorel:
Thanks for the phpstan fixes! :+1:

PR #1179, opened 2021-04-26
Make delete confirmation alerts more descriptive

KevSlashNull:
I noticed that both package delete and version delete buttons show a plain `Are you sure?` confirmation message. IMO it’s better when the alert contains what you are about to delete and that you’re even going to delete that.

Ideally, for the package versions, it would tell the user the name of the version they’re about to delete. That’s because the delete buttons are pretty close to each other, avoiding misclicks.

image

### DIFF
--- a/web/js/view.js
+++ b/web/js/view.js
@@ -162,7 +162,7 @@
     });
     $('.package .delete').on('submit', function (e) {
         e.preventDefault();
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, I added the version name.

KevSlashNull:
Thank you @Seldaek! :smile:

PR #1182, opened 2021-05-05
Upgrade to Symfony 5, drop FOSUserBundle and HWIOAuthBundle

Seldaek:
Reviews appreciated here 😅


src/Security/GitHubAuthenticator.php
```diff
@@ -0,0 +1,141 @@
...
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+
+class GitHubAuthenticator extends SocialAuthenticator
+{
+    private $clientRegistry;
+    private $em;
+    private $router;
+
+    public function __construct(ClientRegistry $clientRegistry, EntityManagerInterface $em, RouterInterface $router)
```
stof, line 25: You should typehint `UrlGeneratorInterface` instead, as you only care about the generator part of the router here

src/Security/Provider/UserProvider.php
```diff
@@ -12,118 +12,59 @@
...
+        return $this->getRepo()->find($user->getId());
     }
 
     /**
      * {@inheritDoc}
      */
     public function supportsClass($class)
     {
-        return $this->userProvider->supportsClass($class);
+        return $class === User::class;
```
stof, line 63: this could break if the class of the serialized token is the doctrine proxy class rather than the entity class directly.

templates/bundles/FOSUserBundle/Profile/edit.html.twig
```diff
@@ -0,0 +1,68 @@
+{% extends "bundles/FOSUserBundle/layout.html.twig" %}
```
stof, line 1: why keeping those templates as overridden FOSUserbundle templates (which are not accessed under the `@FOSUser` namespace anymore anyway) ?

templates/registration/confirmation_email.html.twig
```diff
@@ -0,0 +1,10 @@
+<p>Hi,</p>
+
+<p>
+    Please confirm your email address by clicking the following link: <br><br>
+    <a href="{{ signedUrl|raw }}">Confirm my Email</a>.<br><br>
```
stof, line 5: why `|raw` ? This is wrong. HTML escaping must be applied

templates/registration/confirmation_email.txt.twig
```diff
@@ -0,0 +1,10 @@
+Hi,
+
+Please confirm your email address by clicking the following link:
+
+{{ signedUrl|raw }}
```
stof, line 5: `|raw` is useless. A `.txt.twig` file already disables autoescaping everywhere (plain text does not require escaping)

src/Entity/User.php
```diff
@@ -58,17 +61,22 @@
  *         )
  *     )
  * })
+ * @UniqueEntity(fields={"username"}, message="There is already an account with this username")
```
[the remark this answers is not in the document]
stof: If you want to render the error on the `username` field, use the `errorPath` property of the annotation to customize that.

stof (changes requested):

config/routes/hwioauth_fosub.yaml
```diff
@@ -1,37 +1,16 @@
...
+#hwi_oauth_connect:
+#    resource: '@HWIOAuthBundle/Resources/config/routing/connect.xml'
+#    prefix:   /connect
+#
+## overrides the fosub /login page
+#hwi_oauth_login:
+#    resource: '@HWIOAuthBundle/Resources/config/routing/login.xml'
+#    prefix:   /login
+#
+#hwi_oauth_redirect:
```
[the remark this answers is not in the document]
Seldaek: I did all this in quite a rush this afternoon/evening, so it's definitely in need of a good cleanup still :) That answers most of your questions I think.. but thanks for the quick review I'd love to wrap this up tomorrow.

templates/registration/confirmation_email.html.twig
Seldaek: Yeah interesting, this was in the generated template from makerbundle AFAIK, will check/report there if that's the case.

Seldaek:
Ok I think this is ready to go now. If you have some more time to review @stof that'd be great 👍🏻

PR #1183, opened 2021-05-05
Fix conditions for displaying maintainer forms

### LINKED ISSUE #1181: Unable to add maintainers to any package
When I view maintainers for a package, I see this:
image

The big + to add a maintainer appears to be missing for some reason.

stof:
The negation of a OR is not the OR of the negations. This was broken when switching that condition to include it in the early return (in https://github.com/composer/packagist/commit/c3f88efc5a96003eb38e368da96ce05cace013ec).

Closes #1181


Seldaek:
Thanks

PR #1190, opened 2021-05-17
Update the warning message displayed to composer 1 users

stof:
This now hints at the deprecation of composer 1 support on packagist.org, to raise awareness


Seldaek:
Waiting a  bit longer before merging this, as in practice this isn't confusing people that much it seems, and I rather point them to the 2.0 announcement blog post for now still.

stof:
Well, we could do a middle step: update the message to explicitly tell that some packages won't be available anymore for Composer 1, but keep the link to the announcement (which links to the deprecation blog post at the end anyway).

naderman (approved):

PR #1191, opened 2021-05-17
Update all deps, and bump php version to php8

Seldaek:


composer.json
```diff
@@ -33,24 +33,29 @@
...
         "knpuniversity/oauth2-client-bundle": "^2.8",
         "laminas/laminas-diagnostics": "^1.6",
         "laminas/laminas-feed": "^2.13",
-        "league/oauth2-github": "^2.0",
+        "league/oauth2-github": "^3.0",
         "nelmio/cors-bundle": "^2",
         "nelmio/security-bundle": "^2.4",
-        "pagerfanta/pagerfanta": "^2.4",
+        "pagerfanta/doctrine-orm-adapter": "^3.1",
+        "pagerfanta/pagerfanta": "^3.1",
```
mbabker, line 47: If you're installing the `pagerfanta/pagerfanta` monorepo, you don't need the sub-split packages (ORM adapter and Twig) since those are replaced.

stof: that requirement should probbaly be on `pagerfanta/core` instead

Seldaek: Thanks, fixed

Issue #1194, opened 2021-05-22
`/p2` API endpoint only returns a license key for the first version object


chris48s:
Hello. I've been having a look at the new `/p2` API endpoint and it appears that

- The older `/p` endpoint returns a license key for every version object in the JSON response e.g: https://packagist.org/p/symfony/symfony.json
- The new `/p2` endpoint only returns a license key for the first version object in the array e.g: https://packagist.org/p2/symfony/symfony.json

Is this intended behaviour or a bug?

stof:
That's not actually true. What it does is returning metadata in a minified way, where any key that is identical to the previous version gets omitted (this helps a lot reducing the size of the JSON file, as lots of things don't change between patch versions).
This is part of the v2 format of composer metadata: https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors

If your code is written in PHP, you can use https://github.com/composer/metadata-minifier to expand the metadata to their full form. If you use another language, you will need to port the logic.

PR #1195, opened 2021-05-22
Write the minified key at the beginning of the dumped file

stof:
This make it easier to see it when a human look at the JSON returned by packagist, as the list of versions is quite long.

I got this idea while answering #1194


Seldaek:
Thanks

PR #1201, opened 2021-06-06
Add a "tagPattern" to allow some tags and ignore others

Nyholm:
My issue is that I want to be able to avoid adding some tags to packagist. Ie, I dont want the next major release to be available but I still want to tag patch releases on existing versions. I can also imagine that people are using tags for pre-releases or part of a very custom build process. 

This PR introduce a `$tagPattern` property on the Package where the user will be able to specify a regex. If a **new** tag does not match the regex it will not be added. 

This PR is missing a GUI but I would be happy to see if this feature (or a version of it) would be accepted first. 

### TODO

- [x] Add a GUI for adding a value for `Package::$tagPattern`

### DIFF
--- a/src/Controller/PackageController.php
+++ b/src/Controller/PackageController.php
@@ -1038,6 +1038,7 @@ class PackageController extends Controller
 
         $form = $this->createFormBuilder($package, ["validation_groups" => ["Update"]])
             ->add('repository', TextType::class)
+            ->add('tagPattern', TextType::class, ['required'=>false, 'label'=>'packages.tag_pattern.label'])
             ->setMethod('POST')
             ->setAction($this->generateUrl('edit_package', ['name' => $package->getName()]))
             ->getForm();
--- a/src/Entity/Package.php
+++ b/src/Entity/Package.php
@@ -188,6 +188,11 @@ class Package
      */
     private $suspect;
 
+    /**
+     * @ORM\Column(type="string", length=255, nullable=true)
+     */
+    private $tagPattern;
+
     private $entityRepository;
     private $router;
 
@@ -881,6 +886,25 @@ class Package
     public function getSuspect(): ?string
     {
         return $this->suspect;
+    }
+
+    /**
+     * If defined, tags/versions must match this pattern or they will be ignored.
+     */
+    public function getTagPattern(): ?string
+    {
+        return $this->tagPattern;
+    }
+
+    /**
+     * @param string $tagPattern
+     */
+    public function setTagPattern(string $tagPattern): void
+    {
+        if ($tagPattern === '') {
+            $tagPattern = null;
+        }
+        $this->tagPattern = $tagPattern;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


Seldaek:
I'd really like to get more information about the use case here because from your description it mostly sounds like hypothetical cases you imagine, and I can't imagine many (any?) cases where this is a real issue, and it seems nobody else ran into this in 10 years of operating packagist, so I hope you'll forgive me for being slightly skeptical :)

Nyholm:
Thank you for the quick reply. 

>  I hope you'll forgive me for being slightly skeptical :)

No worries. =)

My exact use case is that I don't want symfony/symfony 6.0 to show up on packagist. I want to discourage users to depend on that package. If some private application really need to, they can still add the following to composer.json: 

```
    "repositories": [
        { "type": "vcs", "url": "https://github.com/symfony/symfony"}
    ],
```

---------------

One alternative would be to mirror the repository like [explained here](https://github.com/composer/composer/discussions/9657), but that would lead to confusion. 

The other alternative would be to not tag 6.0 on symfony/symfony. That would require rewrite of the current release process and the "private applications" mentioned above would be left hanging. But this is my next step if a PR similar to this is not accepted. 

------------

Another non-hypothetical case (but I don't care too much for it) is that this PR could be used to make sure temporary branches do not show up on packagist. Ie, say that someone made a PR from a non-fork. 

Screenshot 2021-06-07 at 06 23 27

Seldaek:
Ok I see.. I do wonder if this really warrants an option on packagist or not. 

I mean is it really that important that people cannot install symfony/symfony ^6 this way? Isn't it likely to backfire, in that the confused people will end up installing symfony/symfony 5.x then instead of 6.x when starting new projects and using `composer require symfony/symfony`? That sounds like an even worse outcome to me..

Nyholm:
> I do wonder if this really warrants an option on packagist or not.

Me too =)

> Isn't it likely to backfire...

I don't think so. If a user is using an outdated way to install Symfony, Im fine with them getting an outdated version.

Seldaek:
Ok I guess this should be acceptable to merge then, but I can't give you a precise timeline as to when I'll have time to take a closer look at the PR. Feel free to add GUI (I guess just a field in the package edit form should be enough), and I'll take a look when I can.

Nyholm:
Thank you. 
Can you please merge any of my other PRs so I'll be approved for running the CI?

If you have a look sometime the next few months I'll be happy. There is no rush

Seldaek:
Ok next few months is definitely doable :) And you're good to go for mining bitcoins now 😄

Nyholm:
I've added a GUI and the tests are green. This PR is ready for a review. 

Screenshot 2021-06-19 at 09 56 23

Nyholm:
I would be super happy to get a review on this PR.

fabpot:
For `symfony/symfony`, I think that we won't need this patch. I will patch the subtree splitter instead to allow it to create tags based on a sha1 instead of a tag. So, feel free to not merge it if it's only for Symfony.

Nyholm:
Okey. cool cool. 

Thank you @Seldaek for your time.

PR #1203, opened 2021-06-14
Fix readme h2 link color

IonBazan:
`.wrapper .content h2 a` selector matches the readme header link field causing it to be dark grey instead of orange like any other links. Instead, it the rule should be amended to match only the package title in page header.

```markdown
## Now available as [GitHub Action](https://packagist.org/)!
```

**Before:**
image

**After:**
image

BTW I think it would be worth to introduce SCSS as there are many nested styles and rules that are repeated. Let me know if I can help with that.

### DIFF
--- a/web/css/main.css
+++ b/web/css/main.css
@@ -1498,7 +1498,7 @@ input:focus:invalid:focus, textarea:focus:invalid:focus, select:focus:invalid:fo
     vertical-align: middle;
 }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, SCSS yeah maybe one day but right now it's not really a top priority for me to overhaul the frontend..

PR #1206, opened 2021-08-09
Added support for GitHub Security Advisories

Toflar:
Here we go - using the GitHub token of one of the maintainers to automatically search the GitHub security advisories database 🎉 

This would allow packagist.org to become the new central place to ask for PHP vulnerabilities [via its already existing API](https://packagist.org/apidoc#list-security-advisories).
Developers who are already using the GitHub advisories feature do not need to create a PR for FriendsOfPHP/security-advisories anymore (unless you want to support software that's fetching data directly from that repo instead of the Packagist API).

I had to update the interface to pass on the package but I thought that's fine as I think that's an internal interface anyway. The same applies for a few other methods.

I have also added a simple way to run Redis as that was a missing step in the README to run the tests :-)

I hope this will further ease the process of reporting PHP vulnerabilities in a standardized manner and help to make the PHP application world a little bit safer 😊

### DIFF
--- a/src/SecurityAdvisory/GitHubSecurityAdvisoriesSource.php
+++ b/src/SecurityAdvisory/GitHubSecurityAdvisoriesSource.php
@@ -1,0 +1,127 @@
+<?php
+
+
+namespace App\SecurityAdvisory;
+
+use App\Entity\Package;
+use Composer\IO\ConsoleIO;
+use GuzzleHttp\Client;
+
+/**
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class GitHubSecurityAdvisoriesSource implements SecurityAdvisorySourceInterface
+{
+    public const SOURCE_NAME = 'GitHub';
+
+    private Client $guzzle;
+
+    public function __construct(Client $guzzle)
+    {
+        $this->guzzle = $guzzle;
+    }
+
+    public function getAdvisories(ConsoleIO $io, Package $package): ?array
+    {
+        $githubToken = null;
+        $maintainers = $package->getMaintainers();
+        foreach ($maintainers as $maintainer) {
+            if ($maintainer->getGithubToken()) {
+                $githubToken = $maintainer->getGithubToken();
+                break;
+            }
+        }
+
+        if (null === $githubToken) {
+            return [];
+        }
+
+        /** @var RemoteSecurityAdvisory[] $advisories */
+        $advisories = [];
+        $hasNextPage = true;
+        $after = '';
+
+        while ($hasNextPage) {
+            $opts = [
+                'headers' => ['Authorization' => 'Bearer ' . $githubToken],
+                'json' => ['query' => $this->getQuery($package->getName(), $after)],
+            ];
+
+            $response = $this->guzzle->request('POST', 'https://api.github.com/graphql', $opts);
+            $data = json_decode($response->getBody()->getContents(), true);
+            $data = $data['data'];
+
+            foreach ($data['securityVulnerabilities']['nodes'] as $node) {
+                $remoteId = null;
+                $cve = null;
+
+                foreach ($node['advisory']['identifiers'] as $identifier) {
+                    if ('GHSA' === $identifier['type']) {
+                        $remoteId = $identifier['value'];
+                        continue;
+                    }
+                    if ('CVE' === $identifier['type']) {
+                        $cve = $identifier['value'];
+                    }
+                }
+
+                if (null === $remoteId) {
+                    continue;
+                }
+
+                if (isset($advisories[$remoteId])) {
+                    $advisories[$remoteId] = $advisories[$remoteId]->withAddedAffectedVersion($node['vulnerableVersionRange']);
+                    continue;
+                }
+
+                $advisories[$remoteId] = new RemoteSecurityAdvisory(
+                    $remoteId,
+                    $node['advisory']['summary'],
+                    $package->getName(),
+                    $node['vulnerableVersionRange'],
+                    $node['advisory']['permalink'],
+                    $cve,
+                    \DateTime::createFromFormat(\DateTimeInterface::ISO8601,  $node['advisory']['publishedAt']),
+                    null
+                );
+            }
+
+            $hasNextPage = $data['securityVulnerabilities']['pageInfo']['hasNextPage'];
+            $after = $data['securityVulnerabilities']['pageInfo']['endCursor'];
+        }
+
+        return array_values($advisories);
+    }
+
+    private function getQuery(string $packageName, string $after = ''): string
+    {
+        if ('' !== $after) {
+            $after = sprintf(',after:"%s"', $after);
+        }
+
+        $query = <<<QUERY
+query {
+  securityVulnerabilities(ecosystem:COMPOSER,package:"%s",first:100%s) {
+    nodes {
+      advisory {
+        summary,
+        permalink,
+        publishedAt,
+        identifiers {
+          type,
+          value
+        }
+      },
+      vulnerableVersionRange
+    },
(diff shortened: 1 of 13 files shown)


nicolas-grekas:
I'm mixed on this: GitHub Security Advisories is a private DB, courtesy of GH, but they could change the terms at anytime in the future. FOSPHP is a public DB, which can be hosted anywhere else if GH becomes unreliable for hosting repos. That's not going to be possible for the GSA DB.

Right now, composer relies on a single source for security advisories, and that gives an incentive to populate this public DB. By reading from GSA, we're going to move the info under private terms. I don't see this as a good thing on the long run.

Toflar:
I totally get this argument and it indeed is a possible issue we should consider.
The other side of the coin is that from my experience talking to less experienced developers, they often do not report anything at FriendsOfPHP because of multiple reasons:

a) The typical insecurity when contributing to a repository managed by people they don't know.
b) Understanding the YAML format over there is more complicated than a GUI like GH provides.
c) They do not care about reporting it because it takes even more time on top of already managing GSA.

Sure, the centralized repository we have now has its advantages. It just doesn't really help us in any way if people do not report vulnerabilities at all. Supporting multiple sources and aggregating them at Packagist has the advantage that we could incorporate even more sources than just FOPHP and GH. Many PHP projects have their own platforms and environments to report security issues.

The core statement of what I'm trying to say is: I'd rather have the same vulnerability reported at 5 sources than not at all.

m-vo:
>  GitHub Security Advisories is a private DB, courtesy of GH, but they could change the terms at anytime in the future.

I think it's important that there is an upgrade path in that case. But the data could still get pulled into `FriendsOfPHP/security-advisories` at that point, couldn't it?

An other option would be an automated scheduled import from GSA to FOPHP. The strategy from this PR IMHO has less complications, though.

nicolas-grekas:
> the data could still get pulled into FriendsOfPHP/security-advisories at that point, couldn't it?

I think it'd be interesting to investigate this question. Would anyone be up to dig it?

src/Service/SecurityAdvisoryWorker.php
```diff
@@ -35,6 +36,14 @@ public function __construct(Locker $locker, LoggerInterface $logger, ManagerRegi
...
         $sourceName = $job->getPayload()['source'];
+        $package = $this->doctrine->getRepository(Package::class)->findOneBy(['id' => $job->getPackageId()]);
+
+        if (!$package instanceof Package) {
+            $this->logger->info('Security advisory update failed, skipping', ['packageId' => $job->getPackageId()]);
+
+            return ['status' => Job::STATUS_ERRORED, 'message' => 'Security advisory update failed, skipped'];
+        }
+
         $lockAcquired = $this->locker->lockSecurityAdvisory($sourceName);
```
glaubinix, line 47: Again, didn't really have something like GitHub in mind when building this. This currently only allows a single GitHub package to be processed at once. This will probably lead to a backlog of jobs. This should probably differ whether the source handles multiple package or just a single package and get a different lock key.

src/SecurityAdvisory/GitHubSecurityAdvisoriesSource.php
```diff
@@ -0,0 +1,127 @@
...
+        $hasNextPage = true;
+        $after = '';
+
+        while ($hasNextPage) {
+            $opts = [
+                'headers' => ['Authorization' => 'Bearer ' . $githubToken],
+                'json' => ['query' => $this->getQuery($package->getName(), $after)],
+            ];
+
+            $response = $this->guzzle->request('POST', 'https://api.github.com/graphql', $opts);
```
glaubinix, line 50: This should probably catch `GuzzleException` and log the result to avoid having the worker crash

src/Service/UpdaterWorker.php
```diff
@@ -270,7 +271,9 @@ public function process(Job $job, SignalHandler $signal): array
         }
 
         if ($packageName === FriendsOfPhpSecurityAdvisoriesSource::SECURITY_PACKAGE) {
-            $this->scheduler->scheduleSecurityAdvisory(FriendsOfPhpSecurityAdvisoriesSource::SOURCE_NAME);
+            $this->scheduler->scheduleSecurityAdvisory(FriendsOfPhpSecurityAdvisoriesSource::SOURCE_NAME, $id);
+        } elseif (preg_match('{[@/]github.com[:/]([^/]+/[^/]+?)(\.git)?$}i', $package->getRepository())) {
+            $this->scheduler->scheduleSecurityAdvisory(GitHubSecurityAdvisoriesSource::SOURCE_NAME, $id);
```
glaubinix, line 276: Problem with calling this at the very end of the package update is that it won't be called if any issue is detected during the package update e.g. an invalid composer.json in a single branch. I think this is less of an issue with the friendsofphp package because we can kind of assume that their composer.json is always ok. But might have to move all of this up a bit

glaubinix (commented):
Fetching the GitHub advisories works great! Needs some adjustments to make the current system work with a source that can be triggered by multiple packages.

Before this can get merged, we need to somehow deduplicate advisories that are on GitHub and part of the friendsofphp package that have the same CVE.

naderman:
Deduplication should happen on packagist.org, that was always the idea when this was built with multiple sources in mind. Consumers of the API (potentially composer cli) should not need to figure this out themselves.

Toflar:
Okay, what's the logic then? :)

naderman:
@Toflar not sure what you mean?

Toflar:
Well, I guess if the CVE matches, we can deduplicate. What if there's no CVE? If the `title` is the same? Or `link`? 
And who wins? GitHub or FriendsOfPHP (or ...)?

naderman:
See https://github.com/composer/packagist/pull/1215/files#diff-4a9aa46b4f768c98a1f6c8d7f3647aff7283c196c7ee078f0a0a1f3a3749da8aR153 - like you said, some things are clear, e.g. matching or mismatching CVE id, in other cases we'll compute a similarity score.

Toflar:
Ah, then we wait until that is merged I guess?

Seldaek:
The other PR is merged, so if you ever find time to apply fixes here @Toflar that'd be cool!

src/Service/SecurityAdvisoryWorker.php
```diff
@@ -44,7 +53,7 @@ public function process(Job $job, SignalHandler $signal): array
 
         /** @var SecurityAdvisorySourceInterface $source */
         $source = $this->sources[$sourceName];
-        $remoteAdvisories = $source->getAdvisories($io);
+        $remoteAdvisories = $source->getAdvisories($io, $package);
         if (null === $remoteAdvisories) {
             $this->logger->info('Security advisory update failed, skipping', ['source' => $source]);
```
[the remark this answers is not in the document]
Toflar: So what do you want me to do here then? I don't have any insights into how things run on production or how you guys would like to have it work...so I don't know what's the best approach here. Maybe `main` should be prepared to support this use case first before I adjust this PR?

Toflar:
Well it's not really just "fixes". The whole concept of multiple sources supporting the same packages is missing in `main` according to @glaubinix. That needs to be addressed first imho.

src/Service/SecurityAdvisoryWorker.php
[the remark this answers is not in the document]
glaubinix: The issue here is that `$existingAdvisories = $this->doctrine->getRepository(SecurityAdvisory::class)->findBy(['source' => $sourceName]);` will return all advisories for all packages for a specific source, in your case GitHub, but the advisory job only fetches them for a single package.

This means if you have two packages with advisories, they will each delete each others advisories.

Seldaek:
Closed by #1264

PR #1210, opened 2021-09-24
use session to store 2fa secret between requests

IonBazan:
Prior to this change, user was able to change the 2FA secret into any value by manipulating hidden form field which might be a security concern. This can be fixed by temporarily using session to store the secret between requests. Please note it still generates a new secret each time you refresh the page.

I also changed the QR code writer to SVG for better resolution on Retina screens. Thanks to `user-select: all`, secret is fully selected with a single click which makes it even easier to copy, especially on mobile devices.

### DIFF
--- a/src/Form/Model/EnableTwoFactorRequest.php
+++ b/src/Form/Model/EnableTwoFactorRequest.php
@@ -6,32 +6,8 @@ use Symfony\Component\Validator\Constraints as Assert;
 
 class EnableTwoFactorRequest
 {
-    /**
-     * @var string
-     * @Assert\NotBlank
-     */
-    protected $secret;
-
-    /**
-     * @var ?string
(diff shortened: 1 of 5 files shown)


src/Form/Model/EnableTwoFactorRequest.php
```diff
@@ -6,32 +6,8 @@
...
-    }
-
-    public function setSecret(string $secret): void
-    {
-        $this->secret = $secret;
-    }
+    #[Assert\NotBlank]
+    protected ?string $code = null;
 
     public function getCode(): ?string
```
IonBazan, line 12: Not sure if we still need it here. We could use a typed public property instead.

Seldaek:
Thanks, all good improvements :)

PR #1226, opened 2021-12-22
Allow `align` attribute rendering READMEs

mondeja:
Github accepts the attribute `align` for element alignment. See the differences in elements alignment at the start between [this README on Github](https://github.com/simple-icons/simple-icons/blob/06f09a265d3c9d3dee994f3fdcb245f115e840e1/README.md) and [the same on Packagist](https://packagist.org/packages/simple-icons/simple-icons).

I'm using here the same implementation used by [Python's PYPI README renderer](https://github.com/pypa/readme_renderer).


Seldaek:
Thanks

PR #1229, opened 2022-01-18
Update all deps, update to Symfony 5.4

Seldaek:


src/Entity/Version.php
```diff
@@ -154,47 +155,49 @@ class Version
     /**
      * @ORM\Column(type="text", nullable=true)
      */
-    private $source;
+    private string|null $source;
```
stof, line 158: should probably have `= null` for all nullable fields

Seldaek: Right, I always forget those..

BTW related to this, do you know if it's safe to type the OneToMany & co as `private ArrayCollection $foo` or is there a case where doctrine will initialize with something that's not compatible?

I guess the most accurate type would be `Doctrine\Common\Collections\Collection & Doctrine\Common\Collections\Selectable` but that'll require 8.1.

stof: When loading from the DB, the ORM will use a PersistentCollection, not an ArrayCollection.

you need to use the Collection interface as type, not an implementation of it.

Seldaek: Alright, hopefully the latest iteration is correct. All properties of Version are typed now, PHPStan is happy, we'll need to see if prod agrees 😄

PR #1231, opened 2022-01-21
Autofocus instantsearch input on Browse packages page

94noni:
Small shortcut when browsing explore/ (Browse) page
(also remove the old app_dev.php)
ℹ️ not tested locally


Seldaek:
OK why not I guess, thanks.

PR #1232, opened 2022-01-21
phpstan-dba: improve coverage

staabm:


src/Entity/PackageRepository.php
```diff
@@ -477,6 +480,11 @@ public function getSuggestCount($name): int
         return (int) $this->getEntityManager()->getConnection()->fetchOne($sql, $args);
     }
 
+    /**
+     * @param string $name
+     * @param int $offset
+     * @param int $limit
```
stof, line 486: you could use actual parameter types rather than phpdoc

Seldaek: Yes please go wild with PHP 8 syntax in this repo 😄

src/Entity/PackageRepository.php
```diff
@@ -469,6 +469,9 @@ public function getDependents(string $name, int $offset = 0, int $limit = 15, st
         return $this->getEntityManager()->getConnection()->fetchAllAssociative($sql, $args);
```
Seldaek, line 469: Interesting failure here, it seems to pick up the conditional d.total usage but not the join declaring it? TBH this is quite messy code and I don't expect perfect handling here, but it should at least ignore it rather than error I guess.

src/Package/Updater.php
```diff
@@ -274,11 +274,12 @@ public function update(IOInterface $io, Config $config, Package $package, VcsRep
     }
 
     /**
-     * @return array with keys:
-     *                    - updated (whether the version was updated or needs to be marked as updated)
-     *                    - id (version id, can be null for newly created versions)
-     *                    - version (normalized version from the composer package)
-     *                    - object (Version instance if it was updated)
+     * @return array{updated: bool, id: int|null, version: string, object: Version|null}
```
Seldaek, line 277: Maybe better would be `array{updated: true, id: int|null, version: string, object: Version}|array{updated: false, id: int|null, version: string, object: null}`. It's a bit messy with duplication but it would fix the error above I hope.

src/Entity/PackageRepository.php
staabm: will investigate in https://github.com/staabm/phpstan-dba/issues/188

Seldaek: Cool, no rush, glad to provide this repo as playground :) I have another project with hundreds of pure SQL queries I could run the extension on once it matured a bit, but sadly it's private.

src/Package/Updater.php
staabm: ok, lets see whether this works

PR #1233, opened 2022-01-23
Fixing PhpStan issues for Service namespace

chr-hertel:
Couldn't get rid of all due to the contract of `*Worker::process(..):array`
Maybe that's sth to improve in a follow up.

Let me know if you consider the changes not being defensive enough :)

### DIFF
--- a/src/Service/GitHubUserMigrationWorker.php
+++ b/src/Service/GitHubUserMigrationWorker.php
@@ -31,6 +31,21 @@ class GitHubUserMigrationWorker
     ) {
     }
 
+    /**
+     * @return array{
+     *     status: string,
+     *     message: string,
+     *     after?: \DateTime,
+     *     results?: array{
+     *         hooks_setup: int,
+     *         hooks_failed: array<int, array{
+     *             package: string,
+     *             reason: mixed
+     *         }>,
+     *         hooks_ok_unchanged: int
+     *     }
+     * }
+     */
     public function process(Job $job, SignalHandler $signal): array
     {
         $em = $this->getEM();
@@ -78,7 +93,7 @@ class GitHubUserMigrationWorker
         ];
     }
 
-    public function setupWebHook(string $token, Package $package)
+    public function setupWebHook(string $token, Package $package): null|bool|string
     {
         if (!Preg::isMatch('#^(?:(?:https?|git)://([^/]+)/|git@([^:]+):)(?P<owner>[^/]+)/(?P<repo>.+?)(?:\.git|/)?$#', $package->getRepository(), $match)) {
             return null;
@@ -193,6 +208,19 @@ class GitHubUserMigrationWorker
         return true;
     }
 
+    /**
+     * @return array<int, array{
+     *     id: int,
+     *     name: string,
+     *     active: bool,
+     *     events: string[],
+     *     config: array{
+     *         url: string,
+     *         secret: string
+     *     },
+     *     updated_at: string
+     * }>
+     */
     private function getHooks(string $token, string $repoKey): array
     {
         $hooks = [];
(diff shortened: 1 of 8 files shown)


src/Service/GitHubUserMigrationWorker.php
```diff
@@ -213,6 +241,9 @@ private function getHooks(string $token, string $repoKey): array
         return $hooks;
     }
 
+    /**
+     * @param array<int|string, mixed> $json
```
Seldaek, line 245:
```suggestion
     * @param array<int|string, mixed>|null $json
```

src/Service/Locker.php
```diff
@@ -2,69 +2,67 @@
...
         $this->getConn()->fetchOne(
             'SELECT RELEASE_LOCK(:id)',
             ['id' => $command]
         );
     }
 
-    private function getConn()
+    private function getConn(): Connection
     {
         return $this->getEM()->getConnection();
```
Seldaek, line 67: The code above was written with the intent that this would return a PrimaryReadReplicaConnection potentially, and hence the `'master'` arg to force-connect the primary instance. IMO this should be left as is for safety, because while it is currently not using a replica setup, if we were to do it at some point this code would start breaking in weird ways. So it'd be best IMO to type the return here as PrimaryReadReplicaConnection and leave the rest untouched (except you can change master to primary now)

src/Service/UpdaterWorker.php
```diff
@@ -71,6 +71,9 @@ public function __construct(
         $this->downloadManager = $downloadManager;
     }
 
+    /**
+     * @return array{status: string, message?: string, after?: \DateTime, vendor?: string, details?: string, e?: \Exception|\Throwable}
```
Seldaek, line 75:
```suggestion
     * @return array{status: string, message?: string, after?: \DateTimeInterface, vendor?: string, details?: string, e?: \Throwable}
```

Seldaek:
Looks good overall, thanks, just a few tweaks needed.

src/Service/Locker.php
chr-hertel: Currently this returns `Doctrine\DBAL\Connection` and would break with runtime error
Bildschirmfoto von 2022-01-23 13-54-21

keeping the `Connection` type would than raise:
```
Method Doctrine\DBAL\Connection::connect() invoked with 1 parameter, 0 required.
```

would you prefer to keep this warning in favor of switching to replica setup later?

Seldaek: Hmm right :/ I guess the best way is to wrap all the calls above in a `if ($conn instanceof PrimaryReadReplicaConnection) { connect to  primary }` then?

Seldaek: Because if the connection is a regular one, calling connect is completely useless here :)

src/Service/UpdaterWorker.php
```diff
@@ -72,7 +72,7 @@ public function __construct(
     }
 
     /**
-     * @return array{status: string, message?: string, after?: \DateTime, vendor?: string, details?: string, e?: \Exception|\Throwable}
+     * @return array{status: string, message?: string, after?: \DateTimeInterface, vendor?: string, details?: string, e?: \Exception|\Throwable}
```
Seldaek, line 75: You missed that I dropped `Exception` here, it's AFAIK included in Throwable so redundant.

src/Service/Locker.php
```diff
@@ -2,69 +2,76 @@
...
             ['id' => $command]
         );
     }
 
-    private function getConn()
+    private function connect(): void
+    {
+        $connection = $this->getConn();
+        if ($connection instanceof PrimaryReadReplicaConnection) {
+            $connection->connect('primary');
```
stof, line 70: this should use `$connection->ensureConnectedToPrimary()` instead of the legacy API of passing an argument to `connect`

Seldaek:
Thanks

PR #1234, opened 2022-01-23
Replace AboutController by FrameworkBundle features

chr-hertel:
Thought this would be more sensible on the long run instead of updating the controller to attibutes etc.


Seldaek:
Makes sense thanks

PR #1235, opened 2022-01-23
Fixing Model namespace for PhpStan

chr-hertel:
For now I left two in the baseline
```
-
	message: "#^Parameter \\#2 \\$members of method Predis\\\\ClientInterface\\:\\:sadd\\(\\) expects array, string given\\.$#"
	count: 1
	path: src/Model/ProviderManager.php

-
	message: "#^Method App\\\\Model\\\\RedisAdapter\\:\\:getNbResults\\(\\) should return int\\<0, max\\> but returns int\\.$#"
	count: 1
	path: src/Model/RedisAdapter.php
```


src/Model/RedisAdapter.php
```diff
@@ -12,33 +12,25 @@
 
 namespace App\Model;
 
+use App\Entity\User;
 use Pagerfanta\Adapter\AdapterInterface;
 
 /**
  * @author Jordi Boggiano <j.boggiano@seld.be>
  */
 class RedisAdapter implements AdapterInterface
```
chr-hertel, line 21: Let me know if you miss the generic in here ... ;)

stof: IIRC, pagerfanta added generics on the Pagerfanta class, but not yet on the AdapterInterface

mbabker: https://github.com/BabDev/Pagerfanta/commit/7fb9610d0a81f2c0b14abddd4f0bc82054da79cc added them on the adapter interface, I just need to remember to release it when I get home tonight.

stof: oh great. This commit was not there last time I checked (and doing something like that was on my TODO-list)

src/Model/PackageManager.php
```diff
@@ -143,13 +134,14 @@ public function deletePackage(Package $package)
...
+    public function notifyUpdateFailure(Package $package, \Exception $e, string $details = null): bool
     {
         if (!$package->isUpdateFailureNotified()) {
             $recipients = [];
             foreach ($package->getMaintainers() as $maintainer) {
-                if ($maintainer->isNotifiableForFailures()) {
-                    $recipients[$maintainer->getEmail()] = new Address($maintainer->getEmail(), $maintainer->getUsername());
+                $mail = $maintainer->getEmail();
+                if ($mail && $maintainer->isNotifiableForFailures()) {
+                    $recipients[$maintainer->getEmail()] = new Address($mail, $maintainer->getUsername() ?? '');
```
Seldaek, line 144:
```suggestion
                    $recipients[$mail] = new Address($mail, $maintainer->getUsername() ?? '');
```

Seldaek:
Thanks!

src/Model/RedisAdapter.php
Seldaek: Implemented this & a couple more related fix I spotted here in https://github.com/composer/packagist/commit/951d9e97ccdf9360c2b0e3bd77f2354937f601b6

Seldaek:
> message: "#^Parameter \\#2 \\$members of method Predis\\\\ClientInterface\\:\\:sadd\\(\\) expects array, string given\\.$#"

Fixed in 951d9e9

> message: "#^Method App\\\\Model\\\\RedisAdapter\\:\\:getNbResults\\(\\) should return int\\<0, max\\> but returns int\\.$#"

This IMO needs fixing in Predis type defs directly

PR #1237, opened 2022-01-24
Fixing file path in algolia:configure command

chr-hertel:
Possibly a bug introduced with #1144, ran into it while setting up algolia.

Messed up the diff though.
Bug was the relative path here: 
```php
file_get_contents(__DIR__.'/../config/algolia_settings.yml')
```


Seldaek:
Thanks yeah this command is kinda useless to me now :)

PR #1240, opened 2022-01-27
Fixing PHPStan issues in various smaller namespaces

chr-hertel:
This time various smaller namespaces

### DIFF
--- a/src/Security/EmailVerifier.php
+++ b/src/Security/EmailVerifier.php
@@ -17,15 +17,11 @@ class EmailVerifier
 {
     use DoctrineTrait;
 
-    private $verifyEmailHelper;
-    private $mailer;
-    private ManagerRegistry $doctrine;
-
-    public function __construct(VerifyEmailHelperInterface $helper, MailerInterface $mailer, ManagerRegistry $doctrine)
-    {
-        $this->verifyEmailHelper = $helper;
-        $this->mailer = $mailer;
-        $this->doctrine = $doctrine;
+    public function __construct(
+        private VerifyEmailHelperInterface $verifyEmailHelper,
+        private MailerInterface $mailer,
+        private ManagerRegistry $doctrine,
+    ) {
     }
 
     public function sendEmailConfirmation(string $verifyEmailRouteName, UserInterface $user, TemplatedEmail $email): void
@@ -34,10 +30,15 @@ class EmailVerifier
             throw new \UnexpectedValueException('Expected '.User::class.', got '.get_class($user));
         }
 
+        $mail = $user->getEmail();
+        if (null === $mail) {
+            throw new \InvalidArgumentException('User needs to have an email address set.');
+        }
+
         $signatureComponents = $this->verifyEmailHelper->generateSignature(
             $verifyEmailRouteName,
-            $user->getId(),
-            $user->getEmail(),
+            (string) $user->getId(),
+            $mail,
             ['id' => $user->getId()]
         );
 
@@ -60,7 +61,12 @@ class EmailVerifier
             throw new \UnexpectedValueException('Expected '.User::class.', got '.get_class($user));
         }
 
-        $this->verifyEmailHelper->validateEmailConfirmation($request->getUri(), $user->getId(), $user->getEmail());
+        $mail = $user->getEmail();
+        if (null === $mail) {
+            throw new \InvalidArgumentException('User needs to have an email address set.');
+        }
+
+        $this->verifyEmailHelper->validateEmailConfirmation($request->getUri(), (string) $user->getId(), $mail);
 
         if (!$user->hasRole('ROLE_SPAMMER')) {
             $user->setEnabled(true);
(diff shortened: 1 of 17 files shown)


phpstan.neon
```diff
@@ -31,6 +31,9 @@ parameters:
     doctrine:
         objectManagerLoader: tests/object-manager.php
 
+    typeAliases:
+        AwsMetadata: 'array{primary?: mixed, has_instance_store?: mixed, ec2_node?: mixed, region?: mixed}'
```
Seldaek, line 35: This would be the correct type :) 

```suggestion
        AwsMetadata: 'array{}|array{ec2_node: string, ip: string, region: string, primary: bool, has_instance_store: bool, is_worker: bool, is_web: bool}'
```

chr-hertel: but why `array{}|array{...}`?

src/Security/EmailVerifier.php
```diff
@@ -34,10 +30,15 @@ public function sendEmailConfirmation(string $verifyEmailRouteName, UserInterfac
             throw new \UnexpectedValueException('Expected '.User::class.', got '.get_class($user));
         }
 
+        $mail = $user->getEmail();
+        if (null === $mail) {
+            throw new \InvalidArgumentException('User needs to have an email address set.');
+        }
```
Seldaek, line 36: If you rebase those checks aren't needed anymore as getEmail always returns string

src/Security/GitHubAuthenticator.php
```diff
@@ -79,12 +79,14 @@ public function authenticate(Request $request): Passport
                     return $existingUser;
                 }
 
-                if ($userRepo->findOneBy(['usernameCanonical' => mb_strtolower($ghUser->getNickname())])) {
-                    throw new AccountUsernameExistsWithoutGitHubException($ghUser->getNickname());
+                $nickname = $ghUser->getNickname();
```
Seldaek, line 82: Here too a rebase should help some

src/Security/TwoFactorAuthManager.php
```diff
@@ -64,13 +56,13 @@ public function enableTwoFactorAuth(User $user, string $secret)
         $message = (new Email())
             ->subject('[Packagist] Two-factor authentication enabled')
             ->from(new Address($this->options['from'], $this->options['fromName']))
-            ->to($user->getEmail())
+            ->to((string) $user->getEmail())
```
Seldaek, line 59: Not needed after rebase

src/Security/TwoFactorAuthManager.php
```diff
@@ -95,13 +87,13 @@ public function disableTwoFactorAuth(User $user, string $reason)
         $message = (new Email())
             ->subject('[Packagist] Two-factor authentication disabled')
             ->from(new Address($this->options['from'], $this->options['fromName']))
-            ->to($user->getEmail())
+            ->to((string) $user->getEmail())
```
Seldaek, line 90: Same

src/SecurityAdvisory/FriendsOfPhpSecurityAdvisoriesSource.php
```diff
@@ -67,7 +67,10 @@ public function getAdvisories(ConsoleIO $io): ?array
             $finder->name('*.yaml');
             $advisories = [];
             foreach ($finder->in($localDir) as $file) {
-                $content = Yaml::parse(file_get_contents($file->getRealPath()));
+                if (!$file->getRealPath() || !$yaml = file_get_contents($file->getRealPath())) {
```
Seldaek, line 70:
```suggestion
                if (!$file->getRealPath() || !($yaml = file_get_contents($file->getRealPath()))) {
```

phpstan.neon
Seldaek: Because it's either an empty array or all keys exist, so I think this is preferable to having all keys being optional independently.

Also I'd usually prefer a type def on the class rather than aliases in the phpstan config, keeps things closer together.

chr-hertel: Yup, started with that. But this one is used at multiple classes. I can easily demote this to class level, If you prefer that way.

Good that we talk about it :D was working on VersionData and VersionDataV2 based on the methods of version entity on global level.

Maybe a separate phpstan-types.yml as include?

Seldaek: You can import types from another class, if it's not spread like all over the place IMO this is still clearer than types in an external config.

Seldaek: See phpstan-import-type at https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases

chr-hertel: True, i'll rebase and rework later 👍

src/Security/GitHubAuthenticator.php
chr-hertel: nope, no change here

Bildschirmfoto von 2022-01-30 16-59-02

do i miss sth?

Seldaek: No this didn't change but the githubauthenticator I already stored nickname somewhere, but maybe it's some other place sorry

chr-hertel: ah, you're right. yup, missed that. moved the check in 20e1a196f58ace847bb1d6028d3803bcfe7303ce

chr-hertel:
Pipeline will be green after #1243

Seldaek:
Thanks!

PR #1241, opened 2022-01-29
SecurityAdvisory: handle temporary CVE-2022-xxx format

glaubinix:
Adjusted the security advisory handling for temporary CVE ids as seen in https://github.com/FriendsOfPHP/security-advisories/pull/607/files

I wonder whether the title comparison should/can be done a bit more generic?

### DIFF
--- a/src/Entity/SecurityAdvisory.php
+++ b/src/Entity/SecurityAdvisory.php
@@ -171,7 +171,7 @@ class SecurityAdvisory
             $increase = 1;
 
             // Do not increase the score if the title was just renamed to add a CVE e.g. from CVE-2022-xxx to CVE-2022-99999999
-            if (Preg::match('#^(CVE-[0-9]{4}-[x]+:)(.*)$#', $this->getTitle(), $matches) && str_contains($advisory->getTitle(), $matches[2])) {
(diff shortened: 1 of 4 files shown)


naderman:
@glaubinix I'd probably match any CVE id there? As we consider CVE id change a definitive break, the score for changing the title to modify a CVE id in there will only matter if the actual CVE id doesn't change, so e.g. when the title gets fixed cause the id in it was wrong?

src/Entity/SecurityAdvisory.php
```diff
@@ -162,7 +168,14 @@ public function calculateDifferenceScore(RemoteSecurityAdvisory $advisory): int
         }
 
         if ($advisory->getTitle() !== $this->getTitle()) {
-            $score++;
+            $increase = 1;
+
+            // Do not increase the score if the title was just renamed to add a CVE e.g. from CVE-2022-xxx to CVE-2022-99999999
+            if (Preg::match('#^(CVE-[0-9]{4}-[x]+:)(.*)$#', $this->getTitle(), $matches) && str_contains($advisory->getTitle(), $matches[2])) {
```
Seldaek, line 174: Adding support for more invalid CVE patterns, trying to be creative, hopefully this catches most placeholders people may use :)

```suggestion
            if (Preg::match('#^(CVE-[0-9x*?]{1,4}-?[0x*?]+:)(.*)$#', $this->getTitle(), $matches) && str_contains($advisory->getTitle(), $matches[2])) {
```

PR #1242, opened 2022-01-30
SecurityAdvisory: improve title matching

glaubinix:
Addressing: https://github.com/composer/packagist/pull/1241#issuecomment-1024993233

### DIFF
--- a/src/SecurityAdvisory/AdvisoryParser.php
+++ b/src/SecurityAdvisory/AdvisoryParser.php
@@ -13,7 +13,7 @@ class AdvisoryParser
 
     public static function titleWithoutCve(string $title): string
     {
-        if (Preg::match('#^(CVE-[0-9x*?]{1,4}-?[0-9a-z-]+:)(.*)$#i', $title, $matches)) {
+        if (Preg::match('#^(CVE-[0-9a-z*?-]+:)(.*)$#i', $title, $matches)) {
             return trim($matches[2]);
         }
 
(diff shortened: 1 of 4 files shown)


src/SecurityAdvisory/AdvisoryParser.php
```diff
@@ -0,0 +1,22 @@
...
+class AdvisoryParser
+{
+    public static function isValidCve(?string $cve): bool
+    {
+        return $cve && Preg::match('#^CVE-[0-9]{4}-[0-9]{4,}$#', $cve);
+    }
+
+    public static function titleWithoutCve(string $title): string
+    {
+        if (Preg::match('#^(CVE-[0-9x*?]{1,4}-?[0-9a-z-]+:)(.*)$#i', $title, $matches)) {
```
Seldaek, line 16: The point of my `[0-9x*?]{1,4}-?[0x*?]+` changes was to support `*`, `?` and `0` as placheolder as well as `x`. I guess you need to add `*?` again to the second character class to make sure this works? Also probably can just simplify this a lot..

```suggestion
        if (Preg::match('#^(CVE-[0-9a-z*?-]+:)(.*)$#i', $title, $matches)) {
```

glaubinix: oh, I see. Yeah, hard to predict what people will use in the future because they can't just reuse xxxx all year long.

Seldaek: Yes, always assume the user is drunk :p

PR #1248, opened 2022-02-02
update phpstan-dba

staabm:


staabm:
@Seldaek I think it works as expected with the latest release from today morning.

I am not sure why composer downgraded `lcobucci/jwt` though

Seldaek:
Thanks

PR #1252, opened 2022-02-04
phpstan-dba: enable stringifyTypes

staabm:
requires at least phpstan-dba 0.2.13, but because of fixes in the latest release you should use the most recent release anyway.

required since doctrine by default returns everything as a string, see e.g. https://github.com/composer/packagist/blob/233cddb42f93ee29a7ea6fb0211017177db0571a/src/Entity/PackageRepository.php#L256

### DIFF
--- a/phpstan-bootstrap.php
+++ b/phpstan-bootstrap.php
@@ -15,6 +15,7 @@ $cacheFile = __DIR__.'/.phpstan-dba.cache';
 $config = new RuntimeConfiguration();
 $config->stringifyTypes(true); // TODO remove when upgrading to PHP 8.1
 // $config->debugMode(true);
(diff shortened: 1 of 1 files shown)


phpstan-bootstrap.php
```diff
@@ -15,6 +15,7 @@
 $config = new RuntimeConfiguration();
 $config->stringifyTypes(true); // TODO remove when upgrading to PHP 8.1
 // $config->debugMode(true);
+$config->stringifyTypes(true);
```
chr-hertel, line 18: why twice? see line 16

staabm: Oh lol

PR #1255, opened 2022-02-06
readding custom chart css rules

chr-hertel:
after 89a051568141faab8258571d6274cbb45687ff30 I had first this ...

Bildschirmfoto von 2022-02-06 15-49-05

and after removing the broken reference, i got this ...

Bildschirmfoto von 2022-02-06 16-07-04

so i readded a `charts.css` - not sure if that's the right way to address this ...

### DIFF
--- a/css/charts.css
+++ b/css/charts.css
@@ -1,0 +1,102 @@
+.chart {
+    min-height: 300px
+}
+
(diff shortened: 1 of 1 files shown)


Seldaek:
Oh sorry, forgot a file :D

PR #1261, opened 2022-02-20
Fixing PHPStan issues in Command namespace

chr-hertel:
This includes the proposal to use `__CLASS__` as unique identifier per command for `Locker` methods instead of `getName()` as this is nullable.

Promoted the `ini_set` issues to `phpstan.neon` - could be removed with php 8.1 update

While regerating the baseline a lot of issues dropped out of the baseline which i cannot take credit for. potentially caused by a missing update or phpstan update.

### DIFF
--- a/src/Command/CleanIndexCommand.php
+++ b/src/Command/CleanIndexCommand.php
@@ -59,7 +59,7 @@ class CleanIndexCommand extends Command
             return 0;
         }
 
-        $lockAcquired = $this->locker->lockCommand($this->getName());
+        $lockAcquired = $this->locker->lockCommand(__CLASS__);
         if (!$lockAcquired) {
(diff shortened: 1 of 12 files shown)


src/Command/CleanIndexCommand.php
```diff
@@ -59,7 +59,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             return 0;
         }
 
-        $lockAcquired = $this->locker->lockCommand($this->getName());
+        $lockAcquired = $this->locker->lockCommand(__CLASS__);
```
Seldaek, line 62: The lock name should be <=64 chars, but I guess this works with all the current class names.. so probably fine. `(string)$this->getName()` would work too though.

Seldaek:
Thanks

PR #1262, opened 2022-02-20
Fixing PHPStan issues in Controller namespace

chr-hertel:
Explicit pointer to authorization logic for maintainers in `PackageController`

potentially as a next step we could even move that to a voter instead of controller, e.g. sth like `->isGranted('update', $package)`


src/Controller/ApiController.php
```diff
@@ -49,16 +49,16 @@ public function __construct(Scheduler $scheduler, LoggerInterface $logger)
...
      */
-    public function packagesAction(string $webDir)
+    public function packagesAction(string $webDir): Response
     {
         // fallback if any of the dumped files exist
         $rootJson = $webDir.'/packages_root.json';
-        if (file_exists($rootJson)) {
-            return new Response(file_get_contents($rootJson));
+        if (file_exists($rootJson) && $content = file_get_contents($rootJson)) {
+            return new Response($content);
```
staabm, line 57: Do you really need to read the whole content of the file into memory?
I would assume symfony would provide a Response class meant/optimized for this use case…?

Seldaek:
> potentially as a next step we could even move that to a voter instead of controller, e.g. sth like `->isGranted('update', $package)`

Yeah I also had that on my todos here.. adding voters to clean up some of the in-action redundancies would be nice.

PR #1264, opened 2022-03-02
GitHub Security Advisories: import available advisories via single worker run

glaubinix:
This PR builds on top of https://github.com/composer/packagist/pull/1206 to add support for GitHub advisories

Unfortunately, using GitHub webhooks to get notified about new advisories might not be an option (requires GitHub App with additional permissions), so running the job after every package update might add new advisories only a while after they were created. It would also create a huge amount of unnecessary additional jobs.

Fetching all GitHub advisories in a single run limits the number of background job we generate. It also allows us to trigger them on a regular basis via a cron (every 5/10min) with the added command

Especially, the affected version range for data from the GitHub advisory database is a lot less accurate, therefore this is always using the FOPHP data if available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,10 +47,10 @@ These steps are provided for development purposes only.
    ```
 6. Start Redis:
    ```bash
-   docker compose up -d
+   docker compose up -d # or somehow run redis on localhost without docker
    ```
 7. Run a CRON job `bin/console packagist:run-workers` to make sure packages update.
-7. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files.
+8. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files.
 
 You should now be able to access the site, create a user, etc.
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,11 @@ These steps are provided for development purposes only.
    ```bash
    symfony serve
    ```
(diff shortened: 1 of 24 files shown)


README.md
```diff
@@ -45,7 +45,11 @@ These steps are provided for development purposes only.
    ```bash
    symfony serve
    ```
-6. Run a CRON job `bin/console packagist:run-workers` to make sure packages update.
+6. Start Redis:
+   ```bash
+   docker compose up -d
+   ```
+7. Run a CRON job `bin/console packagist:run-workers` to make sure packages update.
 7. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files.
```
Seldaek, line 53:
```suggestion
8. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files.
```

README.md
```diff
@@ -45,7 +45,11 @@ These steps are provided for development purposes only.
    ```bash
    symfony serve
    ```
-6. Run a CRON job `bin/console packagist:run-workers` to make sure packages update.
+6. Start Redis:
+   ```bash
+   docker compose up -d
```
Seldaek, line 50:
```suggestion
   docker compose up -d # or somehow run redis on localhost without docker
```

Seldaek (approved):
Thanks!

Seldaek:
Thanks & thanks @Toflar as well :)

PR #1267, opened 2022-03-09
remove  #StandWithUkraine banner to updates

michealzh:
The OSS is no need the WAR!!

PR #1270, opened 2022-03-11
feat(peace): remove #StandWithUkraine

bl0cknumber:
We don't want this to be on the CLI screen. Everyone wants peace, but the writing #StandWithUkraine supports one side. This word hurts the feelings of the Russian people who don't want war. We can achieve peace if we side with both sides, regardless, and that's why there is such a thing as negotiation in the political world. This article is very annoying and makes me not want to use composer. You should apologize or delete it for writing this.

I would suggest you should write this on the Repository instead of CLI.

Let's keep the CLI neutral. 

#NoStandJustUnderstandEachOther
#NeutralComposerCLI

rel: https://github.com/composer/packagist/pull/1267#issuecomment-1063636569

### DIFF
--- a/src/Package/SymlinkDumper.php
+++ b/src/Package/SymlinkDumper.php
@@ -318,7 +318,6 @@ class SymlinkDumper
             $this->rootFile['providers-api'] = str_replace('VND/PKG', '%package%', $this->router->generate('view_providers', ['name' => 'VND/PKG', '_format' => 'json'], UrlGeneratorInterface::ABSOLUTE_URL));
             $this->rootFile['warning'] = 'Support for Composer 1 is deprecated and some packages will not be available. You should upgrade to Composer 2. See https://blog.packagist.com/deprecating-composer-1-support/';
             $this->rootFile['warning-versions'] = '<1.99';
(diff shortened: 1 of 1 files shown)


bl0cknumber:
I would suggest you should write this on the Repository instead of CLI.

bl0cknumber:
lets keep the CLI neutral.

lwlwilliam:
> We don't want this to be on the CLI screen. Everyone wants peace, but the writing #StandWithUkraine supports one side. This word hurts the feelings of the Russian people who don't want war. We can achieve peace if we side with both sides, regardless, and that's why there is such a thing as negotiation in the political world. This article is very annoying and makes me not want to use composer. You should apologize or delete it for writing this.
> 
> I would suggest you should write this on the Repository instead of CLI.
> 
> Let's keep the CLI neutral.
> 
> #NoStandJustUnderstandEachOther #NeutralComposerCLI
> 
> rel: [#1267 (comment)](https://github.com/composer/packagist/pull/1267#issuecomment-1063636569)

Friend, haven't you seen through these hypocritical people? They don't really stand with Ukraine. They just want to carve up Russia. You can see their attitude towards Palestine. They never dare to stand with Palestine.

naderman:
Who is being hurt or rather often killed right now are civilians in Ukraine. That needs to end urgently.

PR #1275, opened 2022-03-18
Fix: Resetting a non-lazy manager service is not supported.

glaubinix:
Happens when trying to reset the EM in the QueueWorker after the EM was closed

[LogicException]
In ManagerRegistry.php line 54:
 Resetting a non-lazy manager service is not supported. Try running "composer require symfony/proxy-manager-bridge".


Seldaek:
Thanks

PR #1281, opened 2022-03-24
Bootstrap Build

chr-hertel:
First step before tackling Bootstrap update:
Introducing esbuild script & moving Bootstrap to NPM

- [x] I need to test this though ... :construction_worker_man:


package.json
```diff
@@ -11,11 +12,11 @@
...
     },
     "devDependencies": {
-        "esbuild": "^0.13.15"
+        "esbuild": "^0.13.15",
+        "esbuild-plugin-sass": "^1.0.1"
     },
     "scripts": {
-        "dev": "esbuild js/app.js js/charts.js --bundle --outdir=web/build --sourcemap --loader:.gif=file --loader:.eot=file --loader:.ttf=file --loader:.svg=file --loader:.woff=file --loader:.woff2=file --target=chrome58,firefox57,safari11,edge95 --watch",
-        "build": "esbuild js/app.js js/charts.js --bundle --outdir=web/build  --minify --loader:.gif=file --loader:.eot=file --loader:.ttf=file --loader:.svg=file --loader:.woff=file --loader:.woff2=file --target=chrome58,firefox57,safari11,edge95",
-        "analyze": "esbuild js/app.js js/charts.js --bundle --outdir=web/build  --minify --loader:.gif=file --loader:.eot=file --loader:.ttf=file --loader:.svg=file --loader:.woff=file --loader:.woff2=file --target=chrome58,firefox57,safari11,edge95 --analyze"
```
Seldaek, line 19: If you can retain the analyze command somehow that'd be nice :)

chr-hertel:
This rework is not really necessary, but felt like the right way somehow :) i'm not an esbuild user though

css/app.scss
```diff
@@ -10,6 +10,9 @@
 
  */
 
+@import '../node_modules/bootstrap/dist/css/bootstrap.min.css';
+@import '../node_modules/bootstrap/dist/css/bootstrap-theme.min.css';
```
stof, line 14: be careful. When the URL explicitly ends with `.css`, Sass will generate a CSS `@import` statement instead of doing a Sass import. You probably want to omit the extension here, which would perform an actual Sass import.

esbuild.js
```diff
@@ -0,0 +1,31 @@
...
+            '.eot':'file',
+            '.ttf':'file',
+            '.svg':'file',
+            '.woff':'file',
+            '.woff2':'file',
+        },
+        target: ['chrome58', 'firefox57', 'safari11', 'edge95'],
+        plugins: [sassPlugin()],
+    })
+    .catch(() => process.exit(1));
```
stof, line 25: this should rather be handled by a `try/catch` wrapping the whole body of the async function, or by a `.catch` on the result of the async function (which would then also account for exceptions coming from the second `await`)

css/app.scss
chr-hertel: Removing the extension introduces path errors with bootstrap's font files, but cannot find an `@import` in generated css file though.

esbuild.js
chr-hertel: i hope [b51b327](https://github.com/composer/packagist/pull/1281/commits/b51b3271bd33778d7fec2a814f75a1ebd1c1fb71) is what you meant :D

stof: it is

PR #1284, opened 2022-03-29
#StandWithPeace

dschotman:


dreamoov:
Good initiative, technology should not be used to convey personal opinions.  

We have to get out of the sentimental system and think about the consequences. In France, for example, we sell weapons to Ukraine to 'support' them, I wonder if the owners of the repo also support this kind of behavior?

In any case, for everyone's respect, and from an ethical point of view, it would be nice to pass this update as soon as possible, technology should unite not divide, we are the future, it's up to us to do better than that, thx to read me, Peace

dreipoe:
Stop dragging politics into open source projects because they can be forked easily ;)

xabbuh:
Have fun maintaining this "easily".

EnderCaster:
> Stop dragging politics into open source projects because they can be forked easily ;)

To view the code, I think he just removed 'politics'

PR #1295, opened 2022-04-14
Add a note about the allowed characters for a package

darakian:
Hey all, 

I didn't see the set of allowed characters for a package mentioned anywhere on the site and after being forwarded to this regex
https://github.com/composer/packagist/blob/699c69d32f770b37b46469805e8ba9eb2ba6de6f/src/Entity/Package.php#L312

I felt that this could be helpful to document for others that end up looking. Happy to change the wording on this if need be.


Seldaek:
Thanks

PR #1300, opened 2022-04-29
Update IRC link to libera.chat

d7415:
https://getcomposer.org/doc/08-community.md implies that the channel has moved. If this is not desired, feel free to close.


Seldaek:
Yup, we forgot this :) Thanks

PR #1302, opened 2022-05-02
Readding bootstrap js

chr-hertel:
Don't ask me, I have no clue why this additional weird `jquery.js` workaround is needed

at least it fixes the issue with the navbar
Bildschirmfoto von 2022-05-02 20-19-47


Seldaek:
Ah thanks :)

PR #1304, opened 2022-05-24
test phpstan-dba feature branch

staabm:
just testing a bleeding edge phpstan dba feature


staabm:
@chr-hertel since I don't want to annoy Jordi again with my question, I send this question to you - would be great if you could spent a few minutes to answer them.

with this PR I am testing a new phpstan-dba feature, which analyzes which sql queries don't use an index.
as you can see from the 'files changed'-tab, we got a few error here.

could you have a look at these and tell me whether you would consider those a "false-positive"?
maybe we can figure out, whether/how to read a mysql explain result-set to identify queries which should be optimized by the author.

the current implemenation can be seen in https://github.com/staabm/phpstan-dba/pull/377/files#diff-325c727a8f7834a6b7b42f1b3fbfcaa5c41073e796df7eda0fa5f25a48757eb3R55-R83

stof:
@staabm looking at the reported issues, they are indeed detecting cases where queries don't use indexes (but then, I'm wondering whether it should always be reported as an error, as having indexes for everything might have drawbacks as it slows down writes, which might not be worth it if a query is used rarely in a CLI command for instance, but that's a separate topic).
The only one that might not be expected is the one reporting a missing index on a derived table (which *cannot* have an index AFAICT)

staabm:
just update phpstandba and now we no longer get errors.

it seems the few which were reported previous are related to small tables

stof:
Well, is the CI running with data in tables or with an empty DB ?

staabm:
I guess its an empty schema currently

see https://github.com/composer/packagist/blob/6a40c01ac4d1eb48236ae5242bca152c9034266b/.github/workflows/phpstan.yml#L41-L52

staabm:
> maybe should be opt-in or at least have an option to opt-out if it is too noisy.

its an opt-in feature and the min-table size arg can be [configured with the RuntimeConfiguration](https://github.com/staabm/phpstan-dba/blob/ad5da7c3d6f31ac4abfecd2fb1023e6907d32261/src/QueryReflection/RuntimeConfiguration.php#L119)

PR #1311, opened 2022-05-30
Fix advisories API documentation to reflect actual behaviour.

### LINKED ISSUE #1310: Security Advisories API requires "packages" even if "updatedSince" is included in request
## Description
[The documentation](https://packagist.org/apidoc#list-security-advisories) for the security advisories API suggests that `packages` can be omitted if `updatedSince` is passed in - but this isn't the case.
> Either a list of packages as query or request parameter or a timestamp as updatedSince query parameter need to be passed.

But if I send a request that includes `updatedSince` but omits `packages` I get an error as below. I suspect it is the documentation rather than the implementation that is incorrect - I can't think of a use case where getting _all_ updates since a given time would be useful.

## Reproduce
`curl -X POST -F 'updatedSince=1653893419' https://packagist.org/api/security-advisories/`
> {"status":"error","message":"Missing array of package names as the \u0022packages\u0022 parameter"}

GuySartorelli:
Fixes #1310

### DIFF
--- a/templates/api_doc/index.html.twig
+++ b/templates/api_doc/index.html.twig
@@ -384,7 +384,7 @@ GET https://{{ packagist_host }}/statistics.json
 
   <h4 id="list-security-advisories-by-package">{{ 'api_doc.list_security_advisory'|trans }}</h4>
 
-    <p>This endpoint provides a list of security advisories. Either a list of packages as query or request parameter or a timestamp as updatedSince query parameter need to be passed.</p>
+    <p>This endpoint provides a list of security advisories. A list of packages as query or request parameter needs to be passed. An optional timestamp as updatedSince parameter can also be passed.</p>
 
     <pre>
 GET https://{{ packagist_host }}/api/security-advisories/?updatedSince=[timestamp]&amp;packages[]=[vendor/package]


stof (approved):

templates/api_doc/index.html.twig
```diff
@@ -384,7 +384,7 @@ GET https://{{ packagist_host }}/statistics.json
 
   <h4 id="list-security-advisories-by-package">{{ 'api_doc.list_security_advisory'|trans }}</h4>
 
-    <p>This endpoint provides a list of security advisories. Either a list of packages as query or request parameter or a timestamp as updatedSince query parameter need to be passed.</p>
+    <p>This endpoint provides a list of security advisories. A list of packages as query or request parameter needs to be passed. An optional timestamp as updatedSince parameter can also be passed.</p>
```
glaubinix, line 387: This is not correct. Sending one of them is enough however, `updatedSince` is currently expected to be set as query parameter. See https://github.com/composer/packagist/blob/main/src/Controller/ApiController.php#L322

glaubinix (changes requested):

templates/api_doc/index.html.twig
GuySartorelli: Ahh, interesting. I'm not quite sure I understand why the code is making this distinction then... but in any case this PR does not address the problem, if indeed there is a problem. I'll close it.

glaubinix: Could probably just add the functionality to handle `updatedSince` sent as POST as you expected. I guess this just wasn't added yet because nobody needed it so far.

GuySartorelli: I don't need it either now that I understand how to add the var to a POST request. It's a bit confusing, but it works.

Issue #1313, opened 2022-06-04
Feature Request: Jump to search with `/`


LiamKearn:
Google now almost universally adopted this UX principal. Would be really nice for something with heavy searching like packagist.

If anyone else is interested would be appreciated happy to make a PR.

Seldaek:
Not against it if you want to PR it. But tab should work already FYI

LiamKearn:
Yeah its the first element on the page that is focusable. That should make it easy but for some reason for me I always find myself: `mod+f -> search "sign in" because it is visually before the searchbar -> tab` and it never works 😄. In reality the mental pattern I should have is `mod+r -> tab` but I feel spoilt by the recent influx of websites that support `/` to search (even github right now).

stof:
@Seldaek this should be closed as it has been implemented

PR #1314, opened 2022-06-05
Make `/` keypress focus search input

LiamKearn:
#1313.

I feel like this may break stuff. I've done 10 minutes or so of testing, But I don't know packagist very well despite how much I use the public site and having this globally means accounting for a lot of things like this stealing focus from other inputs.

As an example: without the check for inputs as active elements this would break signing in (with a password that had a `/`).

Is there any chance that there could be abnormal keyboard inputs somewhere that don't use the input tag as they should?

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -7,6 +7,26 @@ document.getElementById('search_query_query').addEventListener('keydown', functi
     }
 });
 
+// Add accessibility functionality:
+// "Press '/' to focus the searchbar".
+document.addEventListener('keydown', function (e) {
+    if (e.code && e.code !== 'Slash') {
+        return;
+    }
+    var searchInput = document.getElementById('search_query_query');
+    // Just ignore if we can't find the search input for some reason maybe we are on a page without it.
+    if (!searchInput) {
+        return;
+    }
+    // If we already have input focus ignore.
+    if (document.activeElement.tagName === 'INPUT') {
+        return;
+    }
+    searchInput.focus();
+    // Prevent '/' being inserted on focus.
+    e.preventDefault();
+});
+
 var searchParameters = {};
 
 if (decodeURI(location.search).match(/[<>]/)) {


js/search.js
```diff
@@ -7,6 +7,26 @@ document.getElementById('search_query_query').addEventListener('keydown', functi
...
+    if (e.code && e.code !== 'Slash') {
+        return;
+    }
+    var searchInput = document.getElementById('search_query_query');
+    // Just ignore if we can't find the search input for some reason maybe we are on a page without it.
+    if (!searchInput) {
+        return;
+    }
+    // If we already have input focus ignore.
+    if (document.activeElement.tagName === 'INPUT') {
```
GuySartorelli, line 22: Not sure if it's relevant but may want to exclude `input type="button"` so that buttons are treated the same whether the use an input or a button tag.

LiamKearn: Not sure I follow?

Assume you mean adding something like `&& document.activeElement.getAttribute('type') !== 'button'` so if the focus is on a button it can still be moved to the search?

That is a good idea if so, It's a shame so much of the web's javascript is obfuscated. I was trying to steal an existing implementation  but even looking at the list of keydown events registered on Github or any google site doesn't show anything for `/` despite supporting the functionality. I wonder if there is a nicer API etc for this, I don't do a lot of JS and have a sneaking feeling they've got some other API or maybe its just the fact that its a million pollyfills of abstraction away.

Faisalawanisee (approved):

Seldaek:
Thanks, seems to work well enough!

staabm:
not sure if I get it right.. but neither on firefox nightly nor on google chrome canary I get the search field focused, when pressing `/`. thats what this PR tried to implement, right?

LiamKearn:
Does the listener get triggered at all? Does your browser block the JS file or this JS from executing?

Do you get any console errors?

Try run:
```js
document.addEventListener('keydown', function (e) {
	break;
}
``` in your console, check the window attached event listeners.

I find it hard to believe this PR wouldn't work. If anything I suspected it would too well 😬

Seldaek:
Could be a keyboard layout thing too perhaps, if you need shift or smth to get a `/`? Not sure..

Seldaek:
Yes right, I somehow remember double checking event.code was fine when reviewing the PR.. but I guess I misread :D

Seldaek:
Yes that confirms it even further. I'll get a fix out later.

Seldaek:
OK fixed in dfb9671d

staabm:
fix confirmed, thx!

PR #1321, opened 2022-08-03
Add Rector and use PHP 8 attributes

IonBazan:
This PR adds [Rector](https://getrector.org/) which helps with PHP and libraries/frameworks upgrades.
Because the diff would be too large for manual review, I left executing the upgrade to the maintainer.

Changes include:
- Adding `rector/rector` dependency and upgrading `phpstan/phpstan`
- Adding Symfony 5.4 upgrade list - uses `#[Route]` and `#[Assert\*]` attributes instead of annotations
- Adding Doctrine ORM 2.9 list - uses PHP 8 attributes instead of annotations for entity mapping (a change is needed in `config/packages/doctrine.yaml`)
- Adding PHPUnit 9.x list - uses `$this->createMock()` where applicable
- Converting `App\Validator\*` constraints to PHP 8 attributes

Unfortunately, adding PHP 8 upgrade rules list causes Rector to crash on my machine so I didn't add it this time.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -89,6 +89,7 @@
         "phpstan/phpstan-doctrine": "^1",
         "phpstan/phpstan-symfony": "^1.1",
         "phpunit/phpunit": "^9.5",
+        "rector/rector": "^0.13.9",
         "staabm/phpstan-dba": "^0.2",
         "symfony/browser-kit": "^5.2",
         "symfony/css-selector": "^5.2",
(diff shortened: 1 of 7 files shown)


Seldaek:
Closing as https://github.com/composer/packagist/pull/1329 includes the commit here. Thanks for just sending the config and not doing all the changes :)

PR #1324, opened 2022-08-14
Add Gitea webhook for Packagist

toddy15:
Hi,

Gitea recently added support for a Packagist webhook, enabling automatic updates for packages. I've added a section about how to enable it.

Regards,
Tobias

### DIFF
--- a/templates/about/about.html.twig
+++ b/templates/about/about.html.twig
@@ -90,8 +90,8 @@ v2.0.4-p1</code></pre>
         <section class="col-md-6">
             <h3 id="update-schedule">Update Schedule</h3>
             <p>New packages will be crawled <strong>immediately</strong> after submission if you have JS enabled.</p>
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks - I do wonder why they need the packagist.org package URL tho - as what we need to receive in the payload is the canonical repository URL (the gitea one)

toddy15:
@Seldaek Thanks for merging -- I thought the payload needs the packagist URL? At least that's how I understood the section about the manual hook setup.

I did create the PR with support for packagist in Gitea, so I'm wondering if it could be simplified. :-)

Seldaek:
Ah yeah sorry I mis-remembered this :) I believe both URLs are supported, either canonical git repo URL or packagist one, but best would be to try it out to ensure that the response you get is positive (HTTP 202) if you use the gitea repo URL. That'd make the setup slightly easier, and has the benefit of updating multiple packages at once if the same git repo gets submitted with different names over time.

toddy15:
@Seldaek Just FYI: I've tested with the canonical "upstream" git URL and with the packagist URL. Both work fine, so I guess the implementation in Gitea could be simplified. Thanks for your input.

PR #1326, opened 2022-08-31
Symfony 6.0 ugprade

Seldaek:


.env
```diff
@@ -50,3 +45,12 @@ APP_RECAPTCHA_ENABLED=false
...
 APP_RECAPTCHA_SECRET=needed
 ###< beelab/recaptcha2-bundle ###
+
+###> doctrine/doctrine-bundle ###
+# Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
+# IMPORTANT: You MUST configure your server version, either here or in config/packages/doctrine.yaml
+#
+# DATABASE_URL="sqlite:///%kernel.project_dir%/var/data.db"
+# DATABASE_URL="mysql://app:!ChangeMe!@127.0.0.1:3306/app?serverVersion=8&charset=utf8mb4"
+DATABASE_URL="postgresql://app:!ChangeMe!@127.0.0.1:5432/app?serverVersion=14&charset=utf8"
```
stof, line 58: This should uncomment the mysql one instead, as packagist uses mysql

src/DataFixtures/UserFixtures.php
```diff
@@ -22,18 +22,27 @@ public function __construct(private UserPasswordHasherInterface $passwordHasher)
...
+        $dev->initializeApiToken();
 
-        $user->setEmail('dev@packagist.org');
-        $user->setUsername('dev');
-        $user->setPassword($this->passwordHasher->hashPassword($user, 'dev'));
-        $user->setEnabled(true);
+        $manager->persist($dev);
 
+        $user = new User;
+        $user->setEmail('dev@example.org');
```
stof, line 36: is it actually supported to have multiple users with the same email ?

stof:
Why 6.0 and not 6.1 ?

Seldaek:
Thanks, fixed those. Why 6.0? Because I rather do it one by one, but I'll get there :)

PR #1330, opened 2022-09-12
Remove obsolete Polyfill packages and configuration

IonBazan:
Removes unused `symfony/polyfill*` packages and `incenteev-parameters` configuration.


composer.json
```diff
@@ -127,6 +127,11 @@
...
         "symfony/polyfill-ctype": "*",
         "symfony/polyfill-iconv": "*",
+        "symfony/polyfill-php81": "*",
+        "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php74": "*",
+        "symfony/polyfill-php73": "*",
+        "symfony/polyfill-php72": "*",
         "symfony/polyfill-php71": "*",
         "symfony/polyfill-php70": "*",
         "symfony/polyfill-php56": "*"
```
localheinz, line 137: What about these?

IonBazan: These are in `replace` section, so they will not be installed even if other packages require them.

localheinz: Ha, my bad!

👍

localheinz (approved):
👍

Seldaek:
Thanks

PR #1332, opened 2022-09-15
Convert remaining annotations to attributes

IonBazan:
This PR converts some remaining `@Route` and `@IsGranted` annotations to attributes and uses `#[AsEventListener]` and `#[AsEntityListener]` attributes to register EventListeners and Entity listeners, reducing the amount of configuration in `services.yaml`.


Seldaek:
Thanks!

PR #1333, opened 2022-09-15
Simplify services.yaml

IonBazan:
Another round of simplifying the `config/services.yaml` - thanks to the Autowiring and binding the arguments in the `_defaults` section, we are able to omit some configuration.


config/services.yaml
```diff
@@ -108,13 +98,10 @@ services:
         factory: ['Algolia\AlgoliaSearch\SearchClient', create]
 
     App\Service\QueueWorker:
-        arguments:
```
Seldaek, line 111: Ah nice, I never noticed you could skip arguments/bind :)

Seldaek:
Thanks

PR #1334, opened 2022-09-15
Get started on audit log functionality

Seldaek:
Lots to do still in AuditRecordType - but this lays the foundation down already so would be good to merge as first step.

### DIFF
--- a/src/Entity/AuditRecord.php
+++ b/src/Entity/AuditRecord.php
@@ -24,24 +24,24 @@ class AuditRecord
     #[ORM\Column(type: 'ulid')]
     public readonly Ulid $id;
 
-    #[ORM\Column(type: Types::DATETIME_IMMUTABLE)]
+    #[ORM\Column]
     public readonly DateTimeImmutable $datetime;
 
     private function __construct(
-        #[ORM\Column(length: 255, enumType: AuditRecordType::class)]
+        #[ORM\Column]
         public readonly AuditRecordType $type,
 
         /** @var array<string, mixed> */
-        #[ORM\Column]
+        #[ORM\Column(type: Types::JSON)]
         public readonly array $attributes,
 
-        #[ORM\Column(type: Types::INTEGER, nullable: true)]
+        #[ORM\Column(nullable: true)]
         public readonly int|null $userId = null,
 
         #[ORM\Column(nullable: true)]
         public readonly string|null $vendor = null,
 
-        #[ORM\Column(type: Types::INTEGER, nullable: true)]
+        #[ORM\Column(nullable: true)]
         public readonly int|null $packageId = null,
     ) {
         $this->id = new Ulid();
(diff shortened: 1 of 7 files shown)


src/Entity/AuditRecord.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    #[ORM\Id]
+    #[ORM\Column(type: 'ulid')]
+    public readonly Ulid $id;
+
+    #[ORM\Column(type: Types::DATETIME_IMMUTABLE)]
+    public readonly DateTimeImmutable $datetime;
+
+    private function __construct(
+        #[ORM\Column(length: 255, enumType: AuditRecordType::class)]
```
stof, line 31: I'm quite sure `enumType` is automatically guessed from the property type already.

src/Entity/AuditRecord.php
```diff
@@ -0,0 +1,77 @@
...
+        public readonly AuditRecordType $type,
+
+        /** @var array<string, mixed> */
+        #[ORM\Column]
```
stof, line 35: Which column type is used here ? It is stored as serialized PHP or as JSON ?

I suggest adding the type explicitly here instead of relying on the guessed type for `array`

src/Entity/AuditRecord.php
```diff
@@ -0,0 +1,77 @@
...
+        public readonly array $attributes,
+
+        #[ORM\Column(type: Types::INTEGER, nullable: true)]
+        public readonly int|null $userId = null,
+
+        #[ORM\Column(nullable: true)]
+        public readonly string|null $vendor = null,
+
+        #[ORM\Column(type: Types::INTEGER, nullable: true)]
+        public readonly int|null $packageId = null,
```
stof, line 45: shouldn't we have a package name too ?

src/EventListener/PackageListener.php
```diff
@@ -0,0 +1,96 @@
...
+        // let the record be flushed together with the entity
+    }
+
+    /**
+     * @param PreUpdateEventArgs $event
+     */
+    public function preUpdate(Package $package, PreUpdateEventArgs $event): void
+    {
+        if ($event->hasChangedField('repository')) {
+            // buffering things to be flushed in postUpdate as flushing here results in an infinite loop
```
stof, line 71: couldn't you persist them here and let them be flushed together with the entity ?

src/Entity/AuditRecord.php
Seldaek: It's in the attributes, the id is mostly there for SQL querying/filtering ability.

src/EventListener/PackageListener.php
Seldaek: I thought so too but it didn't do anything..

src/Entity/AuditRecord.php
Seldaek: It seems to have defaulted (correctly) to a json column, so all good, but I will put it explicitly just in case.

PR #1338, opened 2022-10-05
Add #WomanLifeFreedom banner to updates 🟩⬜🟥

alirezamirsepassi:
https://en.wikipedia.org/wiki/Woman,_Life,_Freedom


Seldaek:
I fully sympathize with your situation, but unlike in the Russian war on Ukraine, I suspect a message in Composer would hardly reach anyone of interest in Iran. Sorry this will disappoint some of you.

PR #1339, opened 2022-10-12
Add Denmark mirror

KristianI:
Added mirror located in Denmark.

Benchmark shows around 10 seconds of saved time over 20 installs from my location (Fyn, Denmark).

image (5)


Seldaek:
IMO using a third party mirror that offers only minor latency improvement isn't worth the security risk, but it's good to have in case people experience connectivity issues to our network.

PR #1344, opened 2022-12-08
[Chore][CI] Improve CI workflow

IonBazan:
- Makes use of awesome https://github.com/ramsey/composer-install action, which takes care of the caching, thus removing usage of deprecated `::set-output`.
- Bumping `actions/checkout` to `v3` (v2 uses deprecated Node v12)
- Setting `DATABASE_URL` via physical `env` for clarity


Seldaek:
Thanks

PR #1350, opened 2023-01-06
phpstan-dba: enable analysis of writable queries

staabm:
this setting can only be enabled when we are sure phpstan-dba will only ran against a db schema/server supporting transactions (e.g. all tables use InnoDB etc).


staabm:
with the latest dba release writable queries are enabled per default and we no longer need transactions for it

PR #1351, opened 2023-01-11
Don't share private cache via CDN.

vtsykun:
Hi. I noticed that the cache is saved html, which contains the data of the authorized user.  The data does not contain security sensitive information, but you can see the nickname of the last person who viewed this page 
It affected only `/php-statistics` page. The other pages with public cache control `NO_AUTO_CACHE_CONTROL_HEADER` is JSON not affected.

**STR**
1) Login to packagist
2) Open https://packagist.org/php-statistics
3) Open this page in incognito 

**Actual result** 
Depends on who last viewed this page when the cache was builded:

a) One the step 3. You will see the nickname of the last visited user
b) Or on step 2 you will see menu for non-authorized user
Selection_1161

### DIFF
--- a/src/Controller/WebController.php
+++ b/src/Controller/WebController.php
@@ -20,6 +20,7 @@ use App\Search\Algolia;
 use App\Search\Query;
 use App\Util\Killswitch;
 use Predis\Connection\ConnectionException;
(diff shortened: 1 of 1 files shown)


Seldaek:
Ah good point, thanks!

PR #1353, opened 2023-01-14
Add link to package security policy

ramsey:
This is a complementary feature for the security support option added in https://github.com/composer/composer/pull/11271

If the package has a security support link provided in `composer.json`, this will add a "Security Policy" link on Packagist. For example:

Screenshot 2023-01-14 at 13 55 22


Seldaek:
Thanks

PR #1355, opened 2023-01-18
Update deps, recipes and fix phpstan build

Seldaek:


phpstan-baseline.neon
```diff
@@ -45,6 +55,16 @@ parameters:
 			count: 1
 			path: src/Model/FavoriteManager.php
 
+		-
+			message: "#^Call to an undefined method Symfony\\\\Component\\\\HttpFoundation\\\\Session\\\\SessionInterface\\:\\:getFlashBag\\(\\)\\.$#"
+			count: 2
+			path: src/Security/GitHubAuthenticator.php
```
Seldaek, line 61: If anyone has a clue how to fix this I'd be interested.. There's no $this->addFlash available in services like https://github.com/composer/packagist/blob/main/src/Security/GitHubAuthenticator.php#L147 and the symfony interfaces seem to be too minimalistic.

stof: There is a `FlashBagAwareSessionInterface` that contains this method. So this could be solved by adding `assert($session instanceof FlashBagAwareSessionInterface)` (which will always be the case in Symfony if you don't decorate the Session object as the core class implements that interface)

Seldaek: OK. Not great having to add asserts but I guess better than baseline..

PR #1362, opened 2023-02-05
Update to PHPUnit 10.0

chr-hertel:
Just gave it a shot and was surprised.

Added some return type declarations while in tests folder.

Btw, not sure this test suite uses anything from Symfony's phpunit-bridge. Would even propose to remove that dependency.
Deprecation notices are handled nicer in PHPUnit since 10.0 as well

WDYT?


phpunit.xml.dist
```diff
@@ -3,18 +3,16 @@
...
-         convertDeprecationsToExceptions="false"
 >
     <php>
         <ini name="display_errors" value="1" />
         <ini name="error_reporting" value="-1" />
         <server name="APP_ENV" value="test" force="true" />
         <server name="SHELL_VERBOSITY" value="-1" />
         <server name="SYMFONY_PHPUNIT_REMOVE" value="" />
-        <server name="SYMFONY_PHPUNIT_VERSION" value="9.5" />
+        <server name="SYMFONY_PHPUNIT_VERSION" value="10.0" />
```
chr-hertel, line 15: is ignored since phpunit/phpunit is directly installed anyways - just changed it for .. donno, looks better

tests/Security/BruteForceLoginFormAuthenticatorTest.php
```diff
@@ -28,15 +30,15 @@
 class BruteForceLoginFormAuthenticatorTest extends TestCase
 {
     private BruteForceLoginFormAuthenticator $authenticator;
-    /** @var HttpUtils&\PHPUnit\Framework\MockObject\MockObject */
+    /** @var HttpUtils&MockObject */
     private $httpUtils;
```
Seldaek, line 34: Could use this I believe?

```suggestion
    private HttpUtils&MockObject $httpUtils;
```

Seldaek:
Yes please get rid of phpunit-bridge, I don't know why it's there probably **historical reasons**

IonBazan (approved):
Amazing job! 👍🏻

Seldaek:
Nice, thanks :)

PR #1375, opened 2023-05-04
More protection from account takeovers

vtsykun:
Additional fixes for #1374. The user can use an api token as GitHub secret and a custom webhook payload to change the URL of a package with more than 50k downloads.

### DIFF
--- a/src/Controller/ApiController.php
+++ b/src/Controller/ApiController.php
@@ -512,7 +512,7 @@ class ApiController extends Controller
         $updateUrl = false;
 
         // maybe url changed, look up by remoteId
-        if (!$packages) {
+        if (!$packages && $user === null) {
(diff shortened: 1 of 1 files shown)


Seldaek:
This isn't really fixing the issue tho is it? There still need to be a proper permission check here, but thanks for alerting me of this, I'll take care of it.

Seldaek:
Ah nevermind I got what this fixes now by looking into it a bit closer. Thanks!

Seldaek:
Ah sorry looking again a bit further.. Reverted this in the end and fixed it by validating the remoteId, because we do want to update URLs automatically if a repo was moved on github, but we should ensure that github data agrees with what is in the request. See https://github.com/composer/packagist/commit/b420816e0cc73bbe616f22771b7b0c3d058bb5a9

PR #1376, opened 2023-05-12
Webauthn login support

Spomky:
This PR aims at adding Webauthn login.
Webauthn is a [web standard](https://www.w3.org/TR/webauthn-2/) that allows the use of strong public key-based credentials for user authentication.
It is proposed in reaction of the [recent issue](https://thehackernews.com/2023/05/packagist-repository-hacked-over-dozen.html) where authentication failure is a key point for such attacks.

* [ ] Allow users to manage the authenticators
    * [ ] Add a new one
    * [ ] Remove an existing one
    * [ ] Custom name
* [ ] Change login popup
* [ ] Change login page

### DIFF
--- a/.env
+++ b/.env
@@ -60,3 +60,8 @@ DATABASE_URL="mysql://root@127.0.0.1:3306/packagist?serverVersion=8&charset=utf8
 APP_ENV=dev
 APP_SECRET=f7bd3db39aea6b0c35d6703fab361df5
 ###< symfony/framework-bundle ###
+
+###> web-auth/webauthn-symfony-bundle ###
+RELAYING_PARTY_ID=packagist.org
+RELAYING_PARTY_NAME="Packagist"
+###< web-auth/webauthn-symfony-bundle ###
--- a/config/bundles.php
+++ b/config/bundles.php
@@ -20,4 +20,6 @@ return [
     SymfonyCasts\Bundle\VerifyEmail\SymfonyCastsVerifyEmailBundle::class => ['all' => true],
     KnpU\OAuth2ClientBundle\KnpUOAuth2ClientBundle::class => ['all' => true],
     Scheb\TwoFactorBundle\SchebTwoFactorBundle::class => ['all' => true],
+    SpomkyLabs\CborBundle\SpomkyLabsCborBundle::class => ['all' => true],
+    Webauthn\Bundle\WebauthnBundle::class => ['all' => true],
 ];
--- a/src/Entity/WebauthnCredential.php
+++ b/src/Entity/WebauthnCredential.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Entity;
+
+use Doctrine\DBAL\Types\Types;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\Table;
+use Symfony\Component\Uid\AbstractUid;
+use Symfony\Component\Uid\Ulid;
+use Webauthn\PublicKeyCredentialSource as BasePublicKeyCredentialSource;
+use Webauthn\TrustPath\TrustPath;
+
+#[Table(name: "webauthn_credentials")]
+#[Entity(repositoryClass: WebauthnCredentialRepository::class)]
+class WebauthnCredential extends BasePublicKeyCredentialSource
+{
+    #[Id]
+    #[Column(type: Types::STRING, unique: true)]
+    #[GeneratedValue(strategy: "NONE")]
+    private string $id;
+
+    public function __construct(
+        string $publicKeyCredentialId,
+        string $type,
+        array $transports,
+        string $attestationType,
+        TrustPath $trustPath,
+        AbstractUid $aaguid,
+        string $credentialPublicKey,
+        string $userHandle,
+        int $counter
+    )
+    {
+        $this->id = Ulid::generate();
+        parent::__construct($publicKeyCredentialId, $type, $transports, $attestationType, $trustPath, $aaguid, $credentialPublicKey, $userHandle, $counter);
+    }
+
+    public function getId(): string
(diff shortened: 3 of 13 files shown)


.env
```diff
@@ -60,3 +60,8 @@ DATABASE_URL="mysql://root@127.0.0.1:3306/packagist?serverVersion=8&charset=utf8
 APP_ENV=dev
 APP_SECRET=f7bd3db39aea6b0c35d6703fab361df5
 ###< symfony/framework-bundle ###
+
+###> web-auth/webauthn-symfony-bundle ###
+RELAYING_PARTY_ID=packagist.org
```
stof, line 65: This should be `Relying`, not `Relaying` (if this comes from a recipe, the naming might need to be fixed in it)

src/Entity/WebauthnCredential.php
```diff
@@ -0,0 +1,47 @@
...
+use Symfony\Component\Uid\Ulid;
+use Webauthn\PublicKeyCredentialSource as BasePublicKeyCredentialSource;
+use Webauthn\TrustPath\TrustPath;
+
+#[Table(name: "webauthn_credentials")]
+#[Entity(repositoryClass: WebauthnCredentialRepository::class)]
+class WebauthnCredential extends BasePublicKeyCredentialSource
+{
+    #[Id]
+    #[Column(type: Types::STRING, unique: true)]
```
stof, line 23: This `unique: true` is useless. A primary key is already a unique field.

src/Entity/WebauthnCredentialRepository.php
```diff
@@ -0,0 +1,35 @@
...
+                $publicKeyCredentialSource->getTransports(),
+                $publicKeyCredentialSource->getAttestationType(),
+                $publicKeyCredentialSource->getTrustPath(),
+                $publicKeyCredentialSource->getAaguid(),
+                $publicKeyCredentialSource->getCredentialPublicKey(),
+                $publicKeyCredentialSource->getUserHandle(),
+                $publicKeyCredentialSource->getCounter()
+            );
+        }
+        parent::saveCredentialSource($publicKeyCredentialSource);
```
stof, line 33: shouldn't this forward the `$flush` parameter ? Otherwise, it is useless

.env
Spomky: 🫢https://github.com/symfony/recipes-contrib/pull/1513

src/Entity/WebauthnCredential.php
Spomky: Correct. I removed it

src/Entity/WebauthnCredentialRepository.php
Spomky: The [parent](https://github.com/web-auth/webauthn-framework/blob/4.6.x/src/symfony/src/Repository/DoctrineCredentialSourceRepository.php#L35) already calls the flush method

stof: Ah, the parent does not have this argument at all. So why adding it in the child class without using it ?

Spomky:
Hi @stof,

Many thanks for the first comments.
I pushed the last modifications and it is now working on my platform.
During the coming days, I will improve the way the authenticators are managed.
Note that the UX is far from being perfect. I need help for it to make sure it is in line with your requirements.

Spomky:
Hello,

I have two questions 

1. When I run `symfony console doctrine:fixtures:load`, the following exception is thrown:
    > purging database
   > loading App\DataFixtures\UserFixtures
   > loading App\DataFixtures\PackageFixtures
  0/100 [>---------------------------]   0% (< 1 sec left) https://github.com/php-fig/log18:18:10 CRITICAL  [console] Error thrown while running command "doctrine:fixtures:load". Message: "Invalid fallback user was not found" ["exception" => LogicException { …},"command" => "doctrine:fixtures:load","message" => "Invalid fallback user was not found"]

Is there anything I missed (env var or something else)?

2. How to add an entity field? I often use doctrine migration, but it seems it is not used here.
What is the process for adding the following field to the user entity?

```php
    #[ORM\Column(type: 'string', name: 'user_handle', length: 200, nullable: true, unique: true)]
    private ?string $userHandle = null;
```

Many thanks.
Regard.

config/bundles.php
```diff
@@ -20,4 +20,6 @@
     SymfonyCasts\Bundle\VerifyEmail\SymfonyCastsVerifyEmailBundle::class => ['all' => true],
     KnpU\OAuth2ClientBundle\KnpUOAuth2ClientBundle::class => ['all' => true],
     Scheb\TwoFactorBundle\SchebTwoFactorBundle::class => ['all' => true],
+    SpomkyLabs\CborBundle\SpomkyLabsCborBundle::class => ['all' => true],
+    Webauthn\Bundle\WebauthnBundle::class => ['all' => true],
```
94noni, line 24: i see that you need on twig files `templates/user/login.html.twig` & `templates/user/manage_authenticators.html.twig` some javascript, than seems to be kind of generic no?
maybe the bundle can provide them?
a raw js file (or perhaps a stimulus controller)

package.json
```diff
@@ -1,9 +1,9 @@
 {
     "name": "packagist.org",
     "dependencies": {
-        "bootstrap": "3.3.5",
+        "bootstrap": "^3.4.1",
         "d3": "^3.5.17",
-        "instantsearch.js": "^2.7.4",
+        "instantsearch.js": "^4.56.0",
```
94noni, line 6: is this required with your PR?

src/Controller/ManageAuthenticatorsController.php
```diff
@@ -0,0 +1,44 @@
...
+        private readonly WebauthnCredentialRepository $credentialRepository
+    )
+    {
+    }
+
+    #[Route(path: '/profile/webauthn', name: 'manage_authenticators')]
+    public function manage(): Response
+    {
+        $user = $this->getUser();
+        !$user instanceof User || $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
```
94noni, line 33: what about `#[CurrentUser] User $user` in the `manage()` ?

94noni (commented):
some reviews passing by :)

PR #1384, opened 2023-07-03
removed: `php.cnpkg.org`, reason: dead mirror

xiaohuilam:
Since [Jul 30, 2021](http://web.archive.org/web/20210601000000*/https://php.cnpkg.org/), php.cnpkg.org is offline. So please remove from mirror recommendation to avoid consumption of community time.


Seldaek:
Thanks

Issue #1393, opened 2023-09-07
Sudden 404 not found on API for some existing packages


kylekatarnls:
Hello, 👋 

Until recently, I was able to get info from API about `pug/pug` using https://repo.packagist.org/p/pug/pug.json

As you can see the package exists for long:
https://packagist.org/packages/pug/pug

And the web page properly shows it, only the API say it's missing, and some other similar packages still work:
https://repo.packagist.org/p/pug/slim.json

I get: `"404 not found, no packages here"` both locally and via GitHub Actions.

Thanks 🙏

stof:
the `/p/` endpoints are the endpoints for the Composer v1 metadata format, which is deprecated and not guaranteed to provide all packages anymore. See https://blog.packagist.com/deprecating-composer-1-support/

You should use the endpoints of the v2 format.

See https://packagist.org/apidoc#get-package-data

kylekatarnls:
Thank you, I was not aware and so moving to `/p2/` will be just fine for me. However I think mentioning in the doc is of little help for discoverability.

And it sounds like not being sure either the data are correct or not in `/p/` actually make it useless (I can't  see a use-case where unreliable API result would be fine).

Considering this, wouldn't it make more sense to shutdown completely at once and either redirect to `/p2/` or send some generic 5xx error.

stof:
Redirecting is a no-go as the structure of the data is different (that's why the v2 metadata is provided in new endpoints).

And dropping the `/p/` endpoints entirely means dropping support entirely for Composer 1.x, so this needs to be based on the usage of Composer 1.x.

vtsykun:
More simple to use this endpoint if you call API directly 

```
https://packagist.org/packages/pug/pug.json
```

stof:
Well, depending on what you want to do, this can indeed provide more infos. But it is much more costly in term of resources for the server than the metadata endpoints (which are served from static files dumped on the filesystem). So if you call the API a lot for metadata that appear in the composer repository metadata, it is much better to use the metadata endpoints.

Seldaek:
Yeah please don't use the `/packages/` API unless you need data that is only available in it. Sure it's a bit more complex to use the `/p2/` API as you need to use https://github.com/composer/metadata-minifier to decompress the data, but that's not *that* hard it's one extra line of code.

PR #1394, opened 2023-09-08
Mention the partial data for the v1 format in API docs

stof:
This will help devs to understand the issue (assuming they read the documentation). See https://github.com/composer/packagist/issues/1393


Seldaek:
Yeah.. the big "DEPRECATED" is not clearly enough :D Thanks

stof:
Well, the big deprecated does not explain that the available info is partial due to the deprecation. The blog post explains that.

Issue #1392, opened 2023-09-01
Security Advisories page should list all associated identifiers


naderman:
A security advisories page, e.g. https://packagist.org/packages/symfony/http-foundation/advisories?version=2557737 should show all the related identifiers. That includes in particular our own PKSA- identifiers so one can easily copy and reference them.

We should also provide an easy way to look up an advisory through a corresponding PKSA id in the URL.

Seldaek:
Ok the GHSA id and PKSA are shown now on https://packagist.org/packages/symfony/http-foundation/advisories?version=2557737 for example, if you click the PKSA one you get to a page referencing just that one advisory. And if you enter PKSA, GHSA or CVE id in the search it also just redirects you to that page. e.g. https://packagist.org/security-advisories/CVE-2019-18888

PR #1395, opened 2023-09-13
Security advisories: store severity

glaubinix:
This fetches the severity for each advisory from the GitHub security advisory database and make the data available for Composer to use during `composer audit`.

The severity is currently returned/shown as part of `/api/security-advisories/` (I believe this is enough for Composer?) and the UI.

I am not sure if storing the severity as part of the `SecurityAdvisorySource` is necessary. I figured if we have multiple sources reporting the severity, we can calculate a median. Happy to remove this if people think this isn't necessary.

UI Example:
Screenshot 2023-10-27 at 09 57 16

### DIFF
--- a/src/Entity/SecurityAdvisory.php
+++ b/src/Entity/SecurityAdvisory.php
@@ -138,6 +138,10 @@ class SecurityAdvisory
         if ($initialCopy || $this->reportedAt != $advisory->date) {
             $this->reportedAt = $advisory->date;
         }
+
+        if ($initialCopy && $advisory->severity) {
+            $this->severity = $advisory->severity;
+        }
     }
 
     public function getPackagistAdvisoryId(): string
--- a/src/Entity/SecurityAdvisory.php
+++ b/src/Entity/SecurityAdvisory.php
@@ -15,6 +15,7 @@ namespace App\Entity;
 use App\SecurityAdvisory\AdvisoryIdGenerator;
 use App\SecurityAdvisory\AdvisoryParser;
 use App\SecurityAdvisory\FriendsOfPhpSecurityAdvisoriesSource;
+use App\SecurityAdvisory\Severity;
 use DateTimeImmutable;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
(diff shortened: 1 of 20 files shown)


src/Entity/SecurityAdvisory.php
```diff
@@ -86,44 +90,53 @@ public function __construct(RemoteSecurityAdvisory $advisory, string $source)
...
-        if (!in_array($advisory->getSource(), [null, $this->source], true)) {
+        $this->findSecurityAdvisorySource($advisory->source)?->update($advisory);
+
+        $now = new DateTimeImmutable();
+        if (!$this->severity && $advisory->severity) {
+            $this->updatedAt = $now;
+            $this->severity = $advisory->severity;
+        }
+
+        if (!in_array($advisory->source, [null, $this->source], true)) {
```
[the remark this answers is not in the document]
glaubinix: In general, we trust the FriendsOfPHP source over GitHub if available but unfortunately that source doesn't contain any severity information. Therefore, we need to always store the severity from GitHub if available.

I think until we have more sources, we can probably just hard code that severity always comes from GitHub.

src/Entity/SecurityAdvisory.php
```diff
@@ -70,6 +71,9 @@ class SecurityAdvisory
     #[ORM\Column(type: 'string', nullable: true)]
     private string|null $composerRepository = null;
 
+    #[ORM\Column(nullable: true)]
+    private Severity|null $severity = null;
```
vtsykun, line 75: It may make sense to use a numeric score value instead of enum. This will make it easier to compare or sort.

Seldaek (approved):

templates/package/security_advisories.html.twig
```diff
@@ -32,6 +32,7 @@
                                 <div class="row">
                                     <div class="col-sm-8 col-lg-9">
                                         <h4 class="font-bold">
+                                            {% if advisory.severity %}[{{ advisory.severity|upper }}]{% endif %}
```
Seldaek, line 35: Can you please rebase due to https://github.com/composer/packagist/issues/1392#issuecomment-1782453560 and make sure to add the severity display as well somehow to the new security_advisory.html.twig template

glaubinix: rebased and extracted the security advisory list into a shared template

Seldaek:
Thanks

Issue #1293, opened 2022-04-12
2FA requirement when using "password forgotten"


erkens:
I needed to use the "forgot password" feature on a company account because the original maintainer wasn't working for this company, the reset link was mailed and I could change the password.
However after login I got the 2FA request and needed to contact the original maintainer of this account to disable the 2FA. He stated that the password was indeed changed and I had to provide him the original password.

Change requested: Ask for 2FA before storing the new password if 2FA is enabled.

PR #1403, opened 2023-11-17
On password reset require 2FA code before storing new password

glaubinix:
Resolves https://github.com/composer/packagist/issues/1293

- Extracted 2FA code validation into form constraint + validator
- Added 2FA code stub to make 2FA testing easier
- Tests for 2FA enable + password reset
- Reset password now requests a 2FA code before password is reset

### DIFF
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -265,7 +265,7 @@ class UserController extends Controller
             throw $this->createAccessDeniedException('You cannot change this user\'s two-factor authentication settings');
         }
 
-        $secret = (string) $req->getSession()->get('2fa_secret', $authenticator->generateSecret());
+        $secret = (string) $req->getSession()->get('2fa_secret') ?: $authenticator->generateSecret();
         // Temporarily store this code on the user, as we'll need it there to generate the
         // QR code and to check the confirmation code.  We won't actually save this change
         // until we've confirmed the code
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -265,39 +265,28 @@ class UserController extends Controller
             throw $this->createAccessDeniedException('You cannot change this user\'s two-factor authentication settings');
         }
 
-        $enableRequest = new EnableTwoFactorRequest();
-        $form = $this->createForm(EnableTwoFactorAuthType::class, $enableRequest)
-            ->handleRequest($req);
-
-        $secret = (string) $req->getSession()->get('2fa_secret');
-        if (!$form->isSubmitted() || '' === $secret) {
-            $secret = $authenticator->generateSecret();
-            $req->getSession()->set('2fa_secret', $secret);
-        }
-
+        $secret = (string) $req->getSession()->get('2fa_secret', $authenticator->generateSecret());
         // Temporarily store this code on the user, as we'll need it there to generate the
         // QR code and to check the confirmation code.  We won't actually save this change
         // until we've confirmed the code
         $user->setTotpSecret($secret);
 
-        if ($form->isSubmitted()) {
-            // Validate the code using the secret that was submitted in the form
-            if (!$authenticator->checkCode($user, $enableRequest->getCode() ?? '')) {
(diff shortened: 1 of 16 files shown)


src/Controller/UserController.php
```diff
@@ -265,39 +265,28 @@ public function enableTwoFactorAuthAction(Request $req, #[VarName('name')] User
...
-        $form = $this->createForm(EnableTwoFactorAuthType::class, $enableRequest)
-            ->handleRequest($req);
-
-        $secret = (string) $req->getSession()->get('2fa_secret');
-        if (!$form->isSubmitted() || '' === $secret) {
-            $secret = $authenticator->generateSecret();
-            $req->getSession()->set('2fa_secret', $secret);
-        }
-
+        $secret = (string) $req->getSession()->get('2fa_secret', $authenticator->generateSecret());
```
stof, line 268: this is now generating a new secret (that will be ignored) even when the session already has one. If the generation takes time, it might be a bad idea.

I suggest doing `$secret = (string) $req->getSession()->get('2fa_secret') ?? $authenticator->generateSecret();` instead

src/Form/Type/EnableTwoFactorAuthType.php
```diff
@@ -25,13 +27,18 @@ class EnableTwoFactorAuthType extends AbstractType
...
 
     public function configureOptions(OptionsResolver $resolver): void
     {
-        $resolver->setDefaults([
-            'data_class' => EnableTwoFactorRequest::class,
-        ]);
+        $resolver
+            ->setDefaults([
+                'data_class' => EnableTwoFactorRequest::class,
+                'user' => null,
```
stof, line 40: adding a `null` default is broken, as `null` is not a valid value (if you don't provide a user, `new TwoFactorCodeConstraint($options['user'])` will produce a TypeError with the null default).
This should define a required option instead of an option with a default value.

src/EventListener/ResolvedTwoFactorCodeCredentialsListener.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace App\EventListener;
+
+use App\Security\Passport\Badge\ResolvedTwoFactorCodeCredentials;
+use Scheb\TwoFactorBundle\Security\Http\Authenticator\TwoFactorAuthenticator;
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+use Symfony\Component\Security\Http\Event\AuthenticationTokenCreatedEvent;
+
+#[AsEventListener(event: AuthenticationTokenCreatedEvent::class, method: 'onAuthenticationTokenCreated', priority: 512)]
```
stof, line 20: to avoid hardcoding the method name in a string, you could put the attribute on the method instead.

src/Form/ResetPasswordFormType.php
```diff
@@ -12,13 +12,24 @@
...
+use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 class ResetPasswordFormType extends AbstractType
 {
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setDefault('user', null)
```
stof, line 29: `user` should be a required option here too

glaubinix:
Thanks!

src/Form/Validation/TwoFactorCodeConstraint.php
```diff
@@ -0,0 +1,26 @@
...
+ * This file is part of Packagist.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *     Nils Adermann <naderman@naderman.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Form\Validation;
```
stof, line 13: To follow the usual Symfony convention, the `Validation` should not be nested in the `Form` one

Seldaek:
Looks good thanks!

Issue #1368, opened 2023-03-28
Add a package maintainability score


vtsykun:
When searching for a package it would be nice to see a rating/score index of how the package is actively maintained. For example, this value can be calculated based on the date of the last release and commit, the response time per issue and other GitHub API data. Currently, when searching for a package, the most popular package is not always good and this can leads to package selection errors, like here https://github.com/symfony/symfony/pull/49831
![Selection_1379](https://user-images.githubusercontent.com/21358010/228387633-92e65e54-e62f-4a7d-b2ff-725e0d79f867.png)

Spomky:
I agree with this idea. It would be very nive to spot unmaintained or outdated libraries 👍🏽.

zerai:
https://isitmaintained.com/#metrics

94noni:
perhaps related/linked with https://github.com/composer/packagist/issues/1320
maybe we can add the latest version and its release date

(of course the date does not mean anything when a package is battle tested and simple thus not require any further update)

94noni:
PR as discussion draft opened here https://github.com/composer/packagist/pull/1410
Feel free to comment :)

Issue #1320, opened 2022-07-28
[RFC] displaying latest version in search engine view result


94noni:
Hi

I would like to know if it can be useful to display such info on http://packagist.org/ website

here is an example:
<img width="882" alt="Capture d’écran 2022-07-28 à 14 08 27" src="https://user-images.githubusercontent.com/1358361/181501501-35ccd14a-a2dd-4d1d-8869-2e7e5495d8e1.png">

---

We can imagine storing such data in the package when a new version is added to it
And index such data when indexing all data for the search engine

Seldaek:
What's the value of knowing the last version in a listing though?

IonBazan:
I think npm provides similar feature:
![image](https://user-images.githubusercontent.com/1985514/188569159-84e11a19-f6ca-4de7-ad2b-00eb9df3164f.png)
![image](https://user-images.githubusercontent.com/1985514/188569204-17f9b934-aae7-4130-839d-0f11aaf5a154.png)

Not sure if it provides much value here but I think the last publish date would be helpful here to decide which package is still actively maintained without going to the details page.

94noni:
@Seldaek @IonBazan thx for stopping here

> What's the value of knowing the last version in a listing though?

mostly yes to know how a package evolve, some projects has little deps tho knowing a packages have new versions is good
also yes with the date, both could be good (mimic the npm page)

it was just an idea I got this summer tho wanted to write it down, if you think it is not that relevant, could be closed for now, thanks

chr-hertel:
I would argue that the search results should give info that helps a developer to choose if a package is worth investigating or not - the detail view will be used anyways btw.
I can see how stars and downloads help with that, but i don't think the version number is really relevant.
the only thing i can see is that it would be worth providing an info about stability of a package, like is there a stable version or not for example

Seldaek:
I can see some value in the last release date to see if something's been "dead" for years, although in some cases I'd argue stuff is just stable and working and doesn't need work and thus it's fine. For example https://packagist.org/packages/seld/cli-prompt is almost 2y stale and before that it was almost 3y.. yet it "just works".

So even the date needs to be taken with a grain of salt, and newer isn't necessarily better.

The version is even less "comparable" between packages IMO, so why show it in a list? 

Overall not so convinced that more info is more value.

94noni:
@Seldaek @chr-hertel fair enough, thx for taking time to comment

then we can consider closing? if it does not add much value for majority ppl

[...]

Issue #1290, opened 2022-04-08
Search Enhancements


rotexdegba:
I want to really thank you all for the good job you are doing maintaining the packagist website. I was just wondering if the search could include a way to filter out abandoned packages and order by either date a commit was last made or date of last commit for the list of packages returned in a search result.

94noni:
PR as discussion draft opened here https://github.com/composer/packagist/pull/1410
Feel free to comment :)

8ctopus:
Fully agree, it would be great to add filters for:
- abandoned package
- php version compatibility
- latest commit

PR #1410, opened 2024-01-08
feat(search): add snapshoted latest version and date of a package

94noni:
Hello,

Small PR related to issues mentioned/related.

> [!NOTE]
> PR opened as is to gather feedbacks/have raw material for discussions (not finished at all)
> I am open to discuss the easiest/fastest/simplest way for introducing such data that will help a think a lot the search results

---

## Idea of display

Capture d’écran 2024-01-08 à 18 47 42

---

refs:

- https://github.com/composer/packagist/issues/1320 (old one of mine)
- https://github.com/composer/packagist/issues/1290
- https://github.com/composer/packagist/issues/1368

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -174,6 +174,8 @@ search.addWidget(
         <div class="col-sm-3 col-lg-2">
             {{#meta}}
                 <p class="metadata">
+                    <span class="metadata-block"><i class="glyphicon glyphicon-tags"></i> {{ meta.latest_version }}</span>
+                    <span class="metadata-block"><i class="glyphicon glyphicon-calendar"></i> {{ meta.latest_date }}</span>
                     <span class="metadata-block"><i class="glyphicon glyphicon-download"></i> {{ meta.downloads_formatted }}</span>
                     <span class="metadata-block"><i class="glyphicon glyphicon-star"></i> {{ meta.favers_formatted }}</span>
                 </p>
--- a/src/Entity/Package.php
+++ b/src/Entity/Package.php
@@ -178,6 +178,9 @@ class Package
      */
     private array|null $cachedVersions = null;
 
+    #[ORM\Column(type: 'string', length: 255, nullable: true)]
+    private string|null $snapshotedLatestVersion = null;
+
     public function __construct()
     {
         $this->versions = new ArrayCollection();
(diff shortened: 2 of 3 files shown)


src/Entity/Package.php
```diff
@@ -539,6 +542,14 @@ public function getBrowsableRepository(): string
     public function addVersion(Version $version): void
     {
         $this->versions[] = $version;
+
+        $newVersion = strtolower($version->getNormalizedVersion());
+
+        if (null === $this->snapshotedLatestVersion
+            ||
+            version_compare($newVersion, $this->snapshotedLatestVersion, '>=')) {
```
stof, line 550: `version_compare` looks wrong to me here given that you don't only include the version when writing in the property.

src/Entity/Package.php
```diff
@@ -539,6 +542,14 @@ public function getBrowsableRepository(): string
...
+            $this->snapshotedLatestVersion = $newVersion . '&' . (new \DateTime('now'))->format('Y-m-d');
```
stof, line 551: the date of adding the data in the entity is not the date of that version (it depends when the sync happens)

js/search.js
```diff
@@ -174,6 +174,8 @@ search.addWidget(
         <div class="col-sm-3 col-lg-2">
             {{#meta}}
                 <p class="metadata">
+                    <span class="metadata-block"><i class="glyphicon glyphicon-tags"></i> {{ meta.latest_version }}</span>
+                    <span class="metadata-block"><i class="glyphicon glyphicon-calendar"></i> {{ meta.latest_date }}</span>
```
stof, line 178: what if those are not available ?

src/Entity/Package.php
94noni: indeed! we may leverage `Version::releasedAt` here?

js/search.js
94noni: i have not install/run the project locally for now, dont want to invest much time of PR is not accepted
so for now, it was just to make a decent screenshot with fake data ^^

src/Entity/Package.php
94noni: indeed it was my original idea, 2 fields, but for snapshots data I went for one
i handle here the explode

PR #1411, opened 2024-01-10
Repo jacking protection and cleanups

Seldaek:
Replaces https://github.com/composer/packagist/pull/1402

### DIFF
--- a/src/Entity/Package.php
+++ b/src/Entity/Package.php
@@ -37,6 +37,7 @@ enum PackageFreezeReason: string
 {
     case Spam = 'spam';
     case RemoteIdMismatch = 'remote_id';
+    case Gone = 'gone';
 }
 
 /**
@@ -52,8 +53,9 @@ enum PackageFreezeReason: string
 #[ORM\Index(name: 'dumped2_idx', columns: ['dumpedAtV2'])]
 #[ORM\Index(name: 'repository_idx', columns: ['repository'])]
 #[ORM\Index(name: 'remoteid_idx', columns: ['remoteId'])]
-#[ORM\Index(name: 'dumped2_crawled_idx', columns: ['dumpedAtV2', 'crawledAt'])]
+#[ORM\Index(name: 'dumped2_crawled_frozen_idx', columns: ['dumpedAtV2', 'crawledAt', 'frozen'])]
 #[ORM\Index(name: 'vendor_idx', columns: ['vendor'])]
+#[ORM\Index(name: 'frozen_idx', columns: ['frozen'])]
 #[UniquePackage(groups: ['Create'])]
 #[VendorWritable(groups: ['Create'])]
 #[ValidPackageRepository(groups: ['Update', 'Default'])]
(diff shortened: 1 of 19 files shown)


src/Entity/Package.php
```diff
@@ -33,6 +33,13 @@
 use Composer\Util\HttpDownloader;
 use DateTimeInterface;
 
+enum PackageFreezeReason: string
```
stof, line 36: why defining it inline instead of making it autoloadable ?

templates/macros.html.twig
```diff
@@ -1,6 +1,6 @@
 {% macro listPackages(packages, paginate, showAutoUpdateWarning, meta) %}
     <ul class="packages list-unstyled">
-        {% for package in packages|filter(package => package.replacementPackage|default('') != 'spam/spam') %}
+        {% for package in packages|filter(package => not package.isFrozen() or package.getFreezeReason().value != 'spam') %}
```
glaubinix, line 3: Could instead add an isSpam() method to the enum, if you want to get rid of the magic spam string here.

src/Entity/Package.php
Seldaek: Because it belongs together with the class, and is much more readable to me than having tiny enum files lingering around, and there is no reason to use it without package instance, so the package class will be loaded when you need it.

naderman (approved):

src/Entity/Package.php
```diff
@@ -33,6 +33,33 @@
...
+{
+    case Spam = 'spam';
+    case RemoteIdMismatch = 'remote_id';
+    case Gone = 'gone';
+
+    public function translationKey(): string
+    {
+        return 'freezing_reasons.' . $this->value;
```
stof, line 44: if you want to make those translations extractable, a solution would be to make a method returning a TranslatableMessage object (that you would still use in Twig by passing it to the `trans` filter), using a `match()` statement instead of concatenation (as dynamic keys cannot be extracted by the tooling)

templates/package/view_package.html.twig
```diff
@@ -119,7 +119,16 @@
...
                                 </form>
                             {% endif %}
+                            {% if is_granted('ROLE_ADMIN') and package.isFrozen() %}
+                                <form class="action" action="{{ path("unfreeze_package", {name: package.name}) }}" method="POST">
+                                    <input type="hidden" name="token" value="{{ unfreezeCsrfToken }}" />
+                                    <input class="btn btn-danger" type="submit" value="Unfreeze" />
+                                </form>
+                            {% endif %}
                         </div>
+                        {% if package.isFrozen() %}
```
stof, line 129: should this be displayed only for package maintainers and admins (as done right now by being inside the `{% if has_actions %}`) or to all users ?

Seldaek: Maybe we should open it to everyone, I don't expect this to be too frequent anyway.. And it'd be valuable to see the warning for the `gone` packages.

src/Entity/Package.php
Seldaek: Thanks that's good to know but I don't really care about extractability here.

PR #1413, opened 2024-01-12
detect dev for install command

### LINKED ISSUE #1412: Respect 'dev' keyword in packagist UI to show composer require --dev
Composer itself recognizes [few keywords to distinguish dev packages](https://getcomposer.org/doc/04-schema.md#keywords) (like `dev` or `static analysis`), but packagist does not seem to respect those. I believe it should show `composer require --dev` for those. [Example package](https://packagist.org/packages/shipmonk/phpstan-rules).

image

chr-hertel:
Fixes #1412

### DIFF
--- a/src/Entity/Package.php
+++ b/src/Entity/Package.php
@@ -687,16 +687,14 @@ class Package
         return $this->type;
     }
 
-    public function getInstallCommand(): string
+    public function getInstallCommand(Version $version = null): string
     {
         $command = 'create-project';
 
         if ('project' !== $this->getType()) {
             $command = 'require';
 
-            /** @var Version $version */
-            $version = $this->getVersions()->first();
-            if ($version && $version->hasDevTag()) {
+            if (null !== $version && $version->hasDevTag()) {
                 $command .= ' --dev';
             }
         }
--- a/src/Entity/Package.php
+++ b/src/Entity/Package.php
@@ -687,6 +687,23 @@ class Package
         return $this->type;
     }
 
+    public function getInstallCommand(): string
+    {
+        $command = 'create-project';
+
+        if ('project' !== $this->getType()) {
+            $command = 'require';
+
+            /** @var Version $version */
+            $version = $this->getVersions()->first();
+            if ($version && $version->hasDevTag()) {
+                $command .= ' --dev';
+            }
+        }
+
+        return sprintf('composer %s %s', $command, $this->getName());
+    }
+
     public function setRemoteId(string|null $remoteId): void
     {
         $this->remoteId = $remoteId;
--- a/src/Entity/Version.php
+++ b/src/Entity/Version.php
@@ -635,6 +635,17 @@ class Version
         return $this->tags;
     }
 
+    public function hasDevTag(): bool
+    {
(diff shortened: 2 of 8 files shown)


src/Entity/Package.php
```diff
@@ -687,6 +687,23 @@ public function getType(): string|null
...
 
+    public function getInstallCommand(): string
+    {
+        $command = 'create-project';
+
+        if ('project' !== $this->getType()) {
+            $command = 'require';
+
+            /** @var Version $version */
+            $version = $this->getVersions()->first();
```
chr-hertel, line 698: not sure which version to use ...

src/Entity/Version.php
```diff
@@ -635,6 +635,13 @@ public function getTags(): Collection
         return $this->tags;
     }
 
+    public function hasDevTag(): bool
+    {
+        return [] !== array_filter($this->getTags()->toArray(), static function (Tag $tag) {
```
stof, line 640: looping over the tags and stopping when we find a dev tag would be more efficient than filtering the whole list all the time.

src/Entity/Package.php
stof: The `view_package` template already receive a `version` variable (which can be `null`). this is the version that should be used for the install command IMO (which would be done by keeping the command in the template as it is done right now, but adding a `{% if version and version.hasDevTag %} --dev{% endif %}`)

src/Entity/Version.php
chr-hertel: true

src/Entity/Package.php
chr-hertel: yea, i know that moving it from the template to PHP feels like the more complex solution at first, but i tend to kill every if in twig that i can replace with sth that is easily unit tested.

but sure, that's a debatable design decision

stof: Twig already has the right version to use passed to it by the controller. The entity does not have it. So moving this to the entity itself looks like the bad decision IMO.

chr-hertel: wasn't aware of that controller code to sort out the most relevant version - thanks for that pointer!
will give it another thought later :+1:

Seldaek: IMO this function should just accept a version as argument, and then the template can pass it in.

Seldaek:
Thanks!

templates/package/view_package.html.twig
```diff
@@ -46,7 +46,7 @@
 
             <div class="row">
                 <div class="col-md-8">
-                    <p class="requireme"><i class="glyphicon glyphicon-save"></i> <input type="text" readonly="readonly" value="composer {% if package.type == 'project' %}create-project{% else %}require{% endif %} {{ "#{package.vendor}/#{package.packageName}" }}" /></p>
+                    <p class="requireme"><i class="glyphicon glyphicon-save"></i> <input type="text" readonly="readonly" value="{{ package.installCommand(expandedVersion) }}" /></p>
```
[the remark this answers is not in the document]
Seldaek: Yes true it should be version, I'll fix that.

PR #1414, opened 2024-01-12
Bump to PHP 8.3 & Symfony 7

chr-hertel:
Just wanted to give it a shot - more like to see what happens. Don't get why this was that easy, but didn't/wasn't able to test everything.

Feel free to take over ... or close oc


config/packages/scheb_2fa.yaml
```diff
@@ -12,7 +12,7 @@ scheb_two_factor:
         enabled: true
         server_name: '%env(APP_HOSTNAME)%'
         issuer: Packagist
-        window: 1
+        leeway: 10
```
chr-hertel, line 15: wasn't able to test that locally yet ...

https://github.com/scheb/2fa/commit/e9f23f5b438ad2d48d99164fc98ae14a8d538fbd

Chris53897:
I think this should be seperated PRs.
Maybe one to update scheb/2fa-* 6.x to 7.x
One for UserProviderInterface (if that is unrelated from symfony 6 / 7 choice)
One for snc/redis-bundle
After that, one for symfony 7 and php 8.2 as min version. https://symfony.com/releases/7.0
Users should not be enforced to use the latest version.

Seldaek:
@chr-hertel servers are now running php8.3, do you mind rebasing here? I'll then take a look at merging and doing some more testing :)

Chris53897:
Please run `composer recipes` and update the recipes for the Symfony 7 upgrade.

Seldaek:
There was a change to user table. Make sure db schema is up to date then login should work after a retry perhaps

chr-hertel:
oof, recipes update was annoying, but i hope i got it right.

not sure about following changes:
* recipe of symfony/lock was in strange state. not sure about the env var `LOCK_DSN=flock` there or what you're using, see https://symfony.com/doc/current/lock.html#configuring
* discarded everything from phpunit recipe. not sure about the value of that thing tbh :speak_no_evil: 
* doctrine bundle changes, rolled them back with 8a91af64f9968eb990c775182b53fb6506880abd

config/packages/scheb_2fa.yaml
Seldaek: Works 👍🏻

Seldaek:
Thanks @chr-hertel !

PR #1422, opened 2024-02-27
Logout: add CSRF token and clear site data

glaubinix:
Clears the browser cookies, cache + storage on logout. Storage is currently not used as far as I can tell.


stof:
Well, on logout, the remember-me cookie should be deleted by Symfony anyway. So I don't understand why you could not clear it.

glaubinix:
Thanks for the clarification! Updated.

Seldaek:
Thanks

PR #1423, opened 2024-02-28
Fixed type (changed 'we' to "we'd")

tristanisham:
Hi, this is just a minor fix for a typo on the API docs page. 

image

`we rather not do` -> `we'd rather not do`


Seldaek:
Thanks

PR #1426, opened 2024-03-03
Email users on account security changes

### LINKED ISSUE #1419: Email users on account security changes to alert them to potential compromise
Users should be notified by email about critical account security changes, such as:

- password change
- username change
- disabling two factor auth
- (dis)connecting with GitHub
- email change (send to old email)

The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.

Fan2Shrek:
Fixes #1419

Notifying:

- Username change
- Email change
- Password change
- (Dis)connect github

Move the two-factor authentication notifier in an dedicated class

### DIFF
--- a/src/Controller/ChangePasswordController.php
+++ b/src/Controller/ChangePasswordController.php
@@ -14,6 +14,7 @@ namespace App\Controller;
 
 use App\Entity\User;
 use App\Form\ChangePasswordFormType;
+use App\Security\UserNotifier;
 use Symfony\Component\Security\Http\Attribute\IsGranted;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -25,7 +26,7 @@ class ChangePasswordController extends Controller
 {
     #[IsGranted('ROLE_USER')]
     #[Route(path: '/profile/change-password', name: 'change_password')]
-    public function changePasswordAction(Request $request, UserPasswordHasherInterface $passwordHasher, #[CurrentUser] User $user): Response
+    public function changePasswordAction(Request $request, UserPasswordHasherInterface $passwordHasher, UserNotifier $userNotifier, #[CurrentUser] User $user): Response
     {
         $form = $this->createForm(ChangePasswordFormType::class, $user);
         $form->handleRequest($request);
@@ -41,6 +42,8 @@ class ChangePasswordController extends Controller
 
             $this->getEM()->persist($user);
             $this->getEM()->flush();
+
+            $userNotifier->notifyChange($user->getEmail(), 'Your password has been changed');
 
             return $this->redirectToRoute('my_profile');
         }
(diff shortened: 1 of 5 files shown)


src/Controller/ProfileController.php
```diff
@@ -124,6 +125,17 @@ public function editAction(Request $request): Response
...
         if ($form->isSubmitted() && $form->isValid()) {
+            $uow = $this->getEM()->getUnitOfWork();
+            $uow->computeChangeSets();
+            $changeSet = $uow->getEntityChangeSet($user);
+
+            $diffs = array_intersect(array_keys($changeSet), ['email', 'username']);
+
+            if (!empty($diffs) && $user instanceof User) {
+                $reason = sprintf('Your %s has been changed', implode(' and ', $diffs));
+                $userNotifier->notifyChange($changeSet['email'][0] ?? $user->getEmail(), $reason);
```
Seldaek, line 136: I think it would be nice to send the email to both old and new emails here when the email changes.

src/Controller/ChangePasswordController.php
```diff
@@ -42,6 +43,8 @@ public function changePasswordAction(Request $request, UserPasswordHasherInterfa
             $this->getEM()->persist($user);
             $this->getEM()->flush();
 
+            $userNotifier->notifyChange($user->getEmail(), 'Your password has been changed');
```
Seldaek, line 46: It would be nice to reuse this UserNotifier for the 2fa emails too I think, to keep it all the same, if you don't mind refactoring that. It probably needs the ability to pass in a custom email template filename tho as the 2fa emails contain some extra wording.

src/Controller/ProfileController.php
```diff
@@ -112,7 +113,7 @@ public function packagesAction(Request $req, #[VarName('name')] User $user, Favo
     }
 
     #[Route(path: '/profile/edit', name: 'edit_profile')]
-    public function editAction(Request $request): Response
+    public function editAction(Request $request, UserNotifier $userNotifier): Response
     {
         $user = $this->getUser();
         if (!is_object($user)) {
```
Seldaek, line 119: Just an idea, doing this here instead would remove the need for the check below.. it keeps things a little cleaner in the main logic of the function

```suggestion
        if (!$user instanceof User) {
```

templates/email/alert_change.txt.twig
```diff
@@ -0,0 +1,10 @@
+{% autoescape false -%}
+
+A critical account security change has been made to your account:
+
+-------------------------------
+Reason: {{ reason }}
```
Seldaek, line 6:
```suggestion
Change: {{ reason }}
```

src/Controller/ChangePasswordController.php
Seldaek: I don't see this done but no worries I'll do it after merging :)

Seldaek:
Thanks!

src/Controller/ChangePasswordController.php
Fan2Shrek: Woops I havent add the file sorry x)
I can made a PR fixing that later if you want :)

PR #1428, opened 2024-03-06
Protect user password fields with ReCaptcha after 3 failed attempts

glaubinix:
The 2FA code on login is already protected by the Symfony bundle that limits 5 attempts in 15min.

### DIFF
--- a/src/Security/RecaptchaContext.php
+++ b/src/Security/RecaptchaContext.php
@@ -22,7 +22,7 @@ class RecaptchaContext
     public function __construct(
         public readonly string $ip,
         public readonly string $username,
-        public readonly ?string $recaptcha,
+        public readonly bool $hasRecaptcha,
     ) {}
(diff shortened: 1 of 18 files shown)


src/Security/RecaptchaContext.php
```diff
@@ -0,0 +1,47 @@
...
+            $this->username ? self::LOGIN_BASE_KEY_USER . strtolower($this->username) : null,
+        ]);
+    }
+
+    public static function fromRequest(Request $request): self
+    {
+        return new self(
+            $request->getClientIp() ?: '',
+            (string) $request->request->get('_username'),
+            (string) $request->request->get('g-recaptcha-response'),
```
stof, line 44: why casting to string ? Shouldn't `null` be preserved ?

Seldaek:
Thanks

PR #1432, opened 2024-03-08
Add node/npm to compose file

pscheit:
Does npm install and npm run build once `docker compose up` is used.

Can put this to https://docs.docker.com/compose/profiles/ if needed. But dont think so.

- `version:` is not read from compose v2 anymore (always uses the latest)
- anonymous volumes can be now created without a top level entry in volume

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ These steps are provided for development purposes only.
 ### Requirements
 
 - **PHP** for the web app
-- **NPM** for the frontend build
+- **NPM** (or Docker) for the frontend build
 - **[Symfony CLI](https://symfony.com/download)** to run the web server
 - **MySQL** (or Docker) for the main data store
 - **Redis** (or Docker) for some functionality (favorites, download statistics)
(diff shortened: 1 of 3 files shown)


chr-hertel:
i usually prefer not to mount the working directory to a container and work with symfony cli and nvm to have the runtimes on host, but that's opinionated i guess - but feels a bit inconsistent to me now to have a node container, but no php one.

what about adding the override to ignore now?

pscheit:
You mean adding the profile, so that its not started automatically?

If this doesn provide value to other Docker users I am fine to not merge this

chr-hertel:
No, the changes make sense and adding a profile to the node one or adding `docker-compose.override.yml` to `.gitignore` or both would be totally compatible with my way to use the docker setup

pscheit:
yeah, this is what I would prefer over profile as well. 

Would add docker-compose.node.yaml then

README.md
```diff
@@ -53,7 +54,7 @@ These steps are provided for development purposes only.
    bin/console doctrine:schema:create
    ```
 7. Run a CRON job `bin/console packagist:run-workers` to make sure packages update.
-8. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files.
+8. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files. When using Docker run `docker-compose run node npm run dev` to watch css/js files.
```
Seldaek, line 57: Shouldn't this be a space here? I know nothing but seems inconsistent :D

```suggestion
8. Run `npm run build` or `npm run dev` to build (or build&watch) css/js files. When using Docker run `docker compose run node npm run dev` to watch css/js files.
```

pscheit: oh yeah totally. Bad finger habit!

pscheit: docker-compose is legacy v1 where as docker compose (compose as a plugin to docker) is correct

Seldaek:
Alright thanks

PR #1439, opened 2024-03-15
Avoid simple passwords (e.g. email as password)

pscheit:
PR changes that you can't use your email address or your username as a password anymore.

We are aware that it's hard in general to get people to use really good passwords, but at least this crosses out the really dumb mistakes.

### DIFF
--- a/tests/Controller/ProfileControllerTest.php
+++ b/tests/Controller/ProfileControllerTest.php
@@ -1,5 +1,15 @@
 <?php declare(strict_types=1);
 
+/*
+ * This file is part of Packagist.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *     Nils Adermann <naderman@naderman.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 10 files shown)


tests/Controller/ProfileControllerTest.php
```diff
@@ -21,6 +31,13 @@ public function setUp(): void
         parent::setUp();
     }
 
+    public function tearDown(): void
+    {
+        static::getContainer()->get(Connection::class)->rollBack();
+
+        parent::tearDown();
+    }
+
```
pscheit, line 40: this is pretty easy to forget and would make a good case for a nice sideffect on the next functional test.

I would suggest we base class some stuff in the functional tests into a ControllerTestCase and let all Controller tests extend from this?  
Would allow to introduce a getEM() on tests as well

Would of course do this in another PR.

Seldaek: Yeah sounds reasonable

Seldaek:
Thanks

PR #1441, opened 2024-03-20
Introduce a ControllerTestCase with getEm(), assertFormError()

pscheit:
as mentioned in #1439 so that we dont forget to setup/teardown and have a transaction everywhere.


Seldaek:
Thanks

PR #1443, opened 2024-03-23
Algolia Instantsearch Update

chr-hertel:
Just gave that a first shot. Not that trivial unfortunately - at least for me ...

pushing this only if someone wants to help or give it a shot as well.

Current issues I'm aware of
- [ ] couple of deprecations still, e.g. templating
- [ ] routeToState does not initiate a search
- [ ] cursor not at the end of input on focus with query param
- [ ] pagination still visible on zero hits
- [ ] broken UI for hits and facets

image

replaces #1427

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -14,11 +14,12 @@ const search = instantsearch({
     routing: {
         stateMapping: {
             stateToRoute: function (uiState) {
+                const state = uiState[algoliaConfig.index_name];
                 return {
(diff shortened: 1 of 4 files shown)


Seldaek:
Yeah this is unfortunately quite a mess, hence why I happily ignored the update so far :D

chr-hertel:
Sorry, never got to that in the end ... I guess I would start over at zero anyways.

PR #1492, opened 2024-11-27
Update PHPStan to 2.x

Seldaek:


Seldaek:
@staabm if I can bug you here.. I left 3 errors out of the baseline because I think they might be bugs in phpstan-dba. At least the last two errors. The first one I think might be a phpstan-doctrine issue rather.

```
Error 1: Property App\Entity\EmptyReferenceCache::$emptyReferences type     
         mapping mismatch: database can contain array but property expects  
         list<string>.                                                      
```

I'm not sure if there is a way to express that the DB contains only lists and not arrays.. This would have to be on a "trust me it's true" basis of course. Probably a feature request for phpstan-doctrine?

```
Error 2: Method App\Entity\PackageRepository::getSuggestCount() should return  
         int<0, max> but returns int.    
```

I think that could be fixed in phpstan-dba.. a `COUNT(*)` in a query is guaranteed to produce `int<0,max>` I believe?

```
Error 3: Method App\Entity\PhpStatRepository::getStatVersions() should return  
         list<array{version: string, depth: 0|1|2|3}> but returns              
         array<mixed>.   
```

Not sure what's going on there.. the code is [quite simple](https://github.com/composer/packagist/blob/36fc31c61c4310981046ec924de5b37b6bf07ff7/src/Entity/PhpStatRepository.php#L36-L45) so I don't understand why it falls back to array<mixed> here. Is `getArrayResult` not supported perhaps?

staabm:
> Error 1: Property App\Entity\EmptyReferenceCache::$emptyReferences type     
         mapping mismatch: database can contain array but property expects  
         list<string>.                                                      

I guess its a phpstan-doctrine thing.. it doesn't look like related to DBA but more like ORM?

> I think that could be fixed in phpstan-dba.. a `COUNT(*)` in a query is guaranteed to produce `int<0,max>` I believe?

that should already be supported, in case you enable `$config->utilizeSqlAst(true);`





> Not sure what's going on there.. the code is [quite simple](https://github.com/composer/packagist/blob/36fc31c61c4310981046ec924de5b37b6bf07ff7/src/Entity/PhpStatRepository.php#L36-L45) so I don't understand why it falls back to array here. Is `getArrayResult` not supported perhaps?

the query build is not supported in phpstan-dba until now

staabm:
> Not sure what's going on there.. the code is [quite simple](https://github.com/composer/packagist/blob/36fc31c61c4310981046ec924de5b37b6bf07ff7/src/Entity/PhpStatRepository.php#L36-L45) so I don't understand why it falls back to array here. Is `getArrayResult` not supported perhaps?

maybe thats something phpstan-doctrine can provide, as they already do the query-builder introspection for typing

stof:
I think phpstan-doctrine indeed does not support refining types for array results, only for object hydration.

PR #1511, opened 2025-03-08
Fix display URL for the package download stats example

mbabker:
The actual link URL and the docs are correct, but the URL shown in the example missed the `/packages` segment.


Seldaek:
Oh yup, thanks

PR #1533, opened 2025-05-07
feat: add package version delete api

shyim:
## Fixed Edit Package API

```
❯ curl -X PUT 'https://packagist.org/api/packages/monolog/monolog?username=admin&apiToken=********' -d '{"repository":"https://github.com/Seldaek/monolog"}'
{"type":"https:\/\/tools.ietf.org\/html\/rfc2616#section-10","title":"An error occurred","status":500,"detail":"Internal Server Error"}⏎
```

and Docs with the correct HTTP method

## New Delete Package Version API

Similar to whats right now possible to be able to delete a package version

### DIFF
--- a/src/Controller/ApiController.php
+++ b/src/Controller/ApiController.php
@@ -16,6 +16,7 @@ use App\Entity\Package;
 use App\Entity\SecurityAdvisory;
 use App\Entity\User;
 use App\Entity\Vendor;
+use App\Entity\Version;
 use App\Model\DownloadManager;
 use App\Model\ProviderManager;
 use App\Model\VersionIdCache;
@@ -24,6 +25,7 @@ use App\Service\GitHubUserMigrationWorker;
 use App\Service\Scheduler;
 use App\Util\UserAgentParser;
 use Composer\Pcre\Preg;
+use Symfony\Bridge\Doctrine\Attribute\MapEntity;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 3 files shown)


src/Controller/ApiController.php
```diff
@@ -225,6 +227,35 @@ public function editPackageAction(Request $request, Package $package, ValidatorI
...
+        $packageVersion = $package->getVersions()->findFirst(static function (int $index, Version $innerVersion) use ($version) {
+            return $innerVersion->getVersion() === $version;
+        });
+
+        if (!$packageVersion) {
+            return new JsonResponse(['status' => 'error', 'message' => 'Version not found'], 404);
+        }
+
+        $repo = $this->getEM()->getRepository(Version::class);
+        $repo->remove($packageVersion);
```
stof, line 252: this should be wrapped in `$em->transactional()` (which calls `flush()` automatically at the end) or `$em->getConnection()->transactional()` (keeping an explicit flush of the entity manager inside the closure) to ensure that the removal is done in a transactional way.

src/Controller/ApiController.php
```diff
@@ -183,14 +185,14 @@ public function updatePackageAction(Request $request, string $githubWebhookSecre
     }
 
     #[Route(path: '/api/packages/{package}', name: 'api_edit_package', requirements: ['package' => '[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+?'], defaults: ['_format' => 'json'], methods: ['PUT'])]
-    public function editPackageAction(Request $request, Package $package, ValidatorInterface $validator, StatsDClient $statsd): JsonResponse
+    public function editPackageAction(Request $request, #[MapEntity(mapping: ['package' => 'name'])] Package $package, ValidatorInterface $validator, StatsDClient $statsd): JsonResponse
```
stof, line 188: this looks like a different change than adding a new API endpoint.

PR #1568, opened 2025-08-22
Move readme data to a join table to avoid heavy joins on package data

Seldaek:
Migration:

```sql
CREATE TABLE package_readme (package_id INT NOT NULL, contents LONGTEXT NOT NULL, PRIMARY KEY (package_id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci`;
ALTER TABLE package ADD readme_id INT DEFAULT NULL;
ALTER TABLE package ADD CONSTRAINT FK_DE686795EAA9F3D0 FOREIGN KEY (readme_id) REFERENCES package_readme (package_id);
CREATE UNIQUE INDEX UNIQ_DE686795EAA9F3D0 ON package (readme_id);

INSERT INTO package_readme (package_id, contents)
SELECT id, readme
FROM package
WHERE readme IS NOT NULL AND readme != '';

UPDATE package SET readme_id = id WHERE readme IS NOT NULL AND readme != '';

-- deploy here

ALTER TABLE package DROP COLUMN readme;
```


src/Entity/Package.php
```diff
@@ -113,8 +113,9 @@ class Package
     #[ORM\Column(type: 'string', nullable: true)]
     private ?string $language = null;
 
-    #[ORM\Column(type: 'text', nullable: true)]
-    private ?string $readme = null;
+    #[ORM\OneToOne(targetEntity: PackageReadme::class, cascade: ['persist', 'remove'], orphanRemoval: true)]
```
Seldaek, line 116: cascade remove doesn't seem to work when deleting the package the readme entity is left behind, need to investigate why not.

glaubinix: You need `#[JoinColumn(nullable: false, onDelete: 'CASCADE')]` for DB level cascade.

stof: @glaubinix no. A DB-level case will delete the row in `package` when deleting the `package_readme` (which is probably not something we want)

stof: @Seldaek `cascade: 'remove'` will delete the PackageReadme when deleting the Package, **using the ORM** (as this is an ORM-level cascading feature). If you delete packages in SQL, this won't have any impact and you need to delete the `package_readme` manually.

But there is a case where you need to delete only the PackageReadme: when assigning `$this->readme = null;` in `setReadme`. Maybe `orphanRemoval` works for that case but not sure.

src/Entity/PackageReadme.php
```diff
@@ -0,0 +1,46 @@
...
+#[ORM\Entity]
+#[ORM\Table(name: 'package_readme')]
+class PackageReadme
+{
+    public function __construct(
+        #[ORM\Id]
+        #[ORM\Column(type: 'integer', name: 'package_id')]
+        public int $packageId,
+
+        #[ORM\Column(type: 'text', nullable: true)]
```
stof, line 29: why is it nullable ? The Package entity expects to have no PackageReadme at all when the contents is null.

src/Entity/Package.php
glaubinix: oh right, this only works if you have the `OneToOne` definition on the `PackageReadme`

glaubinix: I think this won't help much. Won't Doctrine try to fetch the readme entity every time you load a package entity because it needs to resolve the readme field?

Seldaek: > `cascade: 'remove'` will delete the PackageReadme when deleting the Package, **using the ORM** (as this is an ORM-level cascading feature).

yeah except I tried exactly that and it doesn't work.



> Maybe `orphanRemoval` works for that case but not sure

That works when setting null, which I guess I'll do as well before deleting the package with the ORM then..



> Won't Doctrine try to fetch the readme entity every time you load a package entity because it needs to resolve the readme field?

Yes that is why I did this whole mess this way, because I don't want it to eagerly fetch the readme always.

src/Entity/PackageReadme.php
```diff
@@ -0,0 +1,43 @@
...
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ */
+#[ORM\Entity]
+#[ORM\Table(name: 'package_readme')]
+class PackageReadme
+{
+    public function __construct(
+        #[ORM\Id]
+        #[ORM\OneToOne(targetEntity: Package::class)]
+        #[ORM\JoinColumn(referencedColumnName: 'id', nullable: true, onDelete: 'CASCADE')]
```
Seldaek, line 27: delete cascade works well this way 👍🏻

glaubinix, line 27:
```suggestion
        #[ORM\JoinColumn(referencedColumnName: 'id', nullable: false, onDelete: 'CASCADE')]
```

templates/package/view_package.html.twig
```diff
@@ -342,11 +342,11 @@
                 <p class="col-xs-12">This package has no released version yet, and little information is available.</p>
             {% endif %}
 
-            {% if package.readme != null and (not package.isSuspect() or hasActions) %}
+            {% if readme is not empty and (not package.isSuspect() or hasActions) %}
```
stof, line 345:
```suggestion
            {% if readme is not null and (not package.isSuspect() or hasActions) %}
```

Seldaek: Thanks, fixed

PR #1576, opened 2025-09-01
Add pie information and pie packages to provider pages

### LINKED ISSUE #1489: The package page of a PIE-compatible extension should display the extension metadata
Extensions (i.e. packages with type `php-ext` or `php-ext-zend`) have specific metadata in the `php-ext` field. It would be great to display this metadata (the extension name, whether they support ZTS or NTS builds of PHP, etc...) in the package display page.

### LINKED ISSUE #1494: Providers page does not list extensions
The providers page for packages (e.g. [this one for ext-mongodb](https://packagist.org/providers/ext-mongodb)) does not list any `php-ext` packages that provide the corresponding extension (which in this case would be [mongodb/mongodb-extension](https://packagist.org/packages/mongodb/mongodb-extension)). For any `ext-*` packages, I'd expect the corresponding `php-ext` package to be listed first, with alternatives listed later. This would improve discoverability of extension packages.

Seldaek:
Fixes #1494, fixes #1489

cc @alcaeus @stof

image

I added a pie too next to pie packages in listings

image

### DIFF
--- a/src/Entity/Version.php
+++ b/src/Entity/Version.php
@@ -710,6 +710,41 @@ class Version
         $this->phpExt = $phpExt;
     }
 
+    public function getPieName(): ?string
+    {
+        if (!$this->isPiePackage()) {
+            return null;
+        }
+        $name = $this->getPhpExt()['extension-name'] ?? explode('/', $this->getName())[1];
+
+        if (!str_starts_with($name, 'ext-')) {
+            $name = 'ext-'.$name;
+        }
+
+        return $name;
+    }
+
+    /**
+     * @return list<string>
+     */
+    public function getPieSupportedOS(): array
+    {
+        if (!$this->isPiePackage()) {
+            return [];
+        }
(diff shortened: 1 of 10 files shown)


src/Entity/Version.php
```diff
@@ -710,6 +710,41 @@ public function setPhpExt(?array $phpExt): void
...
+    {
+        if (!$this->isPiePackage()) {
+            return [];
+        }
+
+        if (isset($this->phpExt['os-families'])) {
+            return $this->phpExt['os-families'];
+        }
+
+        return array_values(array_diff(["windows", "bsd", "darwin", "solaris", "linux", "unknown"], $this->phpExt['os-families-exclude'] ?? ['any']));
```
stof, line 740: shouldn't this use `?? []` for the case of no exclusions ? I would make more sense than using `any` as a string that is not one of the supported families.

Issue #1557, opened 2025-07-30
Border offset on login field


Krinkle:
On the login page at https://packagist.org/login/, the border of the two input fields are both off from the icon by 1 or 2 pixels:

<img width="1132" height="448" alt="Image" src="https://github.com/user-attachments/assets/836ca6c7-55c9-4d55-8424-0d7443753e97" />

Seldaek:
Yeah it's been like that for ages.. Just not really high up on my list of things to investigate but feel free if you do have time :)

artengin:
Hi!
I’ve fixed this issue — turns out it just needed a quick height: auto in one spot 😀

aihimel:
In commit 35b5ebc0414e56b34c298d2acc11c3b5bf96b023 , `padding-bottom: 7px;` in app.scss file is creating issue in all the login, register and 2fa forms. Check following images.

Will open a PR shortly, which fixes the issue.

<img width="1054" height="703" alt="Image" src="https://github.com/user-attachments/assets/0478c866-7fff-44ec-b026-d47d21d1e268" />
<img width="839" height="735" alt="Image" src="https://github.com/user-attachments/assets/5bcdc2f6-97fd-4d97-9df5-2ad4b27c299b" />
<img width="1507" height="651" alt="Image" src="https://github.com/user-attachments/assets/279fb7f1-0cc6-47b0-94b4-97510ff5b446" />
<img width="884" height="705" alt="Image" src="https://github.com/user-attachments/assets/9cf8b2c8-0e3e-405b-850e-fffa99203b97" />

PR #1583, opened 2025-09-12
fix: border offset on login field

artengin:
Refs: https://github.com/composer/packagist/issues/1557


Krinkle (approved):
Verified, both the modal and standalone page LGTM after this.

Seldaek:
Thanks

PR #1608, opened 2025-11-07
Log entry display

IgorBenko:
Each audit log type now has its own display class and Twig template in order to display each type in an optimized manner. Also fixes the table layout by adding fixed column widths for date and type.

### DIFF
--- a/css/app.scss
+++ b/css/app.scss
@@ -1863,3 +1863,23 @@ body {
         background-color: #ed9e2e;
     }
 }
+
+.audit-log-table {
+    table-layout: fixed;
+
+    .audit-log-datetime {
+        width: 200px;
+
+        @media (max-width: 767px) {
+            width: 130px;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


src/Audit/Display/AbstractAuditLogDisplay.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return $this->datetime;
+    }
+
+    public function getActor(): ActorDisplay
+    {
+        return $this->actor;
+    }
+
+    public function getTypeTranslationKey(): string
```
stof: instead of returning a translation key, I suggest returning a `TranslatableInterface` (probably using the `TranslatableMessage` implementation available in core in each case).

The benefits are:
1. it will make it possible to use placeholders if needed
2. when not using dynamic keys (unlike what it does here), keys can be extracted by the tooling (as it is aware that the constructor of TranslatableMessage takes a translation key)

PR #1609, opened 2025-11-07
Add command to transfer ownership 

stevenrombauts:
Adds a `packagist:transfer-ownership` command to change maintainers for multiple packages at once.

### DIFF
--- a/src/Command/TransferOwnershipCommand.php
+++ b/src/Command/TransferOwnershipCommand.php
@@ -84,7 +84,7 @@ class TransferOwnershipCommand extends Command
      */
     private function queryAndValidateMaintainers(InputInterface $input, OutputInterface $output): array
     {
-        $usernames = array_map('strtolower', $input->getArgument('maintainers'));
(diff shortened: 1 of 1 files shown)


src/Command/TransferOwnershipCommand.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        foreach ($packages as $package) {
+            $oldMaintainers = $package->getMaintainers()->toArray();
+
+            $package->getMaintainers()->clear();
+            foreach ($maintainers as $maintainer) {
+                $package->addMaintainer($maintainer);
+            }
+
+            $this->doctrine->getManager()->persist(AuditRecord::packageTransferred($package, null, $oldMaintainers, array_values($maintainers)));
```
stof: What about packages that were already own by the new maintainers ? Should they actually be marked as transferred to the same maintainers ?

Seldaek: Yeah I think what we could do here is at least check that the old and new arent' an exact match and if they are skip transferring that package as that's a noop. I'll add that post merge.

PR #1616, opened 2025-11-13
Transparency log: add missing display classes

stevenrombauts:
This PR adds the display classes for the maintainer added/removed and transferred records.

### DIFF
--- a/templates/audit_log/display/package_transferred.html.twig
+++ b/templates/audit_log/display/package_transferred.html.twig
@@ -1,6 +1,6 @@
 {% import 'audit_log/macros.html.twig' as auditLog %}
 
 <strong>{{ auditLog.packageLink(display.packageName) }}</strong><br>
-Previous maintainers: {% for maintainer in display.previousMaintainers %}{{ maintainer.username }}{% if not loop.last %}, {% endif %}{% endfor %}<br>
-Current maintainers: {% for maintainer in display.currentMaintainers %}{{ maintainer.username }}{% if not loop.last %}, {% endif %}{% endfor %}<br>
+Previous maintainers: {{ display.previousMaintainers|column('username')|join(', ') }}<br>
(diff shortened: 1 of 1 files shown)


templates/audit_log/display/maintainer_added.html.twig
```diff
@@ -0,0 +1,12 @@
+{% import "macros.html.twig" as packagist %}
+
+<strong>
+    {%- if display.packageName is existing_package -%}
+        <a href="{{ path('view_package', { 'name': display.packageName }) }}">{{ display.packageName }}</a>
+    {%- else -%}
+        {{ display.packageName }}
+    {%- endif -%}
+    {{ packagist.packageLink(display.packageName, null) }}
```
Seldaek: Here you have a double package name it seems

templates/audit_log/display/maintainer_removed.html.twig
```diff
@@ -0,0 +1,9 @@
+<strong>
+    {%- if display.packageName is existing_package -%}
+        <a href="{{ path('view_package', { 'name': display.packageName }) }}">{{ display.packageName }}</a>
+    {%- else -%}
+        {{ display.packageName }}
+    {%- endif -%}
+</strong><br>
+Maintainer: {{ display.maintainer.username }}<br>
+Removed by: {{ display.actor.username }}
```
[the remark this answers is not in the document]
IgorBenko: Also in favor of keeping the current format for better clarity.

templates/audit_log/macros.html.twig
```diff
@@ -0,0 +1,7 @@
+{% macro packageLink(packageName) %}
```
IgorBenko: Good idea

IgorBenko (changes requested):
Looks good to me. I'd just suggest using `join` instead of `for` in the Twig template

PR #1633, opened 2025-12-11
Upgrade to Symfony7.4 and update all deps

Seldaek:


src/Controller/ApiController.php
```diff
@@ -494,8 +495,6 @@ protected function receiveUpdateRequest(Request $request, string $url, string $u
                     $statsd->increment('update_pkg_api.success', tags: ['mode' => 'github_auto']);
                 } else {
                     $statsd->increment('update_pkg_api.failed', tags: ['reason' => 'invalid_gh_sig']);
-                    $this->logger->error('Failed validating GitHub webhook signature', ['sig' => $sig, 'expected' => $expected, 'query_params' => $request->query->all(), 'repo' => json_decode($request->getContent(), true)['repository']['html_url'] ?? null]);
```
stevenrombauts: What's the reason for removing this? The response should be sufficient and/or to stop logging the signature?

PR #1634, opened 2025-12-12
Transparency log: track metadata if version changes

stevenrombauts:
This PR will store metadata in the audit record's attributes whenever a version reference changes:

- If the version is a non-dev version, it will always create a record with metadata
- For dev versions: only create a record if there are actual changes in the metadata detected

### Test instructions

- Create a new non-dev version, check that a record is created
- Create empty commit in a branch, no record should be created
- Change composer.json in the same branch, a new record should be created

### DIFF
--- a/src/EventListener/VersionListener.php
+++ b/src/EventListener/VersionListener.php
@@ -27,7 +27,6 @@ use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
 
 #[AsEntityListener(event: 'postPersist', entity: Version::class)]
 #[AsEntityListener(event: 'postUpdate', entity: Version::class)]
-#[AsEventListener(event: VersionReferenceChangedEvent::class, method: 'onVersionReferenceChanged')]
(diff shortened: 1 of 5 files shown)


src/EventListener/VersionListener.php
```diff
@@ -49,32 +46,22 @@ public function postPersist(Version $version, LifecycleEventArgs $event): void
...
         }
+
+        $record = AuditRecord::versionReferenceChange(
+            $version,
+            $event->getOldSourceRef(),
+            $event->getOldDistRef(),
+            $event->getNewMetadata()
+        );
+
+        $this->getEM()->getRepository(AuditRecord::class)->insert($record);
```
Seldaek: Can we just persist these new records and let them be flushed to DB together with the new version entities? To ensure data integrity.

stevenrombauts: Alright, makes sense too. Changed it!

PR #1641, opened 2025-12-15
Transparency log: track password reset/change

stevenrombauts:
Creates audit record when user resets password and changes it.


Seldaek (approved):
Besides the small cleanup I commented on looks good to merge IMO

PR #1642, opened 2025-12-16
Transparency log filters

IgorBenko:


src/QueryFilter/AuditLog/UserFilter.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $qb->setParameter($paramName, $pattern);
+
+        $qb->innerJoin(User::class, 'targetUser', 'WITH', 'a.userId = targetUser.id');
+
+        if ($useWildcard) {
+            $qb->andWhere('targetUser.username LIKE :' . $paramName);
+        } else {
+            $qb->andWhere('targetUser.username = :' . $paramName);
+        }
```
Seldaek: Why not using `attributes->>$.user.username` here to filter? I think it'd be best because the user might not exist anymore in the users table.

PR #1648, opened 2025-12-17
Transparency log: add auditor role

stevenrombauts:
Adds a new auditor role, that can be used to provide a specific user with more context in the transparency log.


tests/Audit/Display/AuditLogDisplayFactoryTest.php
```diff
@@ -343,16 +343,18 @@ public function testBuildPackageUnabandonedWithoutPreviousReplacement(): void
         self::assertSame('maintainer', $display->actor->username);
     }
 
-    #[TestWith([false, 999, '**@**.**'])]
-    #[TestWith([true, 999, 'john@doe.com'])]
-    #[TestWith([false, 999, '**@**.**'])]
-    #[TestWith([false, 123, 'john@doe.com'])]
-    public function testBuildUserVerified(bool $isAdmin, int $authenticatedUserId, string $expectedEmail): void
+    #[TestWith([null, 999, '**@**.**'])]
+    #[TestWith(['ROLE_ADMIN', 999, 'john@doe.com'])]
```
Seldaek: I don't think we need to test this. This tests the security config, not the implementation of AuditLogDisplayFactory

Issue #1673, opened 2026-02-09
Error in dependents page


jawira:
Hello, I found a "400 Bad Request" error in Packagist website.

How to reproduce:

1. Go to the dependents page of any package, for example <https://packagist.org/packages/jawira/plantuml/dependents>.
2. Then click in "downloads" link.
3. The error raises.

Screenshots:

<img width="2049" height="774" alt="Image" src="https://github.com/user-attachments/assets/34a6fab2-960c-431b-ae89-6757c4859a57" />

<img width="1570" height="396" alt="Image" src="https://github.com/user-attachments/assets/c399fbeb-c635-497c-9309-b8f846bcde2e" />

PR #1674, opened 2026-02-09
Fix the package dependent list

stof:
The `$requires` variable name was used for 2 different variables, breaking the value passed to the template.

Fixes https://github.com/composer/packagist/issues/1673


Seldaek:
Oh.. sorry fixed in a3f87e06cfb8b056f3553da6dbcea0d6e20b6d4d as someone else reported it to me before I saw this 🙈

PR #1678, opened 2026-02-17
Avoid crashing other JS of the page if algolia is not configured

eiriksm:
Howdy :cowboy_hat_face: 

I was setting up the project locally, and noticed that if I did not explicitly configure algolia, I would not get all of the JS of the app executed, as it would crash upon importing the search.js file.

This makes (among other things, this is just an example) the toggling of the user tokens not work, if algolia is not configured.

Realistically, this is not a problem when deploying the actual packagist.org website. But it makes for a somewhat potential frustrating, and maybe "frictionful", developer experience when attempting to start the project locally :v: 

In this PR we try catch the instantiating of the algolia search, and as a fallback configure no-ops for the methods that are about to be called in the lines following this code path. Personally I would prefer it to be possible to return early if we do not have the algolia setup, but this would make for a much "worse" diff to review, as one cannot simply return from the top level scope in an import like that :sweat_smile:

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -39,7 +39,7 @@ if (decodeURI(location.search).match(/[<>]/)) {
 }
 
 var searchThrottle = null;
-var search = instantsearch({
+var opts = {
     appId: algoliaConfig.app_id,
     apiKey: algoliaConfig.search_key,
     indexName: algoliaConfig.index_name,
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, makes sense

PR #1706, opened 2026-04-13
FilterLists: unified malware list with source property instead of multiple malware lists

glaubinix:
Changes here will further simplify the Composer malware filter feature

* Have one unified malware list where entries have a source, similar to advisories
* Make public ID not nullable
* Drop default-lists from to the packages.json

```sql
ALTER TABLE filter_list_entry CHANGE publicId publicId VARCHAR(255) NOT NULL;

ALTER TABLE filter_list_entry ADD source VARCHAR(255) DEFAULT NULL;
UPDATE filter_list_entry SET source = 'aikido', list = 'malware';
ALTER TABLE filter_list_entry CHANGE source source VARCHAR(255) NOT NULL;

UPDATE audit_log SET attributes = JSON_SET(attributes, '$.entry.source', 'aikido', '$.entry.list', 'malware') WHERE type IN ('filter_list_entry_added','filter_list_entry_deleted');
```


Seldaek:
Thanks!

PR #1711, opened 2026-04-22
GitHub Actions: use zizmor to harden existing workflows

glaubinix:
This introduces [zizmor](https://github.com/zizmorcore/zizmor) a static analyzer for GitHub Actions to harden existing GitHub Action workflows and adds a new workflow to analyze any future changes to those workflows.

I didn't introduce any new concurrency limitations. Happy to add that if wanted but don't think this is necessary.

### DIFF
--- a/.github/workflows/conductor.yaml
+++ b/.github/workflows/conductor.yaml
@@ -1,22 +1,29 @@
 # See the Conductor setup guide at https://packagist.com/docs/conductor/getting-started
 
+name: Private Packagist Conductor
+
 on:
   repository_dispatch:
     types:
       - dependency_update
 
-name: Private Packagist Conductor
+concurrency:
+    group: ${{ github.workflow }}
 
-permissions:
-  contents: write
+permissions: {}
 
 jobs:
   conductor:
     name: Private Packagist Conductor
     runs-on: ubuntu-24.04-arm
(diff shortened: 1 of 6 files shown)


.github/workflows/conductor.yaml
```diff
@@ -1,22 +1,26 @@
...
-  contents: write
+permissions: {}
 
 jobs:
   conductor:
     name: Private Packagist Conductor
     runs-on: ubuntu-24.04-arm
 
+    permissions:
+        contents: write # Allow Conductoor to push commits with composer.json/.lock changes
```
Seldaek:
```suggestion
        contents: write # Allow Conductor to push commits with composer.json/.lock changes
```

.github/workflows/conductor.yaml
```diff
@@ -1,22 +1,26 @@
 # See the Conductor setup guide at https://packagist.com/docs/conductor/getting-started
 
-on:
+on: # zizmor: ignore[concurrency-limits]
```
Seldaek: It'd be nice to fix those IMO rather than ignoring.

glaubinix: done

Seldaek:
Thanks

PR #1725, opened 2026-05-18
FilterLists: generate summary-url file and upload it to S3

glaubinix:
This implements the packagist.org part of https://github.com/composer/composer/pull/12833.

We will regenerate the file every time when we modify any entries or if the file on the CDN has not been updated yet e.g. if cache purge failed. The check runs as part of the background job worker


src/FilterList/Dump/FilterListDumperProvider.php
```diff
@@ -51,4 +51,26 @@ public function getEntriesForDump(array $packageNames): array
...
+        foreach ($this->getEM()->getRepository(FilterListEntry::class)->getAllSummaryEntries() as $entry) {
+            $grouped[$entry->list->value][$entry->packageName][] = $entry->version;
+        }
+
+        $byList = [];
+        foreach ($grouped as $listValue => $packages) {
+            ksort($packages);
+            foreach ($packages as $packageName => $versions) {
+                $unique = array_values(array_unique($versions));
+                sort($unique);
```
Seldaek: Should maybe use Semver::sort from here to sort things more semantically? It's just more readable for people looking at it. But that expects versions and will blow up on constraints, so maybe we should rather just `natsort()` that's good enough I guess.

```suggestion
                natsort($unique);
```

PR #1734, opened 2026-05-21
FilterLists: only import entries with a valid version constraint

glaubinix:
Just adding validation that we don't start to import invalid versions/constraints


Seldaek:
Thanks

PR #1741, opened 2026-05-27
fix: breaking `scss` design removed from

### LINKED ISSUE #1557: Border offset on login field

aihimel:
## Changes
Previously user authentication related forms were broken due to a previous design changes. Removed the `scss` design rule to fix the form to it's previous state.

## Related
* Closes #1557


Seldaek:
This sadly doesn't fix the issue everywhere, I then see a 1px missing everywhere in chrome. I think maybe you're looking at it in firefox?

image

Seldaek:
I don't think this is relevant anymore as we upgraded to bootstrap 5 etc.

PR #1742, opened 2026-05-27
Make stable versions immutable and allow soft-deletion / hiding of versions

Seldaek:
### Why

Today the Updater silently rewrites a stable version row whenever upstream changes the source/dist reference (re-tags, force-pushed annotated tags, etc.). For a published `v1.2.3` this is wrong — downstream lock files, mirrors, and security scanners all assume a tagged release is a frozen snapshot. And `VersionRepository::remove()` is the only way to pull a release: no recovery, no provenance, no way to distinguish maintainer cleanup from an admin takedown.

This PR locks stable versions, adds a reason-aware soft-delete, and a separate admin-only version hiding capability.

### Stable-version immutability gate (`Updater`)

- New rule: once a non-dev version row exists, the Updater never modifies its identity. Effective reference = `source.reference` if present, else `dist.reference`.
- `DELETE_BEFORE` now only wipes dev versions; stable rows fall through to the immutability gate.
- `UPDATE_EQUAL_REFS` renamed to `UPDATE_SOURCE_DIST_URL`. For stable rows, the URL rewrite only fires when both old and new refs are identical 40+ char hashes AND a driver-confirmed dist URL fetch verifies the new URL. Otherwise the version is skipped with a specific reason.

### Soft-delete with reasons (new `VersionDeletionReason` enum)

- `AutoDeletedMissing` — Updater soft-marks a row when upstream loses it; hard-purges after 1 day for dev versions only. Stable versions stay forever.
- `DeletedByMaintainer` — pulled via the per-version UI delete.
- `DeletedByAdmin` — pulled by an admin with an optional public reason.
- `Hidden` — admin took the version down with no public trace (used by the spam-cascade and selective takedowns). Visible to maintainers/admins, hidden from everyone else.

### V2 metadata + UI

- `V2Dumper` filters soft-deleted versions out of both `<name>.json` and `<name>~dev.json`. Composer can no longer resolve a pulled version.
- Package page lists soft-deleted versions grayed-out with a per-reason title (`Deleted by maintainer on ...`, `Removed by admin on ...: <reason>`, `Hidden by admin on ...: <reason>`, `No longer found in upstream repository`).
- New "retag blocked" warning badge on stable rows where `lastBlockedReference IS NOT NULL`, linking to the new doc page.
- Soft-deleted rows show a Recover button (maintainer-recoverable for maintainer/auto-missing reasons; admin-only recovery for versions removed or hidden by admins).

### Audit + notification

- New audit record types: `VersionSoftDeleted`, `VersionRecovered`, `VersionReferenceChangeBlocked`.
- If a version ref change is detected, we email all maintainers with an update failure notification.
- New static doc page at `/about/version-immutability` explaining the rule, what the maintainer sees, and the recommended remediation.

### Out of scope

- Malware-flagged versions: unchanged — they keep appearing in metadata, surfaced via the existing FilterListEntry UI.
- Whole-package deletion (`PackageManager::deletePackage` / `CleanSpamPackagesCommand`) is a separate concept (whole row cascades) and untouched.


src/Entity/Version.php
```diff
@@ -678,6 +689,59 @@ public function getSoftDeletedAt(): ?\DateTimeImmutable
...
+    public function getLastBlockedReference(): ?string
+    {
+        return $this->lastBlockedReference;
+    }
+
+    /**
+     * Effective reference used for immutability identity: source.reference if present,
+     * else dist.reference, else null (no usable identity).
+     */
+    public function getEffectiveReference(): ?string
```
glaubinix: This looks unused

migrations/2026_05_version_immutability.sql
```diff
@@ -0,0 +1,20 @@
+-- Version immutability + soft-delete reasons
+--
+-- Adds:
+--   deletionReason         enum-string column (auto_missing | maintainer | admin)
```
glaubinix:
```suggestion
--   deletionReason         enum-string column (auto_missing | maintainer | admin | hidden)
```

src/Controller/PackageController.php
```diff
@@ -804,7 +825,7 @@ public function viewPackageVersionAction(Request $req, int $versionId): JsonResp
     }
```
glaubinix, line 845: `PackageController::viewPackageVersionAction` still returns all softDeleted versions for json requests. Not sure we care really but think would be nice to either remove them or at least flag them as soft deleted?

src/Controller/UserController.php
```diff
@@ -123,7 +124,10 @@ public function markSpammerAction(Request $req, #[VarName('name')] User $user):
 
             foreach ($packages as $package) {
                 foreach ($package->getVersions() as $version) {
-                    $versionRepo->remove($version);
+                    if ($version->isSoftDeleted()) {
+                        continue;
+                    }
+                    $versionRepo->softDelete($version, VersionDeletionReason::Hidden, 'spam', $adminUser);
```
glaubinix, line 130: This will trigger one package update per version

src/Package/Updater.php
```diff
@@ -87,7 +97,12 @@ class Updater
 {
     use \App\Util\DoctrineTrait;
 
-    public const UPDATE_EQUAL_REFS = 1;
+    /**
+     * Roll out new source/dist URLs across existing versions without changing references.
+     * Stable versions only accept the rewrite when every check in updateSourceDistUrl() passes
```
glaubinix: I cannot find `updateSourceDistUrl()` anywhere

src/Package/Updater.php
```diff
@@ -292,39 +344,65 @@ public function update(IOInterface $io, Config $config, Package $package, VcsRep
...
+        // auto-recover versions still present upstream that had been auto-soft-deleted as missing.
+        // Admin/maintainer-pulled rows (deletionReason in (admin, maintainer)) stay soft-deleted.
         $em->getConnection()->executeStatement(
-            'UPDATE package_version SET updatedAt = :now, softDeletedAt = NULL WHERE id IN (:ids) AND softDeletedAt IS NOT NULL',
-            ['now' => date('Y-m-d H:i:s'), 'ids' => $idsToMarkUpdated],
+            'UPDATE package_version
+                SET updatedAt = :now, softDeletedAt = NULL, deletionReason = NULL, deletionReasonText = NULL
+                WHERE id IN (:ids)
+                  AND softDeletedAt IS NOT NULL
+                  AND (deletionReason IS NULL OR deletionReason = :autoReason)',
```
glaubinix, line 389: Just curious when would deletionReason by NULL here? Is this for versions soft deleted before merging this PR? Wonder if we should mark them as AutoDeletedMissing via migration? Would take it there aren't that many versions flagged as softDelete right now

src/Package/Updater.php
```diff
@@ -292,39 +344,65 @@ public function update(IOInterface $io, Config $config, Package $package, VcsRep
...
             if (
-                // soft-deleted versions are really purged after a day
-                (null !== $version['softDeletedAt'] && new \DateTime($version['softDeletedAt']) < $deleteDate)
-                // remove v1 normalized versions of dev-master/trunk/default immediately as they have been recreated as dev-master/trunk/default in a non-normalized way
-                || ($version['normalizedVersion'] === '9999999-dev')
+                $isDev
+                && (
+                    (null !== $version['softDeletedAt'] && new \DateTime($version['softDeletedAt']) < $deleteDate)
+                    || ($version['normalizedVersion'] === '9999999-dev')
+                )
```
glaubinix, line 413: I think this looks off now. Isn't the `development` flag only a Composer 2 feature? So any Composer 1 `9999999-dev` version would already be skipped by the first part of the if statement.

glaubinix (commented):
Not fully through yet. Still missing the second half of the updater changes + tests and JS code

src/Package/Updater.php
```diff
@@ -381,35 +461,210 @@ public function update(IOInterface $io, Config $config, Package $package, VcsRep
...
+        $oldSource = $existingVersion['source'] ?? null;
+        $oldDist = $existingVersion['dist'] ?? null;
+        $oldSourceRef = $oldSource['reference'] ?? null;
+        $oldDistRef = $oldDist['reference'] ?? null;
+        $newSourceRef = $data->getSourceReference();
+        $newDistRef = $data->getDistReference();
+        $newSourceUrl = $data->getSourceUrl();
+        $newDistUrl = $data->getDistUrl();
+
+        if (!\is_string($oldSourceRef) || $oldSourceRef === '' || !\is_string($oldDistRef) || $oldDistRef === '') {
```
glaubinix, line 553: All the checks here basically mean we will never rewrite source/dist for any self-hosted GitLab/Gitea/Forgejo instances or other git servers even if their URL changes. I think this is fine but just wanted to confirm

src/Package/Updater.php
Seldaek: Yeah I think this became applySourceDistUrlRewrite() - will fix

src/Controller/UserController.php
Seldaek: That's not too bad, scheduleJob() deduplicates and if a pending job already exists.

src/Package/Updater.php
Seldaek: The migration file does set null ones to autodeleted missing actually, so once that ran through I guess we could simplify this condition yes. 

And yeah there are 20K records with soft delete set right now, because some packages just never get updated for various reasons and then old branches are just left there dangling.

src/Package/Updater.php
Seldaek: Yup that is correct, it's a limitation that I can live with for now, we can always change that later once we understand it more, but ideally we'll not have external dist URLs forever so that'd resolve it.

src/Controller/PackageController.php
Seldaek: Fixed in 45ac0e47

src/Package/Updater.php
Seldaek: the development flag there is coming from the packagist db, that's just set true for any dev-*/*-dev version. I agree in theory we shouldn't need the BC path for old v1 versions, but sadly we still have 10000 rows with old `9999999-dev` as normalized version, these should be recreated/deleted, but these are most likely packages which are gone / not updateable anymore, but if for some reason they update I'd rather the code keeps working as intended so leaving this here. Clarified a little bit in 3613bc10 tho

src/Entity/Version.php
Seldaek: Yup, duplicate of Updater::computeEffectiveReference 584b9185

Seldaek:
Ok this is ready to go I think, will deploy it tomorrow morning.

PR #1756, opened 2026-06-09
Pin GitHub webhook signature algorithm to sha256

iliaal:
The `/api/github` handler split the `X-Hub-Signature-256` header on `=` and passed the left side to `hash_hmac` as the algorithm name, so a header like `bogus=...` raised an uncaught `ValueError` and returned 500 instead of rejecting the signature. This pins the algorithm to the sha256 scheme GitHub sends to that header and requires a well-formed hex digest, so malformed headers fall through to the existing 403. Adds a regression test.


Seldaek:
Thanks, makes sense

PR #1757, opened 2026-06-09
Escape GitHub API error text shown on the profile page

iliaal:
`isAcceptableException` returns the upstream GitHub error message for a 403, and `my_profile.html.twig` renders the failure reason with `|raw` because the 404 branch returns trusted markup. This escapes the upstream-controlled text so it can't inject markup into the page.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -62,7 +62,7 @@ parameters:
         SecurityAdvisoryJob: 'array{source: string}'
         FilterListJob: 'array{list: string, source: string}'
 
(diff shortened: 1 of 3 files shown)


stof:
A better solution might be to return `\Twig\Markup|string|false` (or maybe `null` instead of `false` ?), where the known trusted markup is wrapped in a `Markup` object. This way, the template can remove the `raw` filter.

iliaal:
Good call on separating the trusted markup from the upstream text. `\Twig\Markup` doesn't survive this path though. The failure list is returned from the async job and stored in `Job::$result`, a Doctrine `json` column. `Markup` is `JsonSerializable` so it encodes to its string fine, but on decode it comes back as a plain string. The type is gone by the time the profile page renders the persisted job, so dropping `|raw` would escape the trusted 404 link and show it as literal markup.

Went with an explicit `reasonIsHtml` flag on the failure entry instead. It survives the JSON round-trip, and `|raw` now only ever applies to worker-controlled constant strings. The upstream 403 message returns as plain text and Twig auto-escapes it.

Seldaek:
Makes sense, thanks!

PR #1758, opened 2026-06-09
Harden internal metadata write path check and signature

iliaal:
Two fixes to the internal metadata write endpoint, both only reachable with the internal secret. The path was matched with `p2/.*\.json` where `.` also matches `/`, so `..` segments passed and could resolve outside the metadata dir; now restricted to `p2/vendor/package.json` segments with a `..` reject. The HMAC signed path, contents and filemtime with no delimiter, so different boundaries produced the same message; now length-prefixed on both sender and receiver. Signature format changes, so primary and replicas must deploy together (writes are idempotent, `replicaIps` empty by default).

### DIFF
--- a/src/Controller/InternalController.php
+++ b/src/Controller/InternalController.php
@@ -41,13 +41,13 @@ class InternalController extends Controller
         $filemtime = $req->request->getInt('filemtime');
         $sig = (string) $req->headers->get('Internal-Signature');
 
-        $expectedSig = hash_hmac('sha256', $path.$contents.$filemtime, $this->internalSecret);
+        $expectedSig = hash_hmac('sha256', \strlen($path).':'.$path.':'.\strlen($contents).':'.$contents.':'.$filemtime, $this->internalSecret);
         if (!hash_equals($expectedSig, $sig)) {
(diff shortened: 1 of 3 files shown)


Seldaek:
> now length-prefixed on both sender and receiver

Just curious what's the point of doing this? Mitigating length extension attacks?

I mean it seems totally fine so I'll merge but just trying to understand.

iliaal:
Not length extension. HMAC is immune to that by construction, so the old `$path.$contents.$filemtime` form was fine on that front.

It's concatenation ambiguity. HMAC signs the exact byte string you hand it, and `$path.$contents.$filemtime` isn't an injective encoding of the three fields. The boundaries are invisible to the MAC, so different `(path, contents, filemtime)` triples can collapse to the same bytes. Toy version: `("ab", "c", t)` and `("a", "bc", t)` both sign `abc<t>`, so one signature authenticates both splits.

Length-prefixing (`len:path:len:contents:filemtime`) makes the serialization unambiguous, binding each signature to exactly one triple. It's standard MAC canonicalization hygiene, not a patch for a live exploit here: the `..` reject and the tighter `p2/[^/]+/[^/]+\.json` regex already constrain where a shifted split could resolve.

PR #1759, opened 2026-06-09
Validate URL schemes on package metadata link fields

iliaal:
A package's homepage, the support source/issues/chat/forum/wiki/docs/security links, funding URLs and author homepages come from its composer.json and are rendered into href attributes, but were stored without a scheme check while the readme link sanitizer already restricts schemes via `allowLinkSchemes`. This drops any of those fields whose scheme is not http or https at store time, so the stored metadata and the dumped v2 JSON only carry web links.

### DIFF
--- a/src/Package/Updater.php
+++ b/src/Package/Updater.php
@@ -737,7 +737,7 @@ class Updater
             }
         }
 
-        $version->setHomepage($data->getHomepage());
+        $version->setHomepage($this->filterUrl($data->getHomepage()));
         $version->setLicense($data->getLicense() ?: []);
 
         $version->setPackage($package);
@@ -783,8 +783,8 @@ class Updater
         $version->setExtra($data->getExtra());
         $version->setBinaries($data->getBinaries());
         $version->setIncludePaths($data->getIncludePaths());
-        $version->setSupport($data->getSupport());
-        $version->setFunding($data->getFunding());
+        $version->setSupport($this->filterSupportUrls($data->getSupport()));
+        $version->setFunding($this->filterFundingUrls($data->getFunding()));
 
         if ($data->getKeywords()) {
             $keywords = [];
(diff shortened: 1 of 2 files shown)


src/Package/Updater.php
```diff
@@ -1164,6 +1167,61 @@ private function sanitize(?string $str): ?string
...
+     *
+     * @return array<string, string>|null
+     */
+    private function filterSupportUrls(?array $support): ?array
+    {
+        if (null === $support) {
+            return null;
+        }
+
+        foreach (['issues', 'forum', 'wiki', 'source', 'docs', 'rss', 'chat', 'security'] as $key) {
```
Seldaek: I'd argue for `chat` we should allow irc/ircs schemes too, maybe make the list of allowed schemes an optional arg in filterUrl?

Seldaek (approved):
Aside from the IRC nitpick, the rest looks good to go for me.

src/Package/Updater.php
iliaal: Done in cd695d71. `filterUrl` now takes an optional `$allowedSchemes` (default `['http','https']`), and `chat` passes `['http','https','irc','ircs']`. I kept `chat` out of the shared http/https loop rather than widening the loop's list, so irc/ircs stay scoped to `chat` and don't leak into source/issues/etc. Added an `irc://` chat case to the test.

Left as-is: the separate `support.irc` field isn't filtered, since it isn't rendered into an href in the templates (only source/issues/chat/forum/wiki/docs are). Can validate it to irc/ircs too if you'd want that for the dumped metadata.

Seldaek: Ah totally forgot we had support.irc too :D Probably would filter it too for completeness, who knows where this stuff ends up being rendered ultimately..

iliaal: Done in c93ce2e7. `irc` now goes through the same `['http','https','irc','ircs']` check as `chat` (grouped them into one loop), so dangerous schemes get dropped there too. Test covers a `javascript:` `irc` value being stripped.

Seldaek:
Alrighty, thanks!

PR #1762, opened 2026-06-11
Add tooltips to version list

stevenrombauts:
It's not clear why a version could be grayed out and crossed out now, so this PR adds a tooltip to display the reason. 

Since I was playing around with that file now anyway, I also added tooltips to the other buttons in there, and changed the styling for every version row to automatically align all icons to the right in the same way. (Claude Code did that, actually). Only suggestions, let me know if you want to drop/change anything.

CleanShot 2026-06-11 at 11 56 55


Seldaek:
Thanks

PR #1763, opened 2026-06-12
Fix handling of withdrawn advisories

Seldaek:
If one was already stored but then an update tries to withdraw it + add another one for the same CVE, doctrines flush() fails as it tries to insert the new one first


Seldaek:
Merging cause we need this fixed but would appreciate your review anyway when you're back @glaubinix for safety :)

glaubinix (commented):
Very interesting. Still not quite sure what GitHub's thinking here is how an advisory can be live and withdrawn at the same time but code looks good.

PR #1766, opened 2026-06-15
Add links to audit log from package/vendor/user pages

Seldaek:


templates/user/public_profile.html.twig
```diff
@@ -17,6 +17,7 @@
             [SPAMMER]
         {% endif %}
         {{ 'user.member_since'|trans }}: {{ user.createdAt|date('M d, Y') }}
+        &middot; <a rel="nofollow noindex" href="{{ path('view_audit_logs', {user: user.username}) }}">Audit log</a>
```
naderman: Thought we're calling it transparency log in the UI at least now, even though we didn't yet figure out immutability?

Seldaek: Yeah I thought we can link it as audit log because it's shorter and reflects reality more, and more understandable IMO by everyone what to expect. "Transparency log" has been confusing people in my anectodical experience. But not feeling super strong either way here..

Issue #1754, opened 2026-06-08
Transparency Log Improvements


naderman:
A few improvements I'd like to see on the transparency log:

- [ ] Ability to deep link to a specific entry to easily share it with someone else
- [ ] version created/changed events shouldn't just show the commit hash but also the date, so you can see a version created in 2026 is actually from a git tag created with a (potentially fake) date a year ago.
- [ ] package pages should link to the transparency log filtered by their package name to simplify looking this up. => #1766

shyim:
and maybe performance 🙈 
always 504
https://packagist.org/transparency-log?actor=&user=store.shopware.com

naderman:
@shyim see https://github.com/composer/packagist/pull/1768

PR #1768, opened 2026-06-16
Transparency log optimizations and fix searching of filter list entries

Seldaek:
See individual commits for details


glaubinix:
Not reviewed yet but can you also add an index on user/actor to help with queries like: https://github.com/composer/packagist/issues/1754#issuecomment-4728587095

glaubinix (approved):
Looks good!

PR #1772, opened 2026-06-18
Add internal/public deletion reason for the package delete + an internal reason for the version delete

Seldaek:


src/Entity/Version.php
```diff
@@ -228,6 +228,9 @@ class Version
     #[ORM\Column(type: 'text', length: 65535, nullable: true)]
     private ?string $deletionReasonText = null;
 
+    #[ORM\Column(type: 'text', length: 65535, nullable: true)]
+    private ?string $internalDeletionReasonText = null;
```
glaubinix, line 232: This is now stored but we don't actually display it anywhere (except the transparency log where we store it separately as part of the audit record). Did you mean to display the internal text here next to the regular text for admins on the version list page tooltip or do we store this just in case we might need it later?

glaubinix (approved):
Looks good and works as expected

src/Entity/Version.php
Seldaek: I'd say we probably might want to show it, and doing so without digging into the audit log is easier.

glaubinix: Would definitely make it easier to see what is going on.

templates/audit_log/display/package_deleted.html.twig
```diff
@@ -2,4 +2,4 @@
 
 <strong>{{ auditLog.packageLink(display.packageName) }}</strong><br>
 Repository: {{ display.repository }}<br>
-Deleted by: {{ display.actor.username }}
+Deleted by: {{ display.actor.username }}{% if display.reason %}<br>Reason: {{ display.reason }}{% endif %}{% if display.internalReason %}<br>Internal reason: {{ display.internalReason }}{% endif %}
```
stof, line 5: shouldn't the internal reason be displayed only for users with the auditor role, as documented in the comment in the DB migration ?

glaubinix: This is done in `AuditLogDisplayFactory::internalReason`

PR #1775, opened 2026-06-24
Add user/package freezing capability

Seldaek:
### Why
- The enabled flag actually means "email confirmed" — it's flipped on by email verification, password reset, and GitHub login. Reusing it to disable accounts was broken: a disabled user could re-enable themselves just by doing a password reset.
- "Mark as spammer" worked by granting a fake ROLE_SPAMMER, overloading the role system with moderation state.

### What this adds

- A first-class account freeze concept, separate from email confirmation, carrying a reason (Spam, Bad actor, Temporary); null when not frozen.
- A frozen account cannot log in, request or perform a password reset, authenticate via API token, or be added as a package maintainer. Only an admin can unfreeze, and freezing immediately logs the user out.
- An admin-only Freeze / Unfreeze control on the user profile, gated behind a new ROLE_DISABLE_USERS permission, capturing an optional public reason plus an internal (admin/auditor-only) note.
- Freeze/unfreeze are recorded in the transparency log ("User frozen" / "User unfrozen"), with the internal note shown only to auditors.
- ROLE_SPAMMER is retired. "Mark as spammer" now applies a spam freeze and is audit-logged; existing spammers are migrated to a spam freeze and the role is stripped.
- Background worker that processes the package/version freezing so we don't delay user freeze responses

### Behavior changes worth calling out
- A password reset no longer reactivates a disabled account (fixes the self-reactivation hole).
- Marking a user as a spammer is now recorded in the audit log (it wasn't before).


src/Controller/UserController.php
```diff
@@ -146,6 +148,73 @@ public function markSpammerAction(Request $req, #[VarName('name')] User $user, #
...
+        if ($user->getId() === $adminUser->getId()) {
+            $this->addFlash('error', 'You cannot freeze your own account.');
+
+            return $this->redirectToRoute('user_profile', ['name' => $user->getUsername()]);
+        }
+
+        $form = $this->createFormBuilder([])->getForm();
+        $form->submit($req->request->all('form'));
+        if ($form->isSubmitted() && $form->isValid()) {
+            $reason = UserFreezeReason::tryFrom($req->request->getString('reason'));
```
stof: why reading it from the request instead of relying on the form data ?

src/Controller/UserController.php
```diff
@@ -146,6 +148,73 @@ public function markSpammerAction(Request $req, #[VarName('name')] User $user, #
...
+            if ($reason === null) {
+                $this->addFlash('error', 'Invalid freeze reason.');
+
+                return $this->redirectToRoute('user_profile', ['name' => $user->getUsername()]);
+            }
+
+            $reasonText = trim($req->request->getString('reasonText')) ?: null;
```
stof: this trimming and turning empty into null would already be handled when using the data processed by the form component

src/Controller/UserController.php
stof: OK, apparently, the form is managing only the CSRF field, not the full submission. Any reason for that ?

Seldaek: The new commit addresses this (and a bunch more). No particular reason, thanks for pointing it out.

src/Controller/UserController.php
```diff
@@ -89,61 +84,141 @@ public function triggerGitHubSyncAction(#[CurrentUser] User $user): RedirectResp
...
+    {
+        $em = $this->getEM();
+
+        // Suppressed reasons (spam/malware) also drop the package from the search index immediately.
+        if ($reason->suppressesPackage()) {
+            $sql = 'UPDATE package p JOIN maintainers_packages mp ON mp.package_id = p.id SET frozen = :reason, indexedAt = NULL WHERE mp.user_id = :userId';
+        } else {
+            $sql = 'UPDATE package p JOIN maintainers_packages mp ON mp.package_id = p.id SET frozen = :reason WHERE mp.user_id = :userId';
+        }
+        $em->getConnection()->executeStatement($sql, ['reason' => $reason->value, 'userId' => $user->getId()]);
```
glaubinix: This is missing PackageFrozen audit records

src/Controller/UserController.php
```diff
@@ -89,61 +84,141 @@ public function triggerGitHubSyncAction(#[CurrentUser] User $user): RedirectResp
...
+    {
+        if ($this->isGranted('ROLE_DISABLE_PACKAGES')) {
+            return [PackageFreezeReason::Spam, PackageFreezeReason::Malware, PackageFreezeReason::Temporary];
+        }
+        if ($this->isGranted('ROLE_ANTISPAM')) {
+            return [PackageFreezeReason::Spam];
+        }
+
+        return [];
+    }
```
glaubinix: This is duplicate in the ProfileController. Wonder if this should become something like `PackageFreezeReason::allowedPackageFreezeReasons(bool $hasDisabledPackagesRole, bool $hasAntispamRole)` to avoid drifting apart

src/Controller/ProfileController.php
```diff
@@ -127,8 +133,25 @@ public function publicProfile(Request $req, #[VarName('name')] User $user, Favor
...
+                $packageReasons = [PackageFreezeReason::Spam];
+            }
+
+            $data['freezeForm'] = $this->createForm(FreezeType::class, new FreezeRequest(), [
+                'account_reasons' => $accountReasons,
+                'package_reasons' => $packageReasons,
+            ])->createView();
+            $data['unfreezeForm'] = $this->createForm(UnfreezeType::class, new UnfreezeRequest())->createView();
+            $data['canUnfreezeUser'] = $this->isGranted('ROLE_DISABLE_USERS')
+                || $user->getFreezeReason() === UserFreezeReason::Spam;
```
glaubinix:
```suggestion
                || $user->getFreezeReason()->isSpam();
```
If you already have the method?

src/Controller/UserController.php
```diff
@@ -89,61 +84,141 @@ public function triggerGitHubSyncAction(#[CurrentUser] User $user): RedirectResp
...
+        return $this->redirectToRoute('user_profile', ['name' => $user->getUsername()]);
+    }
+
+    /**
+     * @return list<UserFreezeReason>
+     */
+    private function allowedAccountFreezeReasons(): array
+    {
+        return $this->isGranted('ROLE_DISABLE_USERS') ? UserFreezeReason::cases() : [UserFreezeReason::Spam];
+    }
```
glaubinix: Same for this part here. Seems logic is duplicate here and in the ProfileController

src/Controller/UserController.php
```diff
@@ -89,61 +84,141 @@ public function triggerGitHubSyncAction(#[CurrentUser] User $user): RedirectResp
...
+
+        return $this->redirectToRoute('user_profile', ['name' => $user->getUsername()]);
+    }
+
+    #[Route(path: '/users/{name}/unfreeze', name: 'unfreeze_user', methods: ['POST'])]
+    public function unfreezeUserAction(Request $req, #[VarName('name')] User $user, #[CurrentUser] User $adminUser): RedirectResponse
+    {
+        // ROLE_DISABLE_USERS may unfreeze any reason; anti-spam moderators only spam freezes.
+        $mayUnfreeze = $this->isGranted('ROLE_DISABLE_USERS')
+            || ($this->isGranted('ROLE_ANTISPAM') && $user->getFreezeReason() === UserFreezeReason::Spam);
```
glaubinix: This here is also reusing the same logic about which ROLE can use which reason. Could be a simple `$mayUnfreeze = in_array($user->getFreezeReason(), UserFreezeReason::casesForRolesOrSomething())` or even `$user->getFreezeReason()->mayUnfreeze($role1, $role2)`

Then the comment can go where it's defined

src/Controller/UserController.php
```diff
@@ -89,61 +84,141 @@ public function triggerGitHubSyncAction(#[CurrentUser] User $user): RedirectResp
...
 
-    #[Route(path: '/spammers/{name}/', name: 'mark_spammer', methods: ['POST'])]
-    public function markSpammerAction(Request $req, #[VarName('name')] User $user, #[CurrentUser] User $adminUser): RedirectResponse
+    #[Route(path: '/users/{name}/freeze', name: 'freeze_user', methods: ['POST'])]
+    public function freezeUserAction(Request $req, #[VarName('name')] User $user, #[CurrentUser] User $adminUser): RedirectResponse
     {
-        if (!$this->isGranted('ROLE_ANTISPAM')) {
-            throw $this->createAccessDeniedException('This user can not mark others as spammers');
+        // Anti-spam moderators may freeze accounts as spam; ROLE_DISABLE_USERS may use any reason.
+        if (!$this->isGranted('ROLE_ANTISPAM') && !$this->isGranted('ROLE_DISABLE_USERS')) {
```
glaubinix: I see this quite a lot. Wonder if there should be a base role `ROLE_FREEZE_USERS`?

src/Controller/UserController.php
```diff
@@ -89,61 +84,141 @@ public function triggerGitHubSyncAction(#[CurrentUser] User $user): RedirectResp
...
+        ]);
+        $form->handleRequest($req);
 
-        $form->submit($req->request->all('form'));
         if ($form->isSubmitted() && $form->isValid()) {
-            $user->addRole('ROLE_SPAMMER');
-            $user->setEnabled(false);
+            $reason = $freezeRequest->reason;
+            if ($reason === null) {
+                return $this->redirectToRoute('user_profile', ['name' => $user->getUsername()]);
```
glaubinix: not quite sure how reason can be null but should this maybe add an error flash that nothing was changed?

tests/Entity/UserTest.php
```diff
@@ -0,0 +1,97 @@
...
+        self::assertFalse($a->isEqualTo($b), 'freezing must invalidate the session token');
+    }
+
+    private function makeUser(): User
+    {
+        $user = new User();
+        $user->setUsername('sameuser');
+        $user->setEmail('same@example.com');
+        $user->setPassword('samehash');
+        $user->setEnabled(true);
```
glaubinix: could also use the Fixture trait to limit duplication a bit

src/Controller/UserController.php
Seldaek: I think the goal is to get rid of ROLE_ANTISPAM because tbh if you can mark users/packages as spammer you have anyway enough permissions to do as much damage as ROLE_DISABLE_USERS.. Was just trying to avoid doing too much in one go.

src/Service/PackagePurgeWorker.php
```diff
@@ -0,0 +1,88 @@
...
+        if ($package !== null) {
+            $actor = isset($payload['actorId'])
+                ? $this->doctrine->getRepository(User::class)->find($payload['actorId'])
+                : null;
+
+            // Record the actual freeze reason (spam/malware) as the takedown text; the caller froze
+            // the package before scheduling the purge. Fall back to 'spam' if it is somehow no longer
+            // frozen (e.g. unfrozen between scheduling and this retry).
+            $freezeReason = $package->getFreezeReason();
+            $reasonText = $freezeReason !== null ? $freezeReason->value : 'spam';
```
glaubinix: If for some reason and admin undos a freeze. Shouldn't we then instead skip the deletion?

src/Service/Scheduler.php
```diff
@@ -86,6 +86,25 @@ public function scheduleFilterList(FilterLists $list, FilterSources $source, int
...
     }
 
+    /**
+     * Schedules the out-of-band cleanup of a package — freeze, version soft-deletes, and purge of
+     * its published artifacts. Keyed by name so it stays correct even if the row is later removed;
+     * $actorId, when given, is recorded as the actor of the version soft-deletes.
+     *
+     * @return Job<PackagePurgeJob>
+     */
+    public function schedulePackagePurge(Package|int $packageOrId, string $packageName, ?int $actorId = null): Job
```
glaubinix:
```suggestion
    public function schedulePackagePurge(Package $package, ?int $actorId = null): Job
```
Could be simplified as only ever called with a Package

src/Entity/PackageRepository.php
```diff
@@ -100,7 +100,8 @@ public function getPackageNamesUpdatedSince(\DateTimeInterface $date): array
     public function getPackageNames(): array
     {
         $query = $this->getEntityManager()
-            ->createQuery("SELECT p.name FROM App\Entity\Package p WHERE p.frozen IS NULL OR p.frozen != 'spam'");
+            ->createQuery('SELECT p.name FROM App\Entity\Package p WHERE p.frozen IS NULL OR p.frozen NOT IN (:suppressed)')
```
glaubinix, line 103: Do you also need to adjust `getPackageNamesByTypeAndVendor` and `getPackageNamesBy*`?

migrations/2026_06_user_frozen.sql
```diff
@@ -0,0 +1,13 @@
+-- Admin-controlled account freeze, mirroring the package freeze (see src/Entity/UserFreezeReason.php).
```
glaubinix:
```suggestion
-- Admin-controlled account freeze, mirroring the package freeze (see src/Entity/User.php).
```

glaubinix: The enum doesn't have its own file

glaubinix (commented):
~~Unrelated to the changes here but `package_frozen.hmlt.twig` can be simplified to `{{ display.reason.translationKey|trans }}`~~

Edit: ignore, translations don't match

src/Service/PackagePurgeWorker.php
Seldaek: Yeah good point.

migrations/2026_06_user_frozen.sql
stof: Not having its own file means it cannot be autoloaded. This might be an issue if this enum is referenced in places that don't reference the User class first. It might be better to make it autoloadable.

glaubinix (approved):

PR #1780, opened 2026-07-01
Record security advisory changes in the transparency log

glaubinix:
Add transparency log entries for the security advisory lifecycle: created, updated, withdrawn and reinstated. The update worker emits the records after resolving each run, classifying by the net state change so a transient withdraw/reinstate within one run collapses into a single edit. Advisory changes are attributed to the automation actor.

Screenshot 2026-07-03 at 11 41 09


src/Entity/AuditRecord.php
```diff
@@ -442,11 +340,41 @@ public static function filterListEntryDeleted(FilterListEntry $entry, ?User $act
...
-            vendor: self::getVendorFromPackage($entry->getPackageName()),
+        );
+    }
+
+    public static function securityAdvisoryCreated(SecurityAdvisory $advisory, ?User $actor): self
+    {
+        return new self(
+            AuditRecordType::SecurityAdvisoryCreated,
+            self::getSecurityAdvisoryData($advisory, $actor),
+            vendor: self::vendorFromPackageName($advisory->getPackageName()),
```
Seldaek, line 523: Not bad to link to vendor but ideally we'd also link this to the package id if the advisory packageName exists in our DB, right?

src/Entity/AuditRecord.php
```diff
@@ -463,6 +391,70 @@ private static function getUserData(?User $user, string $fallbackString = 'unkno
...
 
+    /**
+     * @return array{name: string, advisoryId: string, packageName: string, source: string, remoteId: string, cve: string|null, title: string, actor: array{id: int, username: string}|string}
+     */
+    private static function getSecurityAdvisoryData(SecurityAdvisory $advisory, ?User $actor): array
+    {
+        return [
+            'name' => $advisory->getPackageName(),
+            'advisoryId' => $advisory->getPackagistAdvisoryId(),
+            'packageName' => $advisory->getPackageName(),
```
Seldaek: I think I'd delete this to avoid the duplication.

src/Entity/AuditRecord.php
glaubinix: Great point! Done!

Seldaek:
Thanks

PR #1735, opened 2026-05-22
FilterLists: add admin UI to enable/disable/edit entries and add additional entries

glaubinix:
### This PR requires DB migrations
Before deploy
```
  ALTER TABLE filter_list_entry
      ADD COLUMN disabled TINYINT(1) NOT NULL DEFAULT 0,
      ADD COLUMN overwriteVersion VARCHAR(255) DEFAULT NULL,
     ADD COLUMN internalNote LONGTEXT DEFAULT NULL;

// Avoid UNIQUE INDEX list_package_version_source_idx (`list`, `packageName`, `version`, `source`);
ERROR 1071 (42000): Specified key was too long; max key length is 3072 bytes
ALTER TABLE filter_list_entry
      MODIFY `list` VARCHAR(32) NOT NULL,
      MODIFY `source` VARCHAR(32) NOT NULL,
      DROP INDEX list_package_version_idx,
      ADD UNIQUE INDEX list_package_version_source_idx (`list`, `packageName`, `version`, `source`);
```

After deploy (sets publicId on all audit log entries which makes it possible to display matching entries on the edit view)
```
UPDATE audit_log a
INNER JOIN filter_list_entry fle
ON fle.packageName = JSON_UNQUOTE(JSON_EXTRACT(a.attributes, '$.entry.package_name'))
AND fle.list       = JSON_UNQUOTE(JSON_EXTRACT(a.attributes, '$.entry.list'))
AND fle.version    = JSON_UNQUOTE(JSON_EXTRACT(a.attributes, '$.entry.version'))
SET a.attributes = JSON_SET(a.attributes, '$.entry.public_id', fle.publicId)
WHERE a.type IN ('filter_list_entry_added', 'filter_list_entry_deleted')
AND JSON_EXTRACT(a.attributes, '$.entry.public_id') IS NULL;
```

### Design decisions
* Creating and modifying an entry triggers a redump of the package. However, the summary-url only gets dumped again during the next worker run

### New filter list view

Shows all filter list entries with filters to better see what is going on. Disabled and overwritten entries are clearly marked in the UI.

Screenshot 2026-05-22 at 10 14 18

### New filter list edit view

Currently only allows you to modify the version constraint. All other properties are only shown for informational purposes.
The view shows at the bottom all audit log entries for the entry
 
Screenshot 2026-05-22 at 10 16 41

### Package filter list view changes

For admins there is now an edit link on the public page.

Screenshot 2026-05-22 at 10 17 30

### Audit/transparency log changes

[...]

PR #1782, opened 2026-07-03
Organizations groundwork

IgorBenko:
## Organizations: foundation & creation

This PR lays the groundwork for **Organizations** and initially allows creation of an organization by packagist admins (`ROLE_ORGANIZATIONS`). The write side is event-sourced from the start so that later stages (rename, soft-delete, teams, membership) can be added as new events without reworking the storage model. This is just a groundwork and underlying concepts might change in the follow-up PRs.

### Event-sourced write model

The canonical source of truth is an append-only event stream (`organization_event`).

- **`EventStore`** appends an aggregate's pending events and runs all projectors inside a single DB transaction (append → project → commit).  
- **Optimistic concurrency** is enforced by a unique `(aggregateId, sequence)` constraint; a conflict surfaces as `ConcurrencyException` for the caller to reload and retry.  
- **`Organization` aggregate** (`App\Organization\Domain`) is the write model. It records `OrganizationCreated` today and reconstitutes itself by replaying history. The read-side `App\Entity\Organization` is a separate projection.

### Read-model projections

Events are projected synchronously (via tagged `Projector` services) into query-friendly tables:

- **`organization`** — the read model used by the UI and lookups. Its unique slug index turns a concurrent duplicate into a rollback, which the application layer maps to a clean "slug already taken" error.  
- **`slug_reservation`** — blocks slugs freed by a rename or soft-delete. The projection exists now so creation already honours reservations, even though the events that populate it come in a later stage.  
- **`audit_log`** — organization creation is written to the public transparency log, with a new `organizationId` column.

### Slug & display-name validation

- **`Slug`** and **`DisplayName`** value objects guard format (lower-case alphanumerics \+ hyphens, max 20 chars for slugs; length limits for names).  
- **`OrganizationSlugClaimGuard`** checks the external facts a value object can't: the reserved-word deny-list, collisions with an existing vendor prefix the user doesn't control, and uniqueness against live orgs \+ active reservations.  
- Reserved words are also rejected as display names, not just slugs.

### Creating an organization

- New `/organizations`, `/organizations/create`, and `/organizations/{organization}` routes with list, create, and show pages, plus a menu entry.  
- **2FA is required** to create an organization (and thereby become an owner); users without it are redirected to configure it first.  
- Packagist admins see all organizations in the list; regular users see the ones they own.

### Access control

- **`OrganizationVoter`** governs org actions. Until membership management ships, the "owner" is the creating user. Admins may perform any owner action for moderation and are the only actor who can restore.  
- `OrganizationActions` enum, plus the `EditDisplayInfo` / `EditSlug` / `SoftDelete` / `Restore` action set, are defined now as groundwork; only creation is wired up so far.

### Database

The raw SQL provided in `migrations/2026_06_organizations.sql` must be executed, creating the `organization`, `organization_event`, and `slug_reservation` tables and adding `organizationId` to `audit_log`. Column names match the entity mappings so `doctrine:schema:create` produces the same schema.

Authors: @glaubinix and @igorbenko


src/Validator/NotReservedWord.php
```diff
@@ -17,6 +17,18 @@
...
+     * Names reserved across Packagist (usernames, organization slugs).
+     *
+     * @var list<string>
+     */
+    public const array RESERVED_WORDS = [
+        'composer',
+        'packagist',
+        'php',
+        'automation', // used to describe background workers doing things automatically in transparency log
+    ];
```
Seldaek: Sorry did a similar change in https://github.com/composer/packagist/commit/eb357d3fd0db46eba54656d1b5cbc8d9946da342 so this conflicts now

config/packages/security.yaml
```diff
@@ -76,8 +76,9 @@ security:
         ROLE_SPAMMER: ~
         ROLE_DISABLE_2FA: ~
         ROLE_AUDITOR: ~ # grants access to full audit records including PII (IP, emails, ..) as well as wildcard search in audit log
+        ROLE_ORGANIZATIONS: ~
```
Seldaek: Should this be ROLE_CREATE_ORGS or something more pinpointed? Or ROLE_ADMIN_ORGS if it applies to mroe than create.. But as is the name feels very unclear so would need at least a comment to explain what this grants.

glaubinix: Maybe we split this into two roles. One which allows you to ROLE_CREATE_ORGS and one that allows you to ROLE_VIEW_ORGS. This way we could let a user create an org and let some of their colleagues join to try it out?

src/Validator/NotReservedWord.php
glaubinix: Thanks, saw this already. Do you want us to rebase now or wait until you are done with the review?

Seldaek: As you wish.. Don't think it'll impact other files much

config/packages/security.yaml
Seldaek: ROLE_CREATE_ORGS is clear to me, but what does ROLE_VIEW_ORGS grant? If it's just about gating the whole org functionality I'd say maybe we skip that, and make them visible to everyone? If you don't have one you'll just see nothing on your org list. Or we do the role but then we'd have to add that role on any org member that gets added, and clean it up later once public.. Seems like more work than necessary somehow. It's not like this is a top secret feature, we just want to prevent people from creating them until we are ready, and this could be stated on the "My Organizations" page.

glaubinix: Yes, the idea was just about gating the feature. If this isn't necessary then happy to skip it

Seldaek: Yup let's keep things simple then I'd say :)

For creation IMO the best would be to have either a request-form (long term once this is closer to ready) and/or if the org creation form would be admin-only but have an owner(s) field so we can create an org for someone, instead of granting them that role and have them be able to create any number of orgs.

glaubinix: Right, the form as is, is mostly for us so we can interact with the feature will working on all the other details. Can discuss how the actual onboarding will work once we have team/permission setup all done

Seldaek: Looking a bit more at the roles set up in the PR, I think it'd be preferable to have a ROLE_ADMIN_ORGS and remove ROLE_CREATE_ORGS again.

The idea of roles right now is that if you have ROLE_ADMIN you have all sub-permissions, but every feature should ideally be gated by a custom smaller scope role name, so that we can eventually one day delegate some subsets of admin permissions to other people. So ROLE_ADMIN should never really be referenced directly in new code.

src/Twig/PackagistExtension.php
```diff
@@ -30,18 +31,21 @@ class PackagistExtension extends AbstractExtension
...
     private Security $security;
+    private OrganizationRepository $organizations;
 
-    public function __construct(ProviderManager $providerManager, RecaptchaHelper $recaptchaHelper, Security $security)
+    public function __construct(ProviderManager $providerManager, RecaptchaHelper $recaptchaHelper, OrganizationRepository $organizations, Security $security)
     {
         $this->providerManager = $providerManager;
         $this->recaptchaHelper = $recaptchaHelper;
         $this->security = $security;
+        $this->organizations = $organizations;
```
Seldaek: Would prefer this for consistency with other such variables

```suggestion
        $this->organizationRepo = $organizationRepo;
```

src/Organization/OrganizationSlugClaimGuard.php
```diff
@@ -0,0 +1,56 @@
...
+use App\Entity\PackageRepository;
+use App\Entity\SlugReservationRepository;
+use App\Entity\User;
+use App\Organization\Domain\Exception\InvalidSlugException;
+use App\Organization\Domain\Exception\SlugTakenException;
+use App\Organization\Domain\Slug;
+use App\Validator\NotReservedWord;
+
+/**
+ * TODO a Levenshtein-similarity check against a protected-names should be added later
```
Seldaek: We can discuss later, but I wonder if this is really necessary considering organization names aren't directly installable or anything, typo squatting is way less problematic as far as I can tell.

src/Organization/OrganizationSlugClaimGuard.php
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public function assertClaimable(Slug $slug, User $user): void
+    {
+        if (\in_array($slug->value, NotReservedWord::RESERVED_WORDS, true)) {
+            throw new InvalidSlugException(sprintf('"%s" is a reserved name and cannot be used.', $slug->value));
+        }
+
+        // A slug that matches an existing vendor prefix may only be claimed by someone
+        // who has access to that prefix (i.e. maintains a package under it).
+        if ($this->packages->isVendorTaken($slug->value, $user)) {
```
Seldaek: Also a discussion point here.. I've been wondering about that for a while now: how do we handle cases where multiple people currently have access to a vendor name, is it just a first-come first-served basis who gets to claim the org name? I mean as long as orgs are created by admins it's fine so we can define this later, but it should be on everyone's radar IMO that this might be a problem we need to have a clear policy for at least.

src/Organization/OrganizationManager.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->slugChecker->assertClaimable($slug, $owner);
+
+        if (\in_array(mb_strtolower($displayName->value), NotReservedWord::RESERVED_WORDS, true)) {
+            throw new InvalidDisplayNameException(sprintf('"%s" is a reserved name and cannot be used.', $displayName->value));
+        }
+
+        $organization = Organization::create(new Ulid(), $slug, $displayName);
+
+        try {
+            $this->eventStore->append($organization, Actor::owner($owner), $ip);
```
Seldaek, line 48: The actor should probably become the admin or at least the user actually creating the org here, which may be different than the owner.

src/Organization/OrganizationSlugClaimGuard.php
glaubinix: Right, this is one of the open questions. In how far do we have to protect certain names. I can remove the TODO here as the question is tracked elsewhere.

src/Organization/OrganizationSlugClaimGuard.php
glaubinix: Also already on our list of open questions but I will clarify that we also need to work out a policy of what happens if someone comes and would like to claim an existing organization name.

src/Validator/NotReservedWord.php
IgorBenko: Rebased and cleaned up the conflict.

src/Controller/OrganizationController.php
```diff
@@ -0,0 +1,96 @@
...
+        private readonly OrganizationManager $organizationManager,
+        private readonly OrganizationRepository $organizations,
+    ) {
+    }
+
+    #[Route(path: '/organizations', name: 'organization_list', methods: ['GET'])]
+    public function list(#[CurrentUser] User $user): Response
+    {
+        if ($this->isGranted('ROLE_ADMIN')) {
+            $organizations = $this->organizations->findAllOrdered();
```
Seldaek: This probably should move into a /admin/organizations section instead, but leave it for now.. We can do this once https://github.com/composer/packagist/pull/1735 is merged and build up the admin panel.

IgorBenko: OK, then at this point  /organizations will show only your own orgs and we move the full list to /admin/organizations

src/Organization/OrganizationSlugClaimGuard.php
IgorBenko: Yes, let's remove the TODO here for now

config/packages/security.yaml
IgorBenko: Replaced ROLE_ORGANIZATIONS with ROLE_ADMIN_ORGS now. Only admins can create orgs now and they can only see their own orgs.

migrations/2026_06_organizations.sql
```diff
@@ -0,0 +1,51 @@
...
+-- (organization, slug_reservation). Column names match the Doctrine entity mappings
+-- so a doctrine:schema:create produces the same schema.
+
+CREATE TABLE organization (
+    id BINARY(16) NOT NULL,
+    slug VARCHAR(20) NOT NULL,
+    displayName VARCHAR(60) NOT NULL,
+    status VARCHAR(16) NOT NULL,
+    createdAt DATETIME NOT NULL,
+    createdBy INT DEFAULT NULL,
```
Seldaek, line 12: Strictly speaking - in a follow up PR once we have proper membership roles I think this can be deleted.. Doesn't really matter who is the creator, especially if we have a full audit log for the org it'll be clear who that is from that if we need it. I realize for now you kinda need it tho :)

src/Organization/EventStore/ActorLabel.php
```diff
@@ -0,0 +1,24 @@
...
+namespace App\Organization\EventStore;
+
+/**
+ * Who/what triggered an action, recorded separately from the org role that authorized it.
+ */
+enum ActorLabel: string
+{
+    case User = 'user';
+    case PackagistAdmin = 'packagist-admin';
+    case System = 'system';
```
Seldaek: What's system for? How is it different than automation?

migrations/2026_06_organizations.sql
IgorBenko: Agreed, I think this can be removed later

src/Organization/EventStore/ActorLabel.php
IgorBenko: I think the thinking was `automation` are background workers and `system` could be a one-off command. But now I think we should just keep `automation` and remove `system` as it actually doesn't make sense

IgorBenko: Removed

Seldaek:
Thanks! Great first step :)

PR #1783, opened 2026-07-08
Avoid importing well-known transient branches

Seldaek:


src/Package/Updater.php
```diff
@@ -325,6 +325,10 @@ public function update(IOInterface $io, Config $config, Package $package, VcsRep
             if (!$version instanceof CompletePackageInterface) {
                 throw new \LogicException('Received a package instance of type '.$version::class.', expected a CompletePackageInterface instance');
             }
+            // skip branch names which are commonly short-lived and not meant to be published
+            if (Preg::isMatch('{^dev-(gh-readonly-queue/|renovate/|dependabot/).+}', $version->getPrettyVersion())) {
```
glaubinix: Do you want to add `conductor-` here too?

Seldaek: Yeah I was wondering if that was standard or is conductor/* in use too?

stof: Well, what is the naming used by Private Packagist's conductor feature ? You should be able to know it.

glaubinix: `conductor-` is the current standard. We previously used `conductor/` and `private-packagist-updates/` but skipped the slash because not everyone allows slashes in branch names.

src/Package/Updater.php
```diff
@@ -325,6 +325,10 @@ public function update(IOInterface $io, Config $config, Package $package, VcsRep
             if (!$version instanceof CompletePackageInterface) {
                 throw new \LogicException('Received a package instance of type '.$version::class.', expected a CompletePackageInterface instance');
             }
+            // skip branch names which are commonly short-lived and not meant to be published
+            if (Preg::isMatch('{^dev-(gh-readonly-queue/|renovate/|dependabot/|conductor-).+}', $version->getPrettyVersion())) {
```
TimWolla: From what I see, the dependabot branches don’t have a `dev-` prefix. I have:

- `dependabot/composer/*redacted*/phpunit/phpunit-13.2.2`
- `dependabot/github_actions/actions/cache-6`

in one of my repositories.

TimWolla: Ah, scratch that. The `dev-` prefix is added by Packagist, of course.

TimWolla (approved):

src/Package/Updater.php
stof: yeah, because the regex is not matching a branch name but a composer version.

Toflar: `gh-pages` might also be a good candidate 😊

Seldaek: Good point will add

Seldaek:
Cleanup SQL query to run in a while once old versions have been deleted:

```
delete s
from audit_log_search s
inner join audit_log a on a.id = s.auditLogId
where a.type in ('version_created','version_deleted')
  and (
        a.attributes->>'$.version' regexp '^dev-(gh-readonly-queue/|renovate/|dependabot/|conductor-|merge-up/).+'
     or a.attributes->>'$.version' in ('dev-gh-pages','dev-release-please--branches--main','dev-gitattributes')
  );

delete from audit_log
where type in ('version_created','version_deleted')
  and (
        attributes->>'$.version' regexp '^dev-(gh-readonly-queue/|renovate/|dependabot/|conductor-|merge-up/).+'
     or attributes->>'$.version' in ('dev-gh-pages','dev-release-please--branches--main','dev-gitattributes')
  );
```

Currently would get rid of 354288 useless audit log records.

PR #1785, opened 2026-07-10
Organization editing

IgorBenko:
This PR builds on the [organizations groundwork](https://github.com/composer/packagist/pull/1782) and lets an owner (or a packagist admin) **edit an existing organization**: for now to change its display name and its slug. Both changes are recorded as new events on the existing event stream. This PR adds on top of `OrganizationCreated` also `OrganizationNameChanged` and `OrganizationSlugChanged` events.

### Form 

A single form is used for both creation and editing. A new `/organizations/{organization}/settings` was added for editing. Editing requires 2FA, the same as the organization creation. 

Only fields that actually change are recorded. Submitting the form unchanged is a no-op. The aggregate enforces this too, so a name/slug that matches the current value records no event.

### Renaming events

*   **`OrganizationNameChanged`** and **`OrganizationSlugChanged`** each carry both the new and the previous value. (Note: we dropped `Renamed` from the Event name and now use `*Changed` for every property change. `OrganizationRenamed` is renamed to `OrganizationNameChanged`.

### Slug changes 

When a slug changes, the previous slug is reserved as a `RenamedFrom` reservation. Rather than 404, a request for the old slug now **302-redirects to the organization's current slug**:

*   `OrganizationResolver` looks up an active `RenamedFrom` reservation for an unknown slug and throws `OrganizationRenamedException`.
*   `OrganizationRedirectListener` turns that into a redirect, rebuilt from the current route with only the `organization` parameter swapped, so it works for any org route (show, settings, ...). A 302 is used on purpose: the reservation is temporary and can be released, so clients should not cache it.
*   `Deleted` reservations keep showing the gone page instead of redirecting.

### Reclaiming a previously owned slug

An organization can reclaim a slug that was freed by an earlier rename (e.g. `acme` → `acme-inc` → `acme`). `OrganizationSlugClaimGuard::assertClaimable()` now ignores reservations owned by the claiming org, and the projection releases the stale reservation on reclaim. 

### Access control & validation

*   The `EditDisplayInfo` / `EditSlug` actions were merged into a single `Edit` action, since a single settings form now covers both.
*   The reserved-word deny-list check moves into `OrganizationSlugClaimGuard`, so it applies to edits as well as creation, and now surfaces as `InvalidSlugException` (vs. `SlugTakenException`) to distinguish "not allowed" from "already taken".

Authors: [@glaubinix](https://github.com/glaubinix) and [@IgorBenko](https://github.com/IgorBenko)


tests/bootstrap.php
```diff
@@ -45,7 +45,7 @@ function executeCommand(string $command, bool $errorHandling = true): void
     }
 }
 
-if (!getenv('QUICK')) {
+if (!getenv('QUICK') && false) {
```
Seldaek: Undo please :)

IgorBenko: Haha, wonder how/when that happened?! 😄

src/ArgumentResolver/OrganizationResolver.php
```diff
@@ -47,10 +48,18 @@ public function resolve(Request $request, ArgumentMetadata $argument): iterable
 
         $organization = $this->organizationRepo->findOneBySlug($slug);
         if (null === $organization) {
+            $rename = $this->slugReservationRepo->findActiveRename($slug);
+            if (null !== $rename) {
+                $target = $this->organizationRepo->find($rename->orgId);
+                if (null !== $target && (!$target->isDeleted() || $this->security->isGranted('ROLE_ADMIN'))) {
```
Seldaek: why not role admin orgs?

src/ArgumentResolver/OrganizationResolver.php
```diff
@@ -47,10 +48,18 @@ public function resolve(Request $request, ArgumentMetadata $argument): iterable
...
+                    throw new OrganizationRenamedException($target->slug);
+                }
+            }
+
             throw new NotFoundHttpException('Organization not found.');
         }
 
-        if ($organization->isDeleted() && !$this->security->isGranted('ROLE_ADMIN_ORGS')) {
+        if ($organization->isDeleted() && !$this->security->isGranted('ROLE_ADMIN')) {
```
Seldaek: why not role admin orgs?

src/Entity/AuditRecord.php
```diff
@@ -182,6 +182,36 @@ public static function organizationCreated(Ulid $organizationId, string $slug, s
         );
     }
 
+    public static function organizationNameChanged(Ulid $organizationId, string $slug, string $displayName, string $previousDisplayName, ?User $actor): self
```
Seldaek: I'd argue actor probably should be non-nullable

```suggestion
    public static function organizationNameChanged(Ulid $organizationId, string $slug, string $displayName, string $previousDisplayName, User $actor): self
```

src/Entity/AuditRecord.php
```diff
@@ -182,6 +182,36 @@ public static function organizationCreated(Ulid $organizationId, string $slug, s
...
+                'org_name_from' => $previousDisplayName,
+                'org_name_to' => $displayName,
+                'actor' => self::getUserData($actor),
+            ],
+            $actor?->getId(),
+            organizationId: $organizationId,
+        );
+    }
+
+    public static function organizationSlugChanged(Ulid $organizationId, string $slug, string $displayName, string $previousSlug, ?User $actor): self
```
Seldaek: I'd argue actor probably should be non-nullable

```suggestion
    public static function organizationSlugChanged(Ulid $organizationId, string $slug, string $displayName, string $previousSlug, User $actor): self
```

src/Entity/Organization.php
```diff
@@ -70,4 +70,14 @@ public function isDeleted(): bool
     {
         return $this->deletedAt !== null;
     }
+
+    public function changeName(string $displayName): void
+    {
+        $this->displayName = $displayName;
+    }
```
Seldaek: Not sure why we need these methods if the props are public?

src/ArgumentResolver/OrganizationResolver.php
IgorBenko: No real reason, just an oversight. Fixed.

src/Entity/Organization.php
glaubinix: removed

src/Entity/Organization.php
```diff
@@ -70,4 +70,9 @@ public function isDeleted(): bool
     {
         return $this->deletedAt !== null;
     }
+
+    public function changeSlug(string $slug): void
```
Seldaek: Still a somewhat useless setter here

src/Security/Voter/OrganizationVoter.php
```diff
@@ -41,7 +41,9 @@ protected function voteOnAttribute(string $attribute, mixed $subject, TokenInter
 
         $action = OrganizationActions::from($attribute);
 
-        if ($this->security->isGranted('ROLE_ADMIN_ORGS')) {
+        // A packagist-admin may perform any owner action on any org for moderation, and is
+        // the only actor who can restore. Their authority derives from admin status.
+        if ($this->security->isGranted('ROLE_ADMIN')) {
```
Seldaek: Also should be ROLE_ADMIN_ORGS I believe?

Seldaek (approved):
Ok done reviewing, couple small nitpicks left but otherwise good to merge IMO, thanks!👍🏻

PR #1788, opened 2026-07-11
Fix overlapping X-axis labels in package stats charts

### LINKED ISSUE #1787: Package stats chart X-axis labels overlap near the right edge
## Summary

On package stats pages, the X-axis labels in the installs chart can overlap near the right edge.

This is visible on the package [page ](https://packagist.org/packages/savinmikhail/add_named_arguments_rector/stats) in the "Daily installs, averaged weekly" chart.

Original screenshot

## Reproduction

Open:
-[ `https://packagist.org/packages/savinmikhail/add_named_arguments_rector/stats`](https://packagist.org/packages/savinmikhail/add_named_arguments_rector/stats)

At the bottom-right of the main chart and the focus chart, the last X-axis label overlaps with its neighbor.

## Expected behavior

X-axis labels should remain readable and should not overlap at the right edge.

## Notes

From local investigation, the overlap appears to come from NVD3 rendering both regular tick labels and separate max/min labels for the X-axis.

savinmikhail:
Closes #1787.

## Summary

This prevents the X-axis labels in package stats charts from overlapping near the right edge.

NVD3 renders regular X-axis tick labels and separate max/min labels. On Packagist package stats pages, the extra right-edge max label can overlap with the neighboring tick label, especially in `lineWithFocusChart`.

This change disables those extra max/min labels for:
- `xAxis`
- `x2Axis`

Regular tick labels remain unchanged.

## Before

Before

## After

After

## Alternatives considered

1. Keep `showMaxMin(true)` and reduce the number of ticks.
   This is less predictable because NVD3 still decides which labels to keep, and it does not reliably eliminate the right-edge overlap for all date ranges and widths.

2. Rotate or stagger X-axis labels.
   This preserves more labels, but it is a much more visible UI change and makes both the main chart and the focus chart harder to scan.

3. Increase chart margins or adjust CSS for the last label.
   This is more fragile because the overlap comes from two different label layers being rendered at the same edge, not only from missing space.

4. Hide only one specific edge label with DOM/CSS hacks.
   This would be tightly coupled to NVD3's generated markup and would be more brittle than using the library option directly.

## Why this approach

Disabling `showMaxMin` is the smallest and most predictable fix:
- it targets the overlapping label layer directly;
- it keeps the chart data, tooltip behavior, and focus interaction unchanged;
- it avoids layout-specific hacks.

## Validation

I reproduced the issue locally with real stats JSON from:
- `https://packagist.org/packages/savinmikhail/add_named_arguments_rector/stats/all.json`
- `https://packagist.org/packages/savinmikhail/add_named_arguments_rector/stats/v0.1.26.json`

Then I rendered `charts.js` before and after the change and confirmed the overlap disappears.

### DIFF
--- a/js/charts.js
+++ b/js/charts.js
@@ -102,11 +102,15 @@ import '../css/charts.css';
                 return a;
             }
 
-            chart.xAxis.tickFormat(formatDate);
+            chart.xAxis
+                .showMaxMin(false)
+                .tickFormat(formatDate);
             chart.yAxis.tickFormat(formatDigit);
 
             if (withDatePicker) {
-                chart.x2Axis.tickFormat(formatDate);
+                chart.x2Axis
+                    .showMaxMin(false)
+                    .tickFormat(formatDate);
                 chart.y2Axis.tickFormat(formatDigit);
             }
 


Seldaek:
Thanks

PR #1792, opened 2026-07-21
Organization member and team management

glaubinix:
This PR builds on the [organizations groundwork](https://github.com/composer/packagist/pull/1782) and [organization editing](https://github.com/composer/packagist/pull/1785), and adds **team & member management** to an organization. It replaces the placeholder "owner = the creating user" model with membership derived from teams: every org is bootstrapped with an **`Owners`** team and an **`All organization members`** team, and ownership is now a membership lookup rather than the `organization.createdBy` column (which is dropped).

Each change is recorded as a new event on the existing organization event stream, so teams and memberships replay uniformly alongside the earlier `Organization*` events.

### New events

Seven events are added to the organization stream:

- **`TeamCreated`** / **`TeamRenamed`** / **`TeamDeleted`** — team lifecycle. `TeamCreated` carries a `kind` (`system` / `custom`) so the protection guards know which teams can't be renamed or deleted.
- **`TeamMemberAdded`** / **`TeamMemberRemoved`** — a user's membership of a single team.
- **`MemberRemoved`** / **`MemberLeft`** — a user removed from the whole org (all teams), or leaving on their own.

`OrganizationCreated` gains `ownersTeamId` + `allMembersTeamId`, and creation now records an explicit five-event batch (the org, its two system teams, then the creator joining each) rather than special-casing bootstrap state in the projections.

### Aggregate invariants

Teams and memberships live inside the `Organization` aggregate's consistency boundary, so cross-member rules are enforced in one transaction:

- **Last owner is protected** — the `Owners` team can never drop to zero; the sole owner cannot be removed or leave.
- **System teams are protected** — `Owners` / `All organization members` cannot be renamed or deleted; the all-members roster is managed automatically and can't be edited directly.
- **Team names** — a new `TeamName` value object guards the format (letters, numbers, spaces, hyphens, max 40 chars); reserved names (`Owners`) and case-insensitive uniqueness within the org are enforced by the aggregate, mirroring how slug rules live outside `Slug`.
- **Joining is invitation-only** — a user must already be an org member to be added to a further team, and becoming an owner requires the target to have 2FA enabled.

### Read-model projections

Two query-friendly tables are projected from the stream:

- **`organization_team`** — teams, with a generated `nameLower` column backing a `(orgId, nameLower)` unique index. A concurrent duplicate rolls back and is mapped to a clean "team name already taken" error.
- **`organization_team_member`** — memberships, with `orgId` denormalised so org-scoped queries skip the join. A user's org membership is the union of their team memberships.

### HTTP, forms & UI

- New routes under `/organizations/{organization}` for **teams** (list, create, rename, delete), **team members** (add, remove), **members** (list, remove), and **leave**, plus `Teams` and `Members` menu entries and templates.
- Three argument resolvers load route parameters into typed controller arguments: `OrganizationTeam` from a team id, and `User` from a **username** for `organizationMember` / `teamMember`. Members are addressed by username (not id) so no internal user id leaks into URLs; a username that exists but isn't a member resolves to a 404 exactly like a missing one.
- Removing the last owner (or having them leave) shows an explanatory page instead of a form.

### Access control

- **`OrganizationVoter`** is rebuilt around membership: `View` / `ViewTeams` / `ViewMembers` / `Leave` require org membership, while all management actions require an **owner with 2FA**. Denials now carry a structured **`OrganizationAccessDeniedReason`**, and the class-level admin guard on `OrganizationController` is replaced by a per-action `#[IsGranted]`.
- **`OrganizationPolicyAccessDeniedListener`** turns an owner-blocked-only-by-2FA denial into a redirect to the 2FA setup page rather than a bare 403.
- `OrganizationActions` gains the team/member action set; packagist admins retain the moderation bypass.

### Transparency log

Every team lifecycle change and every membership change for a user who has actually joined is written to the public `audit_log` (new `AuditRecordType` cases + display classes/templates), identified solely by username.


src/ArgumentResolver/TeamMemberResolver.php
```diff
@@ -0,0 +1,61 @@
...
+        if (User::class !== $argument->getType() || 'teamMember' !== $argument->getName()) {
+            return [];
+        }
+
+        $teamId = $request->attributes->getString('team');
+        if (!Ulid::isValid($teamId)) {
+            throw new NotFoundHttpException('Team member not found.');
+        }
+
+        $member = $this->organizationTeamMemberRepo->findTeamMember(Ulid::fromString($teamId), $request->attributes->getString('teamMember'));
```
Seldaek, line 51: Curious if we should as a defense in depth check here that the logged in user has read access on the team's org (and on the org in the orgMemberResolver)? Just to make sure this cannot resolve the wrong thing in the wrong place accidentally.

src/Controller/OrganizationController.php
```diff
@@ -47,6 +67,7 @@ public function list(#[CurrentUser] User $user): Response
         ]);
     }
 
+    #[IsGranted('ROLE_ADMIN_ORGS')]
```
Seldaek: I'd maybe move these admin only paths under /admin and into a new controller so the isGranted can be kept at class level which is a lot safer.

migrations/2026_07_organization_teams.sql
```diff
@@ -0,0 +1,41 @@
...
+-- NOTE: the organization-created event payload gained ownersTeamId + creatorUserId
+-- in this stage, so any pre-existing dev event data must be discarded:
+--   DELETE FROM organization_event; DELETE FROM organization; DELETE FROM slug_reservation;
+
+CREATE TABLE organization_team (
+    teamId BINARY(16) NOT NULL,
+    orgId BINARY(16) NOT NULL,
+    kind VARCHAR(16) NOT NULL,
+    name VARCHAR(40) NOT NULL,
+    nameLower VARCHAR(40) GENERATED ALWAYS AS (LOWER(name)) STORED,
```
Seldaek: Do we really need this column? If it's only for the unique, it shouldn't be necessary.. just gotta make sure to use the right charset:

> uniqueness follows the column's collation, not the charset, and MySQL's default utf8mb4 collations are case-insensitive.
> 
> - utf8mb4_0900_ai_ci (MySQL 8.0+ default) → case-insensitive and accent-insensitive. So Foo, foo, and fóo all collide.
> - utf8mb4_general_ci / utf8mb4_unicode_ci (older default) → case-insensitive, but accent handling varies. Foo/foo collide.

templates/audit_log/display/organization_member_left.html.twig
```diff
@@ -0,0 +1,3 @@
+{% import 'audit_log/macros.html.twig' as auditLog %}
+
+<strong>{{ display.member.username }}</strong> left <strong>{{ auditLog.organizationLink(display.organization) }}</strong>
```
Seldaek: Can we link the usernames too in all audit log templates? Just makes it easier to click through

src/Entity/AuditRecord.php
```diff
@@ -212,6 +212,111 @@ public static function organizationSlugChanged(Ulid $organizationId, string $slu
...
+        );
+    }
+
+    public static function organizationMemberLeft(Ulid $organizationId, string $slug, string $displayName, ?User $member): self
+    {
+        return new self(
+            AuditRecordType::OrganizationMemberLeft,
+            [
+                'organization' => new OrganizationDisplay((string) $organizationId, $slug, $displayName)->toRecord(),
+                'member' => self::getUserData($member),
```
Seldaek: All the org member/ team member events should use `user` instead of `member`, or we need to add support for `member` in self::getSearchTerms

src/Controller/OrganizationController.php
glaubinix: Generally there is no more need for `#[IsGranted('ROLE_ADMIN_ORGS')]` as the create endpoint is moved elsewhere and keeping that there makes it rather tricky to write tests which checks permissions because the `ROLE_ADMIN_ORGS` works around that. Want us to introduce a temp `ROLE_VIEW_ORGS` for now until we grant users access to it?

templates/audit_log/display/organization_member_left.html.twig
glaubinix: I did this for now only for the new organization logs. Happy to send a follow up PR where this gets applied to any username rendered in the audit log

src/Controller/OrganizationController.php
Seldaek: Think its fine to remove it. I commented on that before i saw the other PR moving things

templates/audit_log/display/organization_member_left.html.twig
Seldaek: Yup all good

src/ArgumentResolver/TeamMemberResolver.php
glaubinix: Added for teams and org members

migrations/2026_07_organization_teams.sql
glaubinix: Learned something new today! Simplified this a bit to limit which characters we accept to keep PHP + MySQL validation identical.

templates/audit_log/macros.html.twig
```diff
@@ -6,6 +6,14 @@
     {%- endif -%}
 {% endmacro %}
 
+{% macro userLink(username) %}
+    {%- if username is existing_user -%}
```
Seldaek, line 10: Not sure if this check is worth the amount of DB queries it'll generate.. We do it for packages because there it's cached and a quick lookup, but for users/org names kinda unsure if we should check or just always link and worst case you get a 404.. This only affects the audit log views which seems like an acceptable trade-off for me, but let's leave it as is for now, just have to keep it in mind if the audit log viewing gets slow.

Seldaek:
Thanks!

PR #1794, opened 2026-07-24
Organization: add organization internal audit log page

glaubinix:
Adds a per-organization audit log page so org managers can review the transparency-log records for their own organization, without the site-wide /transparency-log view. Records are scoped to the single organization and reuse the existing audit display/rendering layer.

- New `GET /organizations/{organization}/audit-log` route and `auditLog()` action on `OrganizationController`, gated by a new `OrganizationActions::ViewAuditLog` permission (wired into `OrganizationVoter` alongside the other management actions). The query filters on `audit_log.organizationId`, orders newest-first, and paginates at 20/page via Pagerfanta.
- Supports the same filters as the global log: type (multi-select), actor, and date/time from/to. The type dropdown is populated from a new `AuditRecordType::organizationCases()` helper (organization-category types only). Actor filtering is admin-aware (`ROLE_AUDITOR`), and the IP column is shown only to auditors.
- Adds an `organization_idx` index on `audit_log.organizationId` so per-organization filtering doesn't scan the whole table.
- Adds an "Audit log" item to the organization menu (and switches the organization overview icon from dashboard to chart). New menu.organization_audit_log translation.
- New `templates/organization/audit_log.html.twig` with the filter form, an active time-range banner, the results table, and an empty state.

Screenshot 2026-07-24 at 14 39 23

### DIFF
--- a/migrations/2026_07_audit_log_organization_index.sql
+++ b/migrations/2026_07_audit_log_organization_index.sql
@@ -1,0 +1,5 @@
+-- Index audit_log.organizationId so the per-organization audit log page
+-- (organization_audit_log) can filter records for a single organization
+-- without scanning the whole table.
+
+ALTER TABLE audit_log ADD INDEX organization_idx (organizationId);
--- a/src/Audit/AuditRecordType.php
+++ b/src/Audit/AuditRecordType.php
@@ -73,6 +73,14 @@ enum AuditRecordType: string
     case OrganizationMemberRemoved = 'organization_member_removed';
     case OrganizationMemberLeft = 'organization_member_left';
 
+    /**
+     * @return list<self>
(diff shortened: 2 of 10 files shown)


Seldaek:
> * Supports the same filters as the global log: type (multi-select), actor, and date/time from/to. The type dropdown is populated from a new `AuditRecordType::organizationCases()`

I'm wondering if we shouldn't include all package/version events that are part of the org here.. Probably would make sense but the org id then needs to be stored on those too.

Seldaek (approved):

glaubinix:
Good point! We definitely should display anything that is related to the organization in there. We'll think about how to best filter that information once organizations have access to packages.

PR #1802, opened 2026-07-30
Organizations invitations

glaubinix:
# Organization invitations

This PR builds on the [organization member and team management](https://github.com/composer/packagist/pull/1792) groundwork and adds **membership invitations**: an owner invites an email address to one or more teams, the invitee receives a single-use link, and accepting it makes them a member. Invitations are modelled as their own event-sourced aggregate persisted on the shared `organization_event` stream (aggregate id = the invitation's ULID, not the org's), so they replay uniformly alongside the existing `Organization*` events.

The same change also replaces the "membership is the union of team memberships" lookup with an explicit **`organization_member`** record, so an org-level join carries facts that team rows can't express (when a user joined) and `isMember` becomes a direct lookup.

### New events

The invitation lifecycle is captured by six events on the invitation stream, all implementing an **`InvitationEvent`** marker interface (it exposes `email()` and lets the org projectors route invitation events to their own handling):

- **`UserInvitationSent`**: the invitation is created. Carries the invitation + org ULIDs, `email` / `emailCanonical`, target `teamIds`, the link `tokenHash`, and `expiresAt`.
- **`UserInvitationResent`**: a fresh link token and a bumped expiry.
- **`UserInvitationRevoked`**: an owner cancels a pending invitation.
- **`UserInvitationDeclined`**: the invitee declines.
- **`UserInvitationExpired`**: a pending invitation is swept past its expiry (recorded lazily, see below).
- **`UserInvitationAccepted`**: the invitee accepts; carries the accepting `userId` and the `teamIds` actually joined.

Acceptance also records a **`MemberJoined`** event on the organization stream (org id + `userId` + the nullable `invitationId`). The founding owner now joins via this same event at creation time, so "how did this user become a member" is a single uniform fact rather than a special case.

### Invitation aggregate & invariants

The `Invitation` aggregate (`src/Organization/Domain/Invitation.php`) owns the lifecycle rules; each transition throws a dedicated domain exception rather than mutating from an invalid state:

- **No duplicate pending invitation**: a second invite to the same canonical email while one is still pending throws `DuplicatePendingInvitationException`.
- **Only pending invitations transition**: resend on a resolved invitation throws `NoPendingInvitationException`; accept on a non-pending one throws `InvitationNotPendingException`, and on a logically-expired one throws `NoPendingInvitationException`. A resolved invitation is never revived; a fresh `UserInvitationSent` is required instead.
- **Email must match**: the invitee's account email (canonical) must equal the invited address, else `EmailMismatchException`.
- **Owner 2FA policy**: accepting into the `Owners` team requires the account to have two-factor enabled, else `PolicyNotMetException`.
- **Not already a member**: accepting when the user already belongs to the organization throws `AlreadyMemberException`, so no duplicate `MemberJoined` is recorded; the landing page offers to dismiss the invitation instead of accepting it.
- **Targets must still exist**: if every invited team has since been deleted, accepting throws `TeamNotFoundException`.

A small **`Email`** value object guards format (non-empty, ≤255 chars, `FILTER_VALIDATE_EMAIL`) and exposes the lower-cased `canonical` form used for uniqueness and matching. `InvitationStatus` is a stable-backed enum. `Pending` is the only active state; `Accepted` / `Revoked` / `Declined` / `Expired` are terminal.

### Read-model projections & schema

`migrations/2026_07_organization_invitations.sql` adds three tables, projected by `InvitationReadModelProjector`:

- **`organization_invitation`**: one row per invitation. Stores only the **SHA-256 `tokenHash`**; the raw token lives solely in the emailed link and is never persisted (event stream included). Indexes back the hot paths: `(orgId)`, `(orgId, emailCanonical, status)` for the duplicate-pending check, and `(status, expiresAt)` for the expiry sweep.
- **`organization_invitation_team`**: the target teams to join on acceptance. Deliberately has **no FK to `organization_team`**, so deleting a team after sending doesn't cascade the historical target away.
- **`organization_member`**: the org-level membership record (`orgId`, `userId`, `joinedAt`). Team memberships still drive access; this carries the org-scoped facts team rows can't. `OrganizationVoter::isMember` now reads `OrganizationMemberRepository` directly.

An invitation is stored with a persisted `status`, but the read model distinguishes an **effective status**: a still-`Pending` row whose `expiresAt` has passed is reported as `expired` and `isActive()`/`effectiveStatus()` treat it as such (`isPending()` stays a plain status check), even before the cron sweep flips the column.

### HTTP, forms & UI

Owner-facing management routes under `/organizations/{organization}/invitations`, all linked from the Members page:

- `organization_invitation_create`: `…/invite` (GET/POST), `InviteMemberType` (email + team checkboxes; the auto-managed *All organization members* team is shown for clarity).
- `organization_invitation_resend`: `…/{invitation}/resend` (GET/POST), `ResendInvitationType`, its own confirmation page (resending rotates the token and extends expiry).
- `organization_invitation_revoke`: `…/{invitation}/revoke` (GET/POST), `RevokeInvitationType`, its own page.

Invitee-facing token pages (invitation addressed by ULID + a 64-hex token, resolved via `OrganizationInvitationResolver`):

- `organization_invitation_show`: `…/{invitation}/{token}` (GET), the accept/decline landing page.
- `organization_invitation_accept`: `…/{token}/accept` (POST).
- `organization_invitation_decline`: `…/{token}/decline` (POST).

`OrganizationInvitationResolver` verifies the token while resolving the invitation, so a controller action only ever sees a valid token on an actionable invitation. A token that does not match, or an invitation resolved for any other reason (accepted/declined/revoked), is a 404. A *matching* token on an expired invitation proves the visitor received the email, so it raises `InvitationExpiredException`, which `InvitationExpiredListener` turns into a **410** rendering `organization/invitation_expired.html.twig` with the expiry date and how to get a fresh link. The landing page never echoes the invited address back: an email mismatch is reported without revealing it.

`InvitationManager` orchestrates the flow (validate teams, generate + hash the token via `InvitationTokenGenerator`, persist, send the email in `templates/email/organization_invitation.txt.twig`). The members and invitations views are merged into a single **Members** page: the members list with an "Invite a member" action, and a secondary **Pending invitations** section (owner-only) listing only pending invitations, with status, last-sent, expiry and Resend/Revoke actions. A pending row already past its expiry stays listed as `expired` without actions until the cron sweep resolves it; accepted/declined/revoked/expired invitations drop off entirely and their history lives in the audit log.

### Access control

`OrganizationActions` gains **`ViewInvitations`**, **`InviteMember`**, **`ResendInvitation`**, and **`RevokeInvitation`**; all four resolve through `OrganizationVoter` to the same manage-level guard as the other org mutations. An **owner with 2FA** (packagist admins keep the moderation bypass). The invitee accept/decline/show pages are token-gated and only require `ROLE_USER`.

### Transparency log

Six audit types are added, `OrganizationInvitationSent` / `Resent` / `Revoked` / `Accepted` / `Declined` / `Expired`, plus `OrganizationMemberJoined`, each with a display class and template. Unlike team/membership records (identified by username), invitation records carry the invited **email**. `OrganizationAuditProjector` publishes them, and `AuditLogDisplayFactory` **obfuscates the email at display time** from anyone who is neither a `ROLE_AUDITOR` nor a member of the invitation's organization.

### Cron: expiring invitations

`ExpireOrganizationInvitationsCommand` sweeps pending invitations past their expiry (`findPendingExpired`, backed by `org_invitation_expiry_idx`, batched by `MAX_PER_RUN` so a backlog drains over several runs, lock-guarded, with a `--dry-run`). Each due invitation is flipped to `Expired` via `InvitationManager::expire()`, which records `UserInvitationExpired`; a `ConcurrencyException` (resolved in the meantime) is skipped and re-checked next run. The invitation expiry window is `InvitationManager::INVITATION_EXPIRY_DAYS = 7`.


migrations/2026_07_organization_invitations.sql
```diff
@@ -0,0 +1,53 @@
...
+--
+-- The invitation aggregate is a separate ULID stream persisted in the shared organization_event
+-- table; no new event table is needed. Pre-membership invitation events (sent/resent/revoked/
+-- declined/expired) stay internal (event stream + these read models) and never reach audit_log.
+
+CREATE TABLE organization_invitation (
+    id BINARY(16) NOT NULL,
+    orgId BINARY(16) NOT NULL,
+    email VARCHAR(255) NOT NULL,
+    emailCanonical VARCHAR(255) NOT NULL,
```
Seldaek: Do we really need a canonical version? utf8mb4 is case insensitive so I don't think this actually achieves anything? I know it's on the user table but that's historic and equally useless AFAIK.

glaubinix: Just force of habit. Let me drop this

src/Entity/AuditRecord.php
```diff
@@ -288,6 +288,28 @@ public static function organizationTeamMemberRemoved(Ulid $organizationId, strin
...
     }
 
+    /**
+     * A user became a member of the org (by accepting an invitation, or as the founding owner at
+     * creation). The teams they landed in are logged separately as organization_team_member_added
+     * entries. This member-facing entry identifies the member by username; the invitation lifecycle
+     * itself (sent/resent/revoked/declined/accepted/expired, which carries the invited email) is logged
+     * separately by the organization_invitation_* records.
+     */
+    public static function organizationMemberJoined(Ulid $organizationId, string $slug, string $displayName, ?User $member): self
```
Seldaek: I'd expect this as signature:

```suggestion
    public static function organizationMemberJoined(Ulid $organizationId, string $slug, string $displayName, User $member, ?User $actor): self
```

Namely: 
- always include an actor, because who knows maybe at some point it'll be different than the member.
- why is member nullable?

src/Entity/OrganizationInvitationTeam.php
```diff
@@ -0,0 +1,41 @@
...
+        #[ORM\Id]
+        #[ORM\Column(type: 'ulid')]
+        public readonly Ulid $invitationId,
+
+        #[ORM\Id]
+        #[ORM\Column(type: 'ulid')]
+        public readonly Ulid $teamId,
+    ) {
+    }
+}
```
Seldaek: Not a huge deal to leave as is, but I wonder why not just use a json field on the invitation table that has the team ids? Is it just about being able to join teams when querying? Or what am I missing?

src/Entity/AuditRecord.php
glaubinix: Members are nullable because if we want to rerun the audit projection and the user has since been deleted. I guess this is unlikely, changing and we can deal with this in case we ever think about wanting to do this.

src/Organization/Domain/Invitation.php
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    public function isExpired(\DateTimeImmutable $now): bool
+    {
+        return $now > $this->expiresAt;
+    }
+
+    private function assertEmailMatches(string $userEmailCanonical): void
+    {
+        if ($userEmailCanonical !== $this->emailCanonical) {
```
Seldaek: Is this really a must? I don't think it should be. If you invite me to foo@example.org but I used foo+packagist@example.org as my packagist email, why can't I accept the invite without creating a second account? That sounds annoying.

It'd be mitigated if we add support for multiple emails, but even then, I don't really see why there should be this restriction.

src/Entity/OrganizationInvitationTeam.php
glaubinix: no real reason. Changed to use a JSON column

src/Organization/Domain/Invitation.php
glaubinix: Sure, dropped. I guess if organizations want to enforce something like it they will be able to define a list of allowed email domains and if really necessary then we can always offer this as an optional organization policy.

src/Entity/OrganizationInvitation.php
```diff
@@ -0,0 +1,119 @@
...
+
+        /** Set when the invitation leaves the pending state for any reason. */
+        #[ORM\Column(type: 'datetime_immutable', nullable: true)]
+        public ?\DateTimeImmutable $resolvedAt = null,
+    ) {
+    }
+
+    public function isPending(): bool
+    {
+        return $this->status === InvitationStatus::Pending;
```
Seldaek: Not sure if this should match https://github.com/composer/packagist/pull/1802/changes#diff-7e9a6db5d072fd151910a2fd13e39264ab31589f0f5c57f467143ef6ecd1a058R198 - Also isActive/effectiveStatus are missing from the Domain\Invitation class. I am not sure if they should mirror each other or not, but it just tripped me up while reviewing as I had a weird déjà vu ;)

templates/organization/invitation_show.html.twig
```diff
@@ -0,0 +1,43 @@
...
+        <div class="alert alert-info">
+            You are already a member of <strong>{{ organization.displayName }}</strong>, so there is nothing to accept.
+            An owner can add you to further teams from the organization's team management.
+        </div>
+
+        <div class="btn-toolbar">
+            {{ form_start(declineForm, { action: path('organization_invitation_decline', { organization: organization.slug, invitation: invitation.id, token: token }), attr: { style: 'display:inline' } }) }}
+                <button class="btn btn-default" type="submit">Dismiss invitation</button>
+            {{ form_end(declineForm) }}
+        </div>
```
Seldaek, line 22: Not sure what the point of this is.. If the user is already a member, what does dismissing the invitation do really? It uses the decline endpoint but I guess it should really just mark it accepted or expired, it's not really relevant, but Dismiss sounds like something bad may happen to your membership. I'd maybe just call the button "Continue to organization" or smth if that is where the decline page leads to after.

Seldaek (approved):

src/Entity/OrganizationInvitation.php
glaubinix: Good call. Cleaned this up. The API should now be similar and only methods we really need are no public

templates/organization/invitation_show.html.twig
glaubinix: Converted this into a "Continue to organization" link which doesn't touch the invitation

Seldaek:
Thanks

PR #1803, opened 2026-08-01
Only re-dump package metadata when its dumpable content changed

Seldaek:
Re-dumping v2 metadata used to be driven almost entirely by `crawledAt`: any crawl re-staled the package, so the monthly "crawl all" flooded the dumper with packages whose files had not changed. This makes re-dumps content-driven instead — a crawl marks the package only when it actually changed dumpable content: a version created or mutated (which also covers the default-branch toggle, the source/dist URL rewrite and the abandoned-state change, all of which surface as a `VersionUpdatedResult`), an auto-recovered version, or an auto-soft-deleted one. An unchanged re-crawl no longer marks anything.

### Marking is non-destructive

Marking used to null `dumpedAtV2`, which loses requests: `V2Dumper::dump()` records `dumpedAtV2` for a whole batch only at the end, so a mark landing between a package's hydration and that write was silently overwritten. `crawledAt` masked that for the Updater, but `SecurityAdvisoryUpdateListener` and `FilterListEntryUpdateListener` never touch `crawledAt` — so a newly published advisory could stay missing from `p2/` until an unrelated crawl. That was a live bug, and invisible to any file-level metric since no file is written.

A new `package.dumpRequestedAt` column records the request instead, and staleness becomes `dumpedAtV2 IS NULL OR dumpRequestedAt >= dumpedAtV2`. `$dumpTime` is captured *before* hydration, so a mark landing mid-run leaves the package stale and it is picked up next run. `>=` rather than `>` because both columns hold whole seconds: a same-second mark is ambiguous and must resolve towards re-dumping.

Two levels of request:

- `Package::markForDump()` — "re-dump if the content changed". Every ordinary content-change path uses this; `writeV2File()` still compares content and early-returns if it matches.
- `Package::forceDump()` — additionally nulls `dumpedAtV2`, which `writeV2File()` reads as "write this regardless", re-uploading byte-identical files to bust the CDN cache and re-run the replica write. Reserved for the deliberate "make it publish again" paths: `Updater::FORCE_DUMP` (manual update button, unfreeze, repo URL rewrite) and `Package::unfreeze()`. It always marks as well, so an in-flight dump run can at worst cost the *forcing*, never the request.

`VersionRepository::softDelete()` / `recover()` now mark directly rather than relying on the scheduled crawl: pulling a version is the security path, and a package whose repository 404s never reaches the end of `Updater::update()`, which would leave a pulled version published indefinitely. `remove()` marks directly too, but not for rows that were already soft-deleted — those were excluded from the dumped metadata already, so hard-purging them changes no bytes.

`Scheduler::scheduleUpdate()` keys its dedup on `(packageId, update_source_dist_url, delete_before)`, so a pending plain job used to swallow a `force_dump` request entirely. It now carries the intent both ways: onto an already-queued job, and onto the replacement job when a scheduled-for-later job is cancelled in favour of an immediate one.

### Convergence fixes in the Updater

`updateInformation()` rebuilt a dev version on *every* crawl, forever, whenever its rebuild trigger could not agree with what the reconciliation stores — bumping `Version::updatedAt` (= `published-time` for dev rows), i.e. a real CDN upload + replica write + purge each time. Three causes, all fixed:

- the replacement was only written while *transitioning* to abandoned, so a standing mismatch never self-healed. It is reconciled on every crawl that sees upstream abandonment now (composer.json wins; a UI-only abandon is untouched);
- the trigger compared the raw upstream value against the sanitized stored one, so any replacement `sanitize()` alters could never match. Both sides go through `normalizeReplacementPackage()` now;
- the trigger fired for every dev version while the reconciliation is scoped to the default branch, so a non-default branch declaring abandonment the default branch disagrees with never converged. Both are default-branch-scoped now — version rows carry no abandoned state of their own, `Version::toV2Array()` reads it off the package at dump time.

### Instrumentation and the transitional net

The `crawledAt` staleness clause is deliberately **unchanged**, so selection remains a strict superset and there is no regression. Dropping it is a follow-up, gated on a new gap-detector: `metadata_dump.file{result:written, requested:false}` counts content changes that only `crawledAt` caught, i.e. a change path that fails to mark, and must sit at ~0 first. Its inverse `{result:skipped, requested:true}` catches paths that mark more eagerly than needed. `result:created` splits out first writes of a missing local file so they do not pollute the gate, and the legacy `filemtime >= 1606210609` re-dump branch — which re-verified any p2 file whose mtime predates 2020-11-24 — is removed for the same reason.

### Schema

Two migrations, one on each side of the deploy, because the stale query pins its index by name with `USE INDEX`:

- `2026_08_package_dump_requested_at.sql` — **before** the deploy: adds `dumpRequestedAt` and `dumped2_requested_crawled_frozen_idx`.
- `2026_08b_drop_dead_dump_columns.sql` — **after** the deploy: drops `dumped_idx`, the dead v1 `dumpedAt` column (the v1 dumper is gone and its readers had no callers), and the superseded `dumped2_crawled_frozen_idx`.

### DIFF
--- a/migrations/2026_08_package_dump_requested_at.sql
+++ b/migrations/2026_08_package_dump_requested_at.sql
@@ -7,7 +7,7 @@
 -- was silently overwritten and the change never dumped. The crawledAt clause masked this for the
 -- Updater, but not for the advisory / filter-list listeners, which never touch crawledAt.
 -- dumpRequestedAt records the request instead, and staleness compares the two, so a mark can no
--- longer be lost. See metadata-dump-followups.md.
+-- longer be lost.
 --
 -- Two statements on purpose: on its own the ADD COLUMN can run ALGORITHM=INSTANT, whereas folding it
 -- into the index DDL drags it down to an INPLACE rebuild of the whole package table.
--- a/migrations/2026_08_package_dump_requested_at.sql
+++ b/migrations/2026_08_package_dump_requested_at.sql
@@ -2,12 +2,9 @@
 -- the new code needs both the column and the index to exist from the moment it starts. The matching
 -- drops live in 2026_08b_drop_dead_dump_columns.sql, which must run AFTER the deploy.
 --
--- Marking a package for re-dump used to null dumpedAtV2, which is destructive: V2Dumper writes
--- dumpedAtV2 once at the end of a run, so a mark landing between a package's hydration and that write
--- was silently overwritten and the change never dumped. The crawledAt clause masked this for the
--- Updater, but not for the advisory / filter-list listeners, which never touch crawledAt.
--- dumpRequestedAt records the request instead, and staleness compares the two, so a mark can no
--- longer be lost.
+-- Marking a package for re-dump used to null dumpedAtV2, which is destructive: V2Dumper writes that
+-- once at the end of a run, so a mark landing mid-run was silently overwritten and never dumped.
+-- dumpRequestedAt records the request instead and staleness compares the two, so it cannot be lost.
 --
 -- Two statements on purpose: on its own the ADD COLUMN can run ALGORITHM=INSTANT, whereas folding it
 -- into the index DDL drags it down to an INPLACE rebuild of the whole package table.
--- a/migrations/2026_08_package_dump_requested_at.sql
+++ b/migrations/2026_08_package_dump_requested_at.sql
@@ -1,10 +1,15 @@
+-- Run BEFORE deploying the code: purely additive, so the currently-deployed code keeps working while
+-- the new code needs both the column and the index to exist from the moment it starts. The matching
+-- drops live in 2026_08b_drop_dead_dump_columns.sql, which must run AFTER the deploy.
+--
 -- Marking a package for re-dump used to null dumpedAtV2, which is destructive: V2Dumper writes
 -- dumpedAtV2 once at the end of a run, so a mark landing between a package's hydration and that write
 -- was silently overwritten and the change never dumped. The crawledAt clause masked this for the
 -- Updater, but not for the advisory / filter-list listeners, which never touch crawledAt.
 -- dumpRequestedAt records the request instead, and staleness compares the two, so a mark can no
 -- longer be lost. See metadata-dump-followups.md.
-ALTER TABLE package
-  ADD dumpRequestedAt DATETIME DEFAULT NULL,
-  ADD INDEX dumped2_requested_crawled_frozen_idx (dumpedAtV2, dumpRequestedAt, crawledAt, frozen),
-  DROP INDEX dumped2_crawled_frozen_idx;
+--
+-- Two statements on purpose: on its own the ADD COLUMN can run ALGORITHM=INSTANT, whereas folding it
+-- into the index DDL drags it down to an INPLACE rebuild of the whole package table.
+ALTER TABLE package ADD dumpRequestedAt DATETIME DEFAULT NULL;
+ALTER TABLE package ADD INDEX dumped2_requested_crawled_frozen_idx (dumpedAtV2, dumpRequestedAt, crawledAt, frozen);
--- a/migrations/2026_08_package_dump_requested_at.sql
+++ b/migrations/2026_08_package_dump_requested_at.sql
@@ -1,0 +1,10 @@
+-- Marking a package for re-dump used to null dumpedAtV2, which is destructive: V2Dumper writes
+-- dumpedAtV2 once at the end of a run, so a mark landing between a package's hydration and that write
+-- was silently overwritten and the change never dumped. The crawledAt clause masked this for the
+-- Updater, but not for the advisory / filter-list listeners, which never touch crawledAt.
+-- dumpRequestedAt records the request instead, and staleness compares the two, so a mark can no
+-- longer be lost. See metadata-dump-followups.md.
+ALTER TABLE package
+  ADD dumpRequestedAt DATETIME DEFAULT NULL,
+  ADD INDEX dumped2_requested_crawled_frozen_idx (dumpedAtV2, dumpRequestedAt, crawledAt, frozen),
+  DROP INDEX dumped2_crawled_frozen_idx;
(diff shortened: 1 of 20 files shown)


migrations/2026_08_package_dump_requested_at.sql
```diff
@@ -0,0 +1,15 @@
+-- Run BEFORE deploying the code: purely additive, so the currently-deployed code keeps working while
+-- the new code needs both the column and the index to exist from the moment it starts. The matching
+-- drops live in 2026_08b_drop_dead_dump_columns.sql, which must run AFTER the deploy.
+--
+-- Marking a package for re-dump used to null dumpedAtV2, which is destructive: V2Dumper writes
+-- dumpedAtV2 once at the end of a run, so a mark landing between a package's hydration and that write
+-- was silently overwritten and the change never dumped. The crawledAt clause masked this for the
+-- Updater, but not for the advisory / filter-list listeners, which never touch crawledAt.
+-- dumpRequestedAt records the request instead, and staleness compares the two, so a mark can no
+-- longer be lost. See metadata-dump-followups.md.
```
glaubinix: I don't see this file anywhere

src/Entity/PackageRepository.php
```diff
@@ -288,32 +271,22 @@ public function getStalePackagesForIndexing(): array
...
     {
         $conn = $this->getEntityManager()->getConnection();
 
-        $sql = 'SELECT p.id FROM package p USE INDEX (dumped2_crawled_frozen_idx) WHERE (p.dumpedAtV2 IS NULL OR (p.dumpedAtV2 <= p.crawledAt AND p.crawledAt < NOW())) AND (p.frozen IS NULL OR p.frozen NOT IN (:suppressed))';
+        // dumpRequestedAt >= dumpedAtV2 means a re-dump was requested at or after the last dump. The >=
+        // is deliberate: both columns hold whole seconds, so a request landing in the same second as
+        // the dump is ambiguous and must count as stale — the redundant pass costs a content comparison
+        // in V2Dumper::writeV2File() and reports result:skipped, whereas a missed one is a lost change.
+        // The crawledAt clause is the transitional safety net kept until the gap-detector proves every
+        // content change marks the package; see metadata-dump-followups.md.
```
glaubinix: Same thing about the file

src/Entity/Package.php
```diff
@@ -693,24 +692,76 @@ public function getIndexedAt(): ?\DateTimeImmutable
...
+        return $this->dumpedAtV2 === null;
+    }
+
+    /**
+     * Whether something requested a re-dump since the package was last dumped.
+     *
+     * Deliberately NOT a mirror of PackageRepository::getStalePackagesForDumpingV2(): the stale query
+     * also sweeps in packages whose crawledAt is newer than their dump, and this must not, or the
+     * metadata_dump.file{requested} gap-detector could no longer tell a package that was marked apart
+     * from one that transitional clause merely caught. See metadata-dump-followups.md.
```
glaubinix: and again here

src/Entity/Package.php
```diff
@@ -693,24 +692,76 @@ public function getIndexedAt(): ?\DateTimeImmutable
...
     {
-        $this->dumpedAtV2 = $dumpedAt;
+        $this->dumpRequestedAt = new \DateTimeImmutable();
     }
 
-    public function getDumpedAtV2(): ?\DateTimeImmutable
+    /**
+     * Request a dump that re-writes the files even if their content is byte-identical, which busts the
+     * CDN cache and can shake a stuck storage replication loose. Use it for the deliberate "make it
+     * publish again" paths — the manual update button, an unfreeze — not for ordinary content changes,
```
glaubinix: Do you want to list all scenarios here or make it clearer that this is just a subset? For instance soft deleting/recovering a version also force dumps

src/Entity/Package.php
```diff
@@ -100,11 +100,10 @@ public static function casesForRole(bool $mayDisablePackages): array
 #[ORM\UniqueConstraint(name: 'package_name_idx', columns: ['name'])]
 #[ORM\Index(name: 'indexed_idx', columns: ['indexedAt'])]
 #[ORM\Index(name: 'crawled_idx', columns: ['crawledAt'])]
-#[ORM\Index(name: 'dumped_idx', columns: ['dumpedAt'])]
 #[ORM\Index(name: 'dumped2_idx', columns: ['dumpedAtV2'])]
```
glaubinix: do you still need this index? Think this is covered by `dumped2_requested_crawled_frozen_idx`

src/Controller/PackageController.php
```diff
@@ -1189,14 +1189,17 @@ public function abandonAction(Request $request, #[MapEntity] Package $package, #
         $form->handleRequest($request);
         if ($form->isSubmitted() && $form->isValid()) {
             $package->setAbandoned(true);
-            $package->setReplacementPackage(str_replace('https://packagist.org/packages/', '', (string) $form->get('replacement')->getData()));
+            // '' means "no replacement" here; setReplacementPackage() maps it to null anyway, but being
+            // explicit keeps the stored value obviously identical to the null an upstream
+            // `"abandoned": true` yields, which is what the Updater reconciles against
+            $replacement = str_replace('https://packagist.org/packages/', '', (string) $form->get('replacement')->getData());
+            $package->setReplacementPackage($replacement === '' ? null : $replacement);
```
glaubinix: This change seems unnecessary. `Package::setReplacementPackage` already takes care of it

src/Entity/Package.php
```diff
@@ -693,24 +692,76 @@ public function getIndexedAt(): ?\DateTimeImmutable
         return $this->indexedAt;
     }
 
-    public function setDumpedAt(?\DateTimeImmutable $dumpedAt): void
+    /**
+     * Only V2Dumper legitimately records a dump time, and it does so in bulk SQL — this setter exists
+     * for tests. Production code goes through markForDump() / forceDump(), never writing either
+     * timestamp by hand.
+     */
+    public function setDumpedAtV2(?\DateTimeImmutable $dumpedAt): void
```
glaubinix:
```suggestion
    public function setDumpedAtV2(\DateTimeImmutable $dumpedAt): void
```
Never called with `null` and would avoid setting this to `null` by accident

migrations/2026_08_package_dump_requested_at.sql
Seldaek: Yeah that was a claude fuckup, fixed

glaubinix (approved):

glaubinix (approved):

PR #1812, opened 2026-08-22
Replace nvd3 charts with Apache ECharts, improve stats

IonBazan:
## Summary

The stats pages (package installs, PHP versions, global stats) currently render their charts with **nvd3 1.8.6 on d3 v3**. Both libraries are effectively abandoned: nvd3's last release was in ~2017 and it is pinned to a 2014-era d3, which shows both visually and maintenance-wise.

This PR replaces the charting stack with [**Apache ECharts**](https://echarts.apache.org/en/index.html), keeping all existing behavior:

- Line charts with the focus/zoom date picker, now powered by ECharts' `dataZoom` (inside + slider)
- 100%-stacked area charts for PHP version statistics (percentages computed client-side), with the existing PHP version color map
- Per-series colors and K/mio value formatting preserved
- Monthly-averaged charts use a category x-axis so zooming in no longer produces repeated identical month labels
- Tooltip dates match chart granularity (no day number for monthly data)

## New features

- "Releases over time" heat map on the stats page showing tagged releases per month, to help users pick actively maintained packages:

image

- Sparkline on the package details page visualizing daily installs (weekly avg) over the past year:

image

## Fixes

- Selecting a very small time frame caused unnecessary repeated ticks to be rendered:

Before:
image

Fixed:
image

- Packages with "0 downloads" used to show an evenly distributed chart:

Before:
image

Now:
image

- The per-version installs chart no longer follows page scrolling

## Charts

Installs:
image

PHP versions:
image

image

## Details

- `d3` and `nvd3` removed from package.json; `echarts` added with tree-shakeable `echarts/core` imports (~654 kB min / ~200 kB gzipped for the standalone charts bundle)
- Chart containers changed from `<svg>` to `<div>` in templates (ECharts renders into div containers); obsolete SVG fallback text and unused `data-scale` attributes dropped
- nvd3-specific CSS hacks removed from `css/charts.css`
- Public JS API unchanged: `initPackageStats()`, `initPhpStats()`, `initGlobalPhpStats()` and the `[data-labels]` auto-initializer


IonBazan:
Rebased and re-tested.

EdouardCourty:
The PHP versions colors seem too bright to be nice to read IMO, I'd use less bright colors

Seldaek (commented):
Went through this one closely — the ECharts migration itself looks clean, and dropping the d3 v3 / nvd3 stack is very welcome. The release heatmap is a nice addition. Comments inline; two of them are real bugs worth fixing before merge, the rest are smaller.

### Worth fixing before merge

- **`js/charts.js:624`** — the heatmap tooltip reports the year mirrored about the middle row. The axis labels are correct, so this only shows on hover.
- **`js/charts.js:133`** — when a version has no data the guard returns before the re-init, leaving the *previous* version's chart on screen under the new label. Reachable via the all-zero-series filtering in `versionPhpStatsAction`.

### Worth a look

`js/charts.js:248` (unescaped `seriesName` in tooltip HTML — likely pre-existing from nvd3, but easy to close now), `js/charts.js:116` (the `dispose` listener never fires, so instances leak), `js/charts.js:282` (inside-zoom captures the mouse wheel and hijacks page scroll), `js/package_downloads.js:5` (second copy of jQuery bundled onto the package page), `js/package_downloads.js:60` (sparkline label off by 7x), `src/Controller/PackageController.php:1274` (`releaseChart` leaking into the public `stats.json`), and the `<p class="row">` wrappers in both `web/*_stats.html.twig` files.

### Two things with no line in the diff to hang a comment on

- **`overallStatsAction`** (`src/Controller/PackageController.php:1634`) has neither `blockAbusers($req)` nor the `Killswitch::DOWNLOADS_ENABLED` check that `statsAction` carries at `:1243-1249`. That was fine while it was only reachable from the stats page; now the sparkline calls it from every package page view. Worth adding both, so the killswitch actually stops the work rather than only hiding the container.
- **`PHP_VERSION_COLORS`** (`js/charts.js:51-68`) has no entry for 8.5, so that series falls through to `MULTI_COLORS[0]` = `#f28d1a` — the brand orange, sitting right next to `'5.3': '#ee8e33'`. Pre-existing, but cheap to fix while the file is open.

### Nits, take or leave

`formatDigit` and `formatDay` are byte-identical copies across `charts.js` and `package_downloads.js` (a shared `js/chartUtils.js` would inline into both); `statistics.last_partial` is reused under the heatmap (`templates/package/stats.html.twig:87`) where "the last data point" reads oddly for a grid of cells; the `position: relative` in `css/charts.css` and at `templates/package/stats.html.twig:44` only existed to anchor the now-removed sticky scroll handler; and `formatDigit` uses `>` rather than `>=`, so exactly 1000000 renders unformatted (carried over verbatim, not new).

### Checked and fine

`echarts.graphic.LinearGradient` and `echarts.format.encodeHTML` are both exported from `echarts/core`; esbuild 0.15.18 with `platform: browser` auto-defines the `process.env.NODE_ENV` that echarts' lib build needs; `{% block scripts %}` in `view_package.html.twig` doesn't shadow any parent content; the `[category, value]` series-data form works on ordinal axes; `average=weekly` produces `Y-m-d` labels that `package_downloads.js` parses correctly; and `getPartialPackageWithVersions` returns a fully hydrated `Package`, so `getReleasedAt()`/`isDevelopment()` are safe on the stats path.

js/charts.js
```diff
@@ -406,7 +555,132 @@ import '../css/charts.css';
...
+        var chart = echarts.init(el);
+        chart.setOption({
+            animation: false,
+            tooltip: {
+                backgroundColor: 'rgba(255, 255, 255, 0.96)',
+                borderColor: '#ddd',
+                textStyle: {color: '#2d2d32', fontSize: 12},
+                formatter: function (params) {
+                    var value = params.value;
+                    var title = MONTHS_LONG[value[0]] + ' ' + years[years.length - 1 - value[1]];
```
Seldaek: **The tooltip reports the wrong year.**

`years` is built newest-first at `:586-589`, and the cell's y-index at `:606` is already an index into that same array. Re-mirroring it here inverts the row.

Worked example with `startYear=2020, endYear=2022`, so `years = [2022, 2021, 2020]`:

- the 2020 cells are pushed as `[month, years.length - 1 - yearIndex, count]` = `y = 2`
- `yAxis.data[2] === "2020"`, so the **axis label is correct**
- the tooltip computes `years[3 - 1 - 2]` = `years[0]` = **2022**

Every tooltip is mirrored about the middle row. It's only accidentally right for a single-year package, or the middle row of an odd-length range.

```suggestion
                    var title = MONTHS_LONG[value[0]] + ' ' + years[value[1]];
```

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+        instances.forEach(function (chart) {
+            chart.resize();
+        });
+    });
+
+    function initPackagistChart(el, labels, series, withDatePicker, type, colorMap) {
+        type = type || 'line';
+        colorMap = colorMap || null;
+
+        if (!el || !labels.length || !series.length) {
```
Seldaek: **An empty result leaves the previous chart on screen.**

This guard returns before the dispose/re-init at `:294-300`, so when a version has no data the old chart stays rendered and the user reads another version's numbers under the label they just clicked. nvd3 drew an empty chart here instead.

This is reachable, not theoretical — `versionPhpStatsAction` unsets every all-zero series (`src/Controller/PackageController.php:1439-1447`) and then trims trailing zero datapoints (`:1449-1459`), so `values: {}` / `labels: []` with a 200 is a normal response.

Clearing the existing instance first is enough:

```js
var existing = echarts.getInstanceByDom(el);
if (!el || !labels.length || !series.length) {
    if (existing) { existing.clear(); }
    return;
}
```

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+        }
+        return value;
+    }
+
+    const instances = new Set();
+
+    function trackInstance(chart) {
+        instances.add(chart);
+
+        chart.getZr().on('dispose', function () {
```
Seldaek: zrender never emits a `dispose` event, so this listener never fires and nothing is ever removed from `instances`.

Checked against the pinned version — `zrender@6.1.0/lib/zrender.js:291-306` tears everything down without triggering anything, and `zr.on()` just proxies to `this.handler.on()`:

```js
ZRender.prototype.dispose = function () {
    if (this._disposed) { return; }
    this.animation.stop();
    this.clear();
    this.storage.dispose();
    ...
    this._disposed = true;
    delInstance(this.id);
};
```

So every version-toggler click disposes an instance at `:294-297` and adds a fresh one, while the dead one is retained along with its series data and DOM reference. Click through 30 versions on the PHP stats page and you retain 30 dead instances. No console noise, because `ECharts.prototype.resize` bails on `this._disposed`.

The tidiest fix also removes the canvas teardown/rebuild on every click:

```js
var chart = echarts.getInstanceByDom(el) || trackInstance(echarts.init(el));
chart.setOption(option, true); // notMerge
return chart;
```

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+                    var title = params[0].axisValueLabel || format(params[0].axisValue);
+                    var html = title + '<br>';
+                    html += params.map(function (param) {
+                        var value = Array.isArray(param.value) ? param.value[1] : param.value;
+                        if (type === 'area') {
+                            value = (isFinite(value) ? value.toFixed(1) : value) + '%';
+                        } else {
+                            value = formatDigit(value);
+                        }
+                        return param.marker + ' ' + param.seriesName + ': <strong>' + value + '</strong>';
```
Seldaek: `seriesName` goes into tooltip HTML unescaped. ECharts assigns a string returned from `tooltip.formatter` via `innerHTML` — it does not escape it.

For the version chart that name is `$version->getVersion()` (`src/Controller/PackageController.php:1701`), the VCS pretty version, sanitized only by `ConsoleIO::sanitize($str, false)` (`src/Package/Updater.php:753`) — which strips ANSI escapes and control characters, **not HTML**. Git ref names permit `<`, `>`, `"` and `=`, so a branch named `` becomes `dev-` and fires when anyone hovers that package's stats chart.

Very likely pre-existing (nvd3's default `contentGenerator` also concatenated `key` into an HTML string), but the rewrite is a good moment to close it. `echarts.format.encodeHTML` is exported from `echarts/core` (`echarts@6.1.0/lib/export/api.js:75-76`):

```js
return param.marker + ' ' + echarts.format.encodeHTML(param.seriesName) + ': <strong>' + value + '</strong>';
```

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+                max: type === 'area' ? 100 : null,
+                splitLine: {lineStyle: {color: '#eee'}}
+            },
+            series: chartSeries
+        };
+
+        if (withDatePicker) {
+            option.dataZoom = [
+                {
+                    type: 'inside',
```
Seldaek, line 273: `zoomOnMouseWheel` defaults to `true`, so this captures the wheel. These charts are full-width and 300-400px tall (`css/charts.css:1-8`), so scrolling past a stats page with the cursor over a chart zooms the chart instead of scrolling the page — with no way to reset short of a reload.

nvd3's `lineWithFocusChart` never captured the wheel. Either scope it to a modifier, or drop the inside zoom and let the slider do the work (which reproduces the old behaviour exactly):

```js
{type: 'inside', throttle: 50, zoomOnMouseWheel: 'ctrl', moveOnMouseWheel: false}
```

js/package_downloads.js
```diff
@@ -0,0 +1,116 @@
...
+                        trigger: 'axis',
+                        axisPointer: {type: 'line'},
+                        backgroundColor: 'rgba(255, 255, 255, 0.96)',
+                        borderColor: '#ddd',
+                        textStyle: {color: '#2d2d32', fontSize: 12},
+                        formatter: function (params) {
+                            params = Array.isArray(params) ? params : [params];
+                            var value = params[0].value;
+
+                            return formatDay(value[0]) + '<br><strong>' + formatDigit(value[1]) + '</strong> avg. weekly installs';
```
Seldaek: This label is off by a factor of 7 — it's average installs **per day**, not per week.

The request is `?average=weekly` (`templates/package/view_package.html.twig:462`), and `computeStats` averages *within* the bucket rather than summing it (`src/Controller/PackageController.php:1718`):

```php
$series[$seriesName][] = ceil($value / \count($values));  // $values = the 7 days in this bucket
```

The stats page words the same number correctly: "Daily installs, averaged weekly" (`stats.daily` + `stats.averaged`). Something like `'</strong> installs/day (weekly avg.)'` would match. The PR description has the same slip.

js/package_downloads.js
```diff
@@ -0,0 +1,116 @@
+import * as echarts from 'echarts/core';
+import { LineChart } from 'echarts/charts';
+import { GridComponent, TooltipComponent } from 'echarts/components';
+import { CanvasRenderer } from 'echarts/renderers';
+import jQuery from 'jquery';
```
Seldaek: This bundles a second copy of jQuery onto the package page.

`esbuild.js` has no `splitting`, so each entry point is a self-contained IIFE — this file ships echarts core + LineChart + Grid + Tooltip + CanvasRenderer **and its own jQuery**, on the page that already loads `app.js`, which contains jQuery and publishes it globally (`js/jquery.js:2` — `window.jQuery = window.$ = jQuery`).

It's used for exactly three things: `$(selector)[0]`, `$.ajax`, and `$(window).on('resize')`. Either drop the import and rely on the global (`layout.html.twig` loads `app.js` before `{% block scripts %}`, so it's there), or go native with `querySelector` + `fetch` + `addEventListener`. Roughly 90 KB min off the site's highest-traffic page.

One behavioural side effect worth being deliberate about: a separate jQuery instance means `app.js`'s global `$.ajaxSetup` error handler doesn't apply, so a failing sparkline request is silent. Probably what you want here, but it's a change.

src/Controller/PackageController.php
```diff
@@ -1257,11 +1257,21 @@ public function statsAction(Request $req, string $name): Response
...
+            $month = $version->getReleasedAt()->format('Y-m');
+            $releaseChart[$month] = ($releaseChart[$month] ?? 0) + 1;
+        }
+        ksort($releaseChart);
         $data = [
             'downloads' => $this->downloadManager->getDownloads($package),
             'versions' => $versions,
             'average' => $this->guessStatsAverage($date),
             'date' => $date->format('Y-m-d'),
+            'releaseChart' => $releaseChart,
```
Seldaek: `releaseChart` is added to `$data` before the JSON branch at `:1277`, so it now ships in `GET /packages/[vendor]/[package]/stats.json` as well.

That endpoint is documented with an explicit response shape (`templates/api_doc/index.html.twig:341-356`) that this PR doesn't update, and it's a high-traffic anonymous API. For an old package this adds ~170 `"YYYY-MM": n` entries to every response.

Simplest fix is to build it below the `if ($req->getRequestFormat() === 'json') { ... return ...; }` block, next to `$data['package'] = $package;` — the heatmap is a web-view concern. (Or keep it and document it, but the payload cost argues for moving it.)

templates/web/stats.html.twig
```diff
@@ -10,27 +10,21 @@
             <h3>{{ 'statistics.over_time'|trans }}</h3>
 
             <p class="row">
```
Seldaek: `<div>` inside `<p>` is invalid, and the browser silently restructures it.

`<svg>` is phrasing content so it was legal here; `<div>` is not, so the parser **auto-closes the `<p>` at the opening `<div>` tag**. The rendered DOM becomes:

```html
<p class="row"></p><div class="chart col-12">...</div><p></p>
```

The column loses its flex row parent (no negative gutter compensation) and you get two stray empty paragraphs picking up `margin-bottom`. Changing the wrapper to `<div class="row">` fixes it.

Same at `:19` and `:26`.

templates/web/php_stats.html.twig
```diff
@@ -14,9 +14,7 @@
             <h3>Package installations by PHP minor version, daily</h3>
 
             <p class="row">
```
Seldaek: Same `<p class="row">` wrapping a `<div>` as in `web/stats.html.twig` — the parser auto-closes the `<p>`, so the column loses its row parent. `<div class="row">` instead.

Also at `:26`.

js/charts.js
```diff
@@ -406,7 +555,132 @@ import '../css/charts.css';
...
+        var MONTHS_SHORT = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+        var MONTHS_LONG = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+
+        var keys = Object.keys(counts).sort();
+        if (!keys.length) {
+            return;
+        }
+
+        var now = new Date();
+        var startYear = parseInt(keys[0].split('-')[0], 10);
```
Seldaek, line 577: `startYear` comes straight from `releasedAt`, which is the VCS tag/commit date and isn't clamped anywhere on this path — so it drives `el.style.height` at `:592` without a floor.

Compare `guessStatsStartDate` (`src/Controller/PackageController.php:2002`), which floors at `2012-04-13`. A single tag carrying a 1970 or 2001 date here produces a 57-row / ~2000px heatmap that's almost entirely empty cells. Worth clamping to something like `max(package.createdAt, 2012)`, server- or client-side.

js/charts.js
```diff
@@ -406,7 +555,132 @@ import '../css/charts.css';
...
+        var el = $(selector)[0];
+        if (!el || !counts) {
+            return;
+        }
+
+        var MONTHS_SHORT = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+        var MONTHS_LONG = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+
+        var keys = Object.keys(counts).sort();
+        if (!keys.length) {
```
Seldaek, line 571: This bail (and the one at `:582`, for a first release dated later this year than today) leaves the section half-rendered: the template has already emitted `<div class="chart js-release-stats">`, which `css/charts.css:1-5` sizes to `height: 300px`, so you get a heading followed by a blank 300px gap.

Worth either hiding the container or rendering a "no releases yet" state.

src/Controller/PackageController.php
```diff
@@ -1257,11 +1257,21 @@ public function statsAction(Request $req, string $name): Response
         $versions = $package->getVersions()->toArray();
         usort($versions, Package::class.'::sortVersions');
         $date = $this->guessStatsStartDate($package);
+        $releaseChart = [];
+        foreach ($versions as $version) {
+            if ($version->isDevelopment() || null === $version->getReleasedAt()) {
```
Seldaek: `$package->getVersions()` is unfiltered, so soft-deleted versions are in this collection — `templates/package/version_list.html.twig:5` renders them with a marker rather than omitting them.

A version pulled for malware still contributes to the "actively maintained" signal the heatmap is meant to convey.

```suggestion
            if ($version->isDevelopment() || $version->isSoftDeleted() || null === $version->getReleasedAt()) {
```

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+                    type: 'line',
+                    stack: 'total',
+                    large: true,
+                    smooth: false,
+                    symbol: 'none',
+                    emphasis: {focus: 'series'},
+                    lineStyle: {width: 0},
+                    areaStyle: {opacity: 1.0},
+                    data: xValues.map(function (x, index) {
+                        var total = totals[index];
```
Seldaek: Percentages are precomputed here against a fixed `totals` array, so legend toggling no longer re-normalizes the 100%-stacked chart.

nvd3's `chart.stacked.style('expand')` recomputed the expansion on whatever was visible, so hiding a PHP version re-expanded the rest to fill 100%. Now hiding a series just leaves a gap, and `yAxis.max: 100` (`:273`) keeps the empty space at the top.

Minor, but the legend is interactive and people will click it.

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+                    data: xValues.map(function (x, index) {
+                        return [x, parseInt(serie.values[index], 10) || 0];
+                    }),
+                };
+            });
+        }
+
+        // name-based lookup first, then positional palette
+        colors = series.map(function (serie, index) {
+            if (colorMap && colorMap[serie.name] !== undefined) {
```
Seldaek: `colorMap` is sometimes an Array (`DEFAULT_COLORS`, `MULTI_COLORS`) and sometimes a plain object (`PHP_VERSION_COLORS`), and `serie.name` is externally derived (a version string). So this lookup can hit prototype/array members: `colorMap['length']` returns `2` and `colorMap['constructor']` returns a function, either of which gets handed to ECharts as a colour.

```js
if (colorMap && Object.prototype.hasOwnProperty.call(colorMap, serie.name)) {
```

The dual semantics of the single parameter — name-keyed map at `:201` *or* positional palette at `:204` — is worth splitting into two explicit arguments while you're here.

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+        // render as repeated identical month-only labels)
+        var xValues = labels.map(function (label) {
+            if (!isDaily) {
+                return label;
             }
+            var parts = label.split('-');
+            return new Date(+parts[0], parts[1] - 1, +parts[2]).getTime();
+        });
+
+        var colors = [];
```
Seldaek: Dead — unconditionally overwritten by the `series.map(...)` assignment at `:200`.

js/charts.js
```diff
@@ -42,90 +78,231 @@ import '../css/charts.css';
...
+                axisPointer: {type: 'line'},
+                backgroundColor: 'rgba(255, 255, 255, 0.96)',
+                borderColor: '#ddd',
+                textStyle: {color: '#2d2d32', fontSize: 12},
+                formatter: function (params) {
+                    if (!Array.isArray(params)) {
+                        params = [params];
+                    }
+
+                    var title = params[0].axisValueLabel || format(params[0].axisValue);
```
Seldaek: The `|| format(...)` fallback is unreachable: `axisValueLabel` is always populated (the category name for ordinal axes, `TimeScale.getLabel` output for time axes).

Related and more visible: at deep zoom levels ECharts' time formatter drops to hour/minute precision, so a daily chart's tooltip title can read `2026-09-03 04:00`. Using `format(params[0].axisValue)` unconditionally when `isDaily` would be both simpler and more stable.

js/charts.js
```diff
@@ -406,7 +555,132 @@ import '../css/charts.css';
...
         );
     };
+
+    window.initReleaseStats = function (selector, counts) {
+        var el = $(selector)[0];
+        if (!el || !counts) {
+            return;
+        }
+
+        var MONTHS_SHORT = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
```
Seldaek: These are hardcoded English while the surrounding pages go through `|trans` — same for `' release'`/`' releases'` at `:625`, and `'avg. weekly installs'` in `js/package_downloads.js`.

Only `messages.en.yml` exists today so nothing breaks right now, just flagging it as a translation gap being introduced.

js/package_downloads.js
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                var values = res.values[seriesName] || [];
+                var points = res.labels.map(function (label, index) {
+                    var parts = label.split('-');
+                    var date = new Date(+parts[0], parts[1] - 1, +parts[2]);
+                    return [date.getTime(), parseInt(values[index], 10) || 0];
+                });
+
+                var chart = echarts.init(el);
```
Seldaek: No `getInstanceByDom` re-init guard here, unlike the sibling at `js/charts.js:294-297`, and the resize handler at `:102` is registered inside the AJAX callback so repeat calls would stack listeners.

Latent with only one call site today, but it diverges from `charts.js` for no particular reason.

IonBazan:
Fixed all reported issues. Worth highlighting:
- The heatmap floor is the 2012 stats start date rather than the package creation date, so packages imported late keep their real release history.
- The killswitch and abuser checks live in computeStats, which also covers the version and major-version stats endpoints, not only overallStatsAction.

Seldaek:
> * The heatmap floor is the 2012 stats start date rather than the package creation date, so packages imported late keep their real release history.

Releases have nothing to do with stats collection tho so not sure if there should be a floor. But on the other hand Composer releases cannot be older than 2011 when we started the project so I guess using 2011-01-01 as floor would be ok?

js/charts.js
```diff
@@ -401,12 +539,138 @@ import '../css/charts.css';
...
+        initPackagistChart($(selector)[0], res.labels, series, {
+            withDatePicker: true,
+            type: 'area',
+            colorMap: PHP_VERSION_COLORS
+        });
+    };
+
+    function monthNames(style) {
+        var locale = document.documentElement.lang || 'en';
+        var formatter = new Intl.DateTimeFormat(locale, {month: style});
```
Seldaek: I think that's enough really

```suggestion
        var formatter = new Intl.DateTimeFormat('en', {month: style});
```

Or you could even hardcode english names.. I don't think we're going to translate the whole website anytime soon :D

js/charts.js
```diff
@@ -42,90 +86,224 @@ import '../css/charts.css';
...
+                    return [x, total ? Math.round(values[index] * 1000 / total) / 10 : 0];
+                });
+            });
+        }
 
-            nv.utils.windowResize(chart.update);
+        var chartSeries;
+        if (type === 'area') {
+            var initialData = areaSeriesData(series.map(function () { return true; }));
+            chartSeries = series.map(function (serie, index) {
```
Seldaek: serie should be series in the function as well technically, as series is both the singular and plural form of the word.

src/Controller/PackageController.php
```diff
@@ -2023,6 +2033,28 @@ private function createDatePoints(\DateTimeImmutable $from, \DateTimeImmutable $
...
 
+    /**
+     * @param Version[] $versions
+     *
+     * @return array<string, int>
+     */
+    private function computeReleaseChart(array $versions): array
+    {
+        $releaseChart = [];
+        $statsRecordDate = new \DateTimeImmutable(self::STATS_RECORD_DATE);
```
Seldaek: Per my other comment

```suggestion
        $statsRecordDate = new \DateTimeImmutable('2011-01-01 00:00:00');
```

IonBazan: Used a const, set it to `2011-01-01 00:00:00` as suggested. Technically there could be older packages like `symfony/symfony`.

js/charts.js
```diff
@@ -28,9 +65,17 @@ import '../css/charts.css';
         '8.2': '#111192',
         '8.3': '#5a4ac1',
         '8.4': '#4a5ce7',
+        '8.5': '#1f77b4',
+        '8.6': '#2f8fd6',
```
IonBazan, line 69: 8.5 is taken from currently rendered colour on packagist.org (it renders without being defined enywhere).
8.6 is added by me - feel free to change to something else if needed

IonBazan:
Also question: Would it be good for this Sparkline to be clickable - redirecting to download stats?
> image

Seldaek:
Nah leave it as is I'd say, clicking on a graph shouldn't change pages IMO

src/Controller/PackageController.php
Seldaek: Of course things existed before Composer but their releases are not visible to Packagist as they had no composer.json, so symfony/symfony's oldest release on packagist for example is 2.0.4 at 2011-10-04 11:47 UTC.

Seldaek: Meaning anything older than 2011 was clearly backdated and/or possibly buggy, and shouldn't be graphed IMO

Seldaek:
Alrighty thanks, another good long-overdue cleanup :)

IonBazan:
I think I missed this when testing- happens on mobile:
image

Seldaek:
Just deployed with a couple more fixes https://github.com/composer/packagist/commits?author=Seldaek&since=2026-09-04&until=2026-09-04 - if you still see issues on mobile feel free to send another PR.

PR #1813, opened 2026-08-23
Migrate frontend from Bootstrap 3 to Bootstrap 5

IonBazan:
## Summary

Upgrades the frontend from Bootstrap 3 to Bootstrap 5. Glyphicons are replaced with Bootstrap Icons, forms now use Symfony's official `bootstrap_5_layout.html.twig` instead of a hand rolled BS3 theme, and modals, navbar, badges, checkboxes and grid markup are updated to their BS5 equivalents.

This is a large, mechanical diff across most templates plus `app.scss` and the JS entrypoints. It is not realistically reviewable line by line, so the practical way to review this is to check it out and click around, comparing it against production. The goal was to keep the look and feel as close to current production as possible, not to redesign anything.

A few notes on approach:

- Grid breakpoints were remapped rather than renamed 1:1 (BS3 `xs/sm/md/lg` map to BS5 `none/md/lg/xl`) so that content keeps stacking at the same screen widths it does today.
- In a couple of spots (footer layout, mainly) the original bespoke breakpoint values were not preserved exactly. Instead they were simplified to plain, common Bootstrap column patterns, since matching the old custom breakpoints pixel for pixel added complexity for no real benefit and made the CSS harder to maintain going forward.
- Colors, fonts and spacing are unchanged. Where BS5 dropped implicit defaults BS3 used to provide (link underlines, `pre` padding, form control heights, grid column `position: relative`), those are restored explicitly rather than left to accident.

Changes are not pixel perfect and small differences are expected here and there, but the overall look and feel should be very close to what is currently live.

## Screenshots

image

image

image

image

image

image

Also updated search results and filters:
image

Replaces #1462

## Test plan

- [x] Check out the branch and click through the main pages (home, search, package page, login/register, profile, org pages) and compare against production
- [x] Confirm forms (login, register, maintainer add/remove, 2FA) submit correctly
- [x] Check mobile nav and responsive behavior at a few widths


Seldaek:
Thanks, this sounds great on paper (i.e. the PR description). I'll try and check it out soonish but got slightly more pressing matters.

IonBazan:
Thanks, glad you find this helpful. Take your time to review it. In the meantime, I will try pushing minor improvements here as I find them. One difference I noticed is different `.container` widths between BS versions but I think this is not something worth changing as user experience is pretty much the same.

If you have any urgent tasks, feel free to create issues, I might have some time and token quota 👾 to help with them ❤️

Seldaek:
Rebasing and checking if https://github.com/composer/packagist/pull/1802 creates new stuff to be migrated would be a good start. There's more of that coming so that's why I'd rather get this one merged as well quickly if possible to avoid further conflicts.

IonBazan:
Rebased and updated templates.

I can see that #1814 could also introduce some conflicts, quite easy to fix:
1. **`templates/package/version_list.html.twig`** - the relocated hide-version form uses `glyphicon`/`data-toggle`; swap for `bi bi-eye-slash-fill` + `data-bs-toggle`/`data-bs-placement`/`data-bs-container`, then drop the old duplicate copy it left behind.
2. **`js/view.js`** - `applyVersionDeleteResponse()` reuses `glyphicon`/`data-original-title`; swap for `bi`/`bi-trash-fill`, and since BS5 tooltips cache their title at construction, update a live tooltip via `Tooltip.getInstance(el)?.setContent({'.tooltip-inner': title})` instead (needs importing `Tooltip` alongside `Modal`).

Seldaek:
Sorry for the data dump, but I figured if you have time and are willing to go through it, you can already keep moving forward before I have time to review further. I have not verified any of these claims, just a deep review by Claude:

Deep review of PR #1813 (Bootstrap 3 → 5 migration) is done. The mechanical conversion is thorough — I grepped the whole tree and found zero leftover `col-xs-*`, `pull-*`, `text-right`, `label-*`, `visible-*`/`hidden-*`, `panel`, `well`, `data-toggle`, or `glyphicon` in templates/JS. What follows is where the conversion changed behaviour rather than pixels.

## Fix before merge

**1. `js/view.js:247-254` — `setDeletionTooltip` still uses the BS3 jQuery tooltip API (High)**

```js
if ($el.data('bs.tooltip')) { $el.tooltip('fixTitle'); } else { $el.tooltip({...}); }
```

BS5 keeps instances in an internal `WeakMap`, so `$el.data('bs.tooltip')` is never truthy — the update branch is dead. When `applyVersionDeleteResponse()` refreshes a row that already had a `.deletion-alert`, it falls to the `else`, where `getOrCreateInstance()` returns the *existing* instance and ignores the new config: the tooltip keeps the old title, and the raw `title` attribute set on the line above stays on the element, so the browser renders a duplicate native tooltip too. And `fixTitle` doesn't exist in BS5 — if the guard ever went truthy it'd throw inside the ajax callback, skipping the `.delete-version, .hide-version` removal on the next line. Fix is to use `Tooltip.getInstance()` + `dispose()` + `new Tooltip()` directly, and give the JS-created badge the same `data-bs-*` attributes as `version_list.html.twig:42`.

**2. `templates/user/public_profile.html.twig:156` — admin freeze modal lost its label text (High)**

`form_label()` on a checkbox resolves to `checkbox_radio_label`, which is guarded by `{% if widget is defined %}` and only renders from *inside* `checkbox_widget`. So that call emits nothing and `"Also act on the user's packages"` is discarded; the visible label comes from the form type instead, and `FreezeType.php:49` sets no `label` — so the modal now reads **"Freeze packages"**. Put the label on the form type and drop the no-op call. Same dead pattern in `user/edit.html.twig:36-40`, benign there because `ProfileFormType` sets the label, but it produces a nested `.form-check` and a duplicated `form-check-input`.

## Follow-up pass (all one-or-two-line fixes)

- **`registration/check_email.html.twig:28-45`** — `form-horizontal` doesn't exist in BS5 (it's a separate form theme) and `.form-group` isn't a grid row, so `col-md-3`/`col-md-9` stack at 25%/75% instead of sitting side by side.
- **`admin/filter_list/index.html.twig:23-46`** — `d-flex flex-wrap` isn't a drop-in for `form-inline`: BS3 set `.form-control { width: auto }`, BS5 doesn't, so all four controls wrap one per line. Needs `w-auto`. Milder in `admin/frozen_users.html.twig:19-27`.
- **`.btn-xs` and `.dl-horizontal`** were removed in BS4/5 and aren't shimmed — 8 buttons render full-size (organization teams/members, filter list) and 6 definition lists now stack (package/web stats, organization show, public profile). Note `.btn-group-xs .btn` *is* still in app.scss:985, which is why invitations look right.
- **`.form-group` lost its `margin-bottom`** (app.scss:786 keeps only `clear: left`); 39 hand-written wrappers remain and the PR converted only some to `mb-3`. Simplest fix is restoring the 15px in app.scss.
- **`css/app.scss:2144-2175`** — the `.badge.bg-*` brand-colour overrides are dead, since BS5's `.bg-*` utilities use `!important`. All converted state badges render in stock Bootstrap colours. `bg-primary` is the exception (resolves via `--bs-primary-rgb`), which is why search chips are orange while `.btn-primary` stays blue — that matches production, but the `:root` comment there is misleading enough to invite a future "fix" that repaints every primary button.
- Lower priority: `js/notifier.jsx:41` still uses BS3 `.close` (unstyled ×), 7 hand-written `<select class="form-control">` should be `form-select`, `.help-block` is only styled under `.audit-log-filters`, footer social icons are mirrored by the float→flex change, and `js/search.js:208` interpolates `refinement.label` into `innerHTML`/`data-value` unescaped — pre-existing, CSP blocks script execution, but the line is being rewritten anyway.

Also worth a look eventually: three tests now accept BS3 *or* BS5 selectors (`.packages .label, .packages .badge`), and since BS3 is gone the legacy halves are dead weight that weakens the assertions.

IonBazan:
Addressed the review feedback:

**Fix before merge**

1. `setDeletionTooltip` (js/view.js) now uses the Bootstrap 5 `Tooltip` API. It disposes any existing instance and rebuilds it, writes the text to `data-bs-title` instead of the native `title` attribute, and the JS-created `.deletion-alert` badge gets the same `data-bs-*` attributes as the Twig template.
2. The freeze modal and profile edit checkboxes render their labels again. The label text moved onto the form types (`FreezeType`, `ProfileFormType` already had it), the no-op `form_label()` calls were removed, and the redundant hand-written `.form-check` wrappers were deleted so `checkbox_widget` renders a single input plus label.

**Follow-up pass**

* `check_email.html.twig`: `form-horizontal` replaced with a Bootstrap 5 grid row (`.row` + `.col-form-label` + `mb-3`).
* Inline admin filter forms: `form-control` selects changed to `form-select`, `w-auto` added so controls sit on one line.
* `.btn-xs` and `.dl-horizontal` shimmed in `app.scss`. `.form-group` bottom margin restored.
* `.badge.bg-*` brand colour overrides now use `!important` so they win over Bootstrap 5's utility classes. Clarified the misleading `--bs-primary` comment.
* Added a global `.help-block` rule (was only styled inside `.audit-log-filters`).
* Footer social icon order restored after the float to flex change.
* Bootstrap 3 `<small>` scale restored: 85% base, 65% inside h1 to h3, 75% inside h4 to h6.
* Notifier close control styled, 7 `<select class="form-control">` converted to `form-select`, `refinement.label` escaped in the search current-refinements markup.

## Additional changes

**BS3 leftovers removed**

* `.navbar-collapse.in` changed to `.navbar-collapse.show` (Bootstrap 5 toggles `.show`).
* Removed the dead `.sortable` order-by rules, including `.sortable .glyphicon.active`. That whole sort control has no markup since search moved to Algolia InstantSearch, so swapping `.glyphicon` for `.bi` would just be a differently named dead rule.
* Full grep of templates, JS, and CSS turned up no other BS3 grid, utility, component, or `data-toggle` usage. `.btn-xs`, `.btn-group-xs`, and `.dl-horizontal` are kept intentionally as custom shims.

**Tests**

* `PackageControllerTest`: dropped the dead `.packages .label` half of the selector.
* Added regression assertions in `UserControllerTest` and `ProfileControllerTest` that both checkbox labels render with the expected text.
* `RegistrationControllerTest` and `IntegrationTestCase` keep `.alert-danger, .invalid-feedback` since both are valid Bootstrap 5 form-theme output (root form vs per field errors), not BS3 residue.

Seldaek:
Thanks for the updates. Please do not touch the branch. I got local fixes I'm working on and then plan on merging it once done.

Seldaek:
Ok I think it's ready to go, just got a couple things to check (mainly pagination I think), will merge & deploy tomorrow so I am at able to react if it breaks anything :)

Seldaek:
Thanks, it is live and from what I can tell it looks overall really solid now at least on the main pages, I only have a couple minor minor nitpicks left to fix.

IonBazan:
Glad I could help and thanks for the fixing the problems I missed - I reviewed them as well.

I noticed that line-height is too small in `pre` blocks like here:
image
Besides of that it looks quite similar to the original design. Will report if anything else draws my attention.

Seldaek:
Fixed the pre line height too.

zoglo:
Absolutely love it @IonBazan <3 Thank you so much

PR #1814, opened 2026-08-24
Allow admins to hide versions that are already soft-deleted by maintainer/updater without recovering them first

Seldaek:
This is to smoothen admin workflows

### DIFF
--- a/js/view.js
+++ b/js/view.js
@@ -239,6 +239,18 @@ const init = function ($) {
         return $(form).closest('.version').find('.version-number').text().trim();
     }
 
+    // Point a deletion tooltip at a new title, creating it when the row was not soft-deleted before.
+    // fixTitle re-reads the native title into data-original-title and blanks it, so the browser
+    // tooltip does not render on top of the Bootstrap one.
+    function setDeletionTooltip($el, title) {
+        $el.attr('title', title);
+        if ($el.data('bs.tooltip')) {
+            $el.tooltip('fixTitle');
+        } else {
+            $el.tooltip({placement: 'top', container: 'body'});
+        }
+    }
+
     function applyVersionDeleteResponse(form, data, deletedToast, softDeletedToast) {
         var row = $(form).closest('.version');
         if (data && data.softDeleted) {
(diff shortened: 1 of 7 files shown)


src/Controller/PackageController.php
```diff
@@ -896,15 +896,24 @@ public function adminHideVersionAction(Request $req, int $versionId, #[CurrentUs
...
         $repo->softDelete($version, VersionDeletionReason::Hidden, $reasonText, $internalReasonText, $user);
         $this->getEM()->flush();
-        $this->getEM()->clear();
 
-        $deletionTitle = 'Hidden by admin on '.gmdate('Y-m-d H:i:s').' UTC'
-            .($reasonText !== null ? ': '.$reasonText : '');
+        // Read before clear(), and from the entity so the date matches what a page reload renders —
+        // softDelete() keeps the original removal time when the version was already soft-deleted.
+        $deletionTitle = $version->getDeletionTitle();
+        $this->getEM()->clear();
```
glaubinix: What's the clear here even for?

src/Controller/PackageController.php
```diff
@@ -896,15 +896,24 @@ public function adminHideVersionAction(Request $req, int $versionId, #[CurrentUs
...
 
         $repo->softDelete($version, VersionDeletionReason::Hidden, $reasonText, $internalReasonText, $user);
         $this->getEM()->flush();
-        $this->getEM()->clear();
 
-        $deletionTitle = 'Hidden by admin on '.gmdate('Y-m-d H:i:s').' UTC'
-            .($reasonText !== null ? ': '.$reasonText : '');
+        // Read before clear(), and from the entity so the date matches what a page reload renders —
+        // softDelete() keeps the original removal time when the version was already soft-deleted.
+        $deletionTitle = $version->getDeletionTitle();
```
glaubinix, line 912: As you are keeping the soft delete date, for cases where an admin hides an already hidden version, this will now read "Hidden by admin on DATE_OF_INITIAL_DELETION".

tests/Controller/PackageControllerTest.php
```diff
@@ -398,6 +398,98 @@ public function testViewPackageVersionHiddenResponseIsNotSharedCached(): void
...
+    /**
+     * Admins can hide a version that is already soft-deleted as gone-from-upstream or
+     * maintainer-pulled; admin-pulled and already-hidden rows must be recovered first.
+     */
+    #[TestWith([null, true, 200])]
+    #[TestWith(['auto_missing', true, 200])]
+    #[TestWith(['maintainer', true, 200])]
+    #[TestWith(['admin', false, 403])]
+    #[TestWith(['hidden', false, 403])]
+    public function testAdminHideVersionAllowedTransitions(?string $reason, bool $buttonShown, int $expectedStatus): void
```
glaubinix: Maybe this to get rid of the magic strings?
```suggestion
    public function testAdminHideVersionAllowedTransitions(?VersionDeletionReason $reason, bool $buttonShown, int $expectedStatus): void
```

js/view.js
```diff
@@ -239,18 +239,22 @@ const init = function ($) {
...
+            // and refresh it rather than leaving the previous icon and title in place.
+            var alert = row.find('.deletion-alert');
+            if (!alert.length) {
+                alert = $('<span class="action-alert deletion-alert"><i class="glyphicon"></i></span>');
                 alert.insertBefore(row.find('form').first());
             }
+            alert.find('i').attr('class', 'glyphicon ' + (data.deletionIcon || 'glyphicon-trash'));
+            var title = data.deletionTitle || 'Deleted';
+            // server-rendered badges have an initialised BS3 tooltip, which reads data-original-title
+            alert.attr('title', title).attr('data-original-title', title);
```
glaubinix: You are updating the tooltip on icon but not on the version

Image

Image

glaubinix:
```suggestion
            alert.attr('title', title).attr('data-original-title', title).tooltip('fixTitle');
```
According to the internet the fixTitle call fixes that the native title is shown as visible in the screenshot in the other comment

src/Controller/PackageController.php
Seldaek: True, probably should override it, audit log is there anyway if we care about detailed timeline

src/Controller/PackageController.php
Seldaek: Looks overkill indeed, but I think it was added because there's a clear() in a similar spot in Updater.

glaubinix (approved):

PR #1815, opened 2026-08-25
Fix null-CVE security advisory matching

### LINKED ISSUE #1709: Security advisories difference score calculation is broken if both the GH and friendsofphp advisory don't have a CVE
See https://github.com/composer/packagist/blob/main/src/Entity/SecurityAdvisory.php#L195 this should probably skip advisories where the CVE is set to `null` to not merge random advisories together.

ousamabenyounes:
## Summary

Fix #1709

- Require a non-null CVE before treating a matching package/CVE pair as the same advisory.
- Add regression coverage proving two distinct CVE-less advisories for one package keep their calculated difference score.

## Test verification (RED → GREEN)

RED — upstream base `66183939` (`main`), production code untouched, only the new test applied:

```
1) App\Tests\Entity\SecurityAdvisoryTest::testCalculateDifferenceScoreDoesNotMatchNullCves
Failed asserting that 0 is identical to 5.
```

GREEN — this branch:

```
OK (4 tests, 4 assertions)
```

The expected score of 5 is the sum of the differences the fixture actually introduces:
remote id, title, link, affected versions and date.

## Full local CI replay

Replayed the `CI`, `PHPStan` and `PHP Lint` jobs on PHP 8.4 (intl/zip/apcu) with MySQL 8 and Redis 6:

| | baseline (`main` 66183939) | this branch |
|---|---|---|
| PHPUnit | 726 tests, 2552 assertions, 1 skipped | 727 tests, 2553 assertions, 1 skipped |
| PHPStan | no errors | no errors |
| PHP Lint | pass | pass |

Both runs share one identical failure, `PackageControllerTest::testEdit`, which shells out to
`git clone --mirror git@github.com:composer/composer.git` and cannot reach GitHub from the
sandboxed container; it is green on this PR's GitHub CI run. Changed executable production
lines: 1/1 covered.

## Files changed

| File | Change |
|---|---|
| `src/Entity/SecurityAdvisory.php` | Exclude null CVEs from the identity shortcut |
| `tests/Entity/SecurityAdvisoryTest.php` | Cover two distinct CVE-less advisories for one package |

*Generated by Ora Studio — vibe coded by ousama*

### DIFF
--- a/tests/Entity/SecurityAdvisoryTest.php
+++ b/tests/Entity/SecurityAdvisoryTest.php
@@ -21,19 +21,6 @@ use PHPUnit\Framework\TestCase;
 
 class SecurityAdvisoryTest extends TestCase
 {
-    private const NULL_CVE_INITIAL_TITLE = 'Remote Code Execution';
-    private const NULL_CVE_UPDATED_TITLE = 'Cross-Site Request Forgery';
-    private const NULL_CVE_INITIAL_LINK = 'https://example.org/advisory/one';
-    private const NULL_CVE_UPDATED_LINK = 'https://example.org/advisory/two';
-    private const NULL_CVE_BRANCH = '1.x';
-    private const NULL_CVE_PUBLISHED_AT = 1494806400;
-    private const NULL_CVE_INITIAL_VERSIONS = ['<1.2'];
-    private const NULL_CVE_UPDATED_VERSIONS = ['>=2.0', '<2.1'];
-    private const NULL_CVE_REFERENCE = 'composer://acme/security-package';
-    private const NULL_CVE_INITIAL_FILENAME = 'acme/security-package/2017-05-15.yaml';
-    private const NULL_CVE_UPDATED_FILENAME = 'acme/security-package/2026-08-25.yaml';
-    private const ADVISORY_MATCH_THRESHOLD = 3;
-
     public function testCalculateDifferenceScore(): void
     {
         $data = [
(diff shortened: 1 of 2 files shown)


tests/Entity/SecurityAdvisoryTest.php
```diff
@@ -21,6 +21,19 @@
...
+    private const NULL_CVE_INITIAL_LINK = 'https://example.org/advisory/one';
+    private const NULL_CVE_UPDATED_LINK = 'https://example.org/advisory/two';
+    private const NULL_CVE_BRANCH = '1.x';
+    private const NULL_CVE_PUBLISHED_AT = 1494806400;
+    private const NULL_CVE_INITIAL_VERSIONS = ['<1.2'];
+    private const NULL_CVE_UPDATED_VERSIONS = ['>=2.0', '<2.1'];
+    private const NULL_CVE_REFERENCE = 'composer://acme/security-package';
+    private const NULL_CVE_INITIAL_FILENAME = 'acme/security-package/2017-05-15.yaml';
+    private const NULL_CVE_UPDATED_FILENAME = 'acme/security-package/2026-08-25.yaml';
+    private const ADVISORY_MATCH_THRESHOLD = 3;
```
glaubinix: All the const here look a bit out of place if you look at the remaining test setup and make the tests really hard to read

glaubinix: This const here definitely doesn't belong here. If you want to extract the threshold into a const then this need to be application code and used by both the resolver and the test case.

tests/Entity/SecurityAdvisoryTest.php
```diff
@@ -44,6 +57,34 @@ public function testCalculateDifferenceScore(): void
...
+            RemoteSecurityAdvisory::createFromFriendsOfPhp(self::NULL_CVE_INITIAL_FILENAME, $data),
+            FriendsOfPhpSecurityAdvisoriesSource::SOURCE_NAME,
+        );
+
+        $data['title'] = self::NULL_CVE_UPDATED_TITLE;
+        $data['link'] = self::NULL_CVE_UPDATED_LINK;
+        $data['branches'][self::NULL_CVE_BRANCH]['versions'] = self::NULL_CVE_UPDATED_VERSIONS;
+        $otherAdvisory = RemoteSecurityAdvisory::createFromFriendsOfPhp(self::NULL_CVE_UPDATED_FILENAME, $data);
+
+        $this->assertGreaterThan(self::ADVISORY_MATCH_THRESHOLD, $advisory->calculateDifferenceScore($otherAdvisory));
```
glaubinix: I am not sure I understand the assertion? If you want to assert the bug is fixed shouldn't you either assert that the score is`!=0` or assert it is the exact score you expect?

glaubinix (commented):
The fix is correct. Just tests need some more work

ousamabenyounes:
Thanks for the review @glaubinix. Addressed all three test comments in 8efc5ac6 (branch also rebased on latest `main`):

- Dropped the per-test constants, so the fixture reads like the neighbouring tests in this class again.
- Dropped `ADVISORY_MATCH_THRESHOLD` rather than duplicating the resolver's threshold in the test suite — the test no longer depends on it at all.
- Assert the exact expected score with `assertSame(5, ...)` instead of `assertGreaterThan(3, ...)`. 5 is the sum of what the fixture actually changes: remote id, title, link, affected versions and date.

Validation: RED on unmodified `main` (66183939) with only the new test applied — `Failed asserting that 0 is identical to 5` — then GREEN on this branch (4 tests, 4 assertions). Full local replay of the CI/PHPStan/Lint jobs on PHP 8.4 + MySQL 8 + Redis 6: 727 tests against the 726-test `main` baseline, no new failures. GitHub CI is green on 8efc5ac6.

glaubinix (approved):

PR #1820, opened 2026-08-28
Add filterable admin user directory (including frozen)

IonBazan:
## What this does

Replaces the dedicated `/admin/frozen-users` review queue with a general `/admin/users`
directory. The old frozen queue becomes one filter on the new page, so nothing is lost.

## Features

- User listing at `/admin/users`, 30 per page, each row linking to the public profile.
  Columns: roles, status, 2FA, GitHub ID, package count, registration date, last login.
- Search by username or email.
- Filters, laid out to match the transparency log page:
  - Freeze status (all, not frozen, any reason, or a specific reason like the old Temporary queue)
  - Registration date, from and to
  - 2FA enabled or disabled
  - GitHub account linked or not linked
  - GitHub ID (exact match on the stored numeric id)
- 2FA now shows as a Yes/No badge in the listing, and was added to the admin "User Metadata"
  block on the profile page, which previously only displayed 2FA when it was on.
- Section menu icon updated.

## Screenshots

image

image

## Notes

- Package counts use a correlated subquery so zero-package users still show up without row duplication.
- The GitHub username is not stored (it is fetched live from GitHub), so the filter and column
  use the numeric `githubId` we do keep.
- No schema changes. Tests cover the repository query builder and the controller filters.


Seldaek:
This looks good I think but it would be nice to still have a "Frozen users" menu item that shows the same page with preselected filters for frozen=temporary. This works as a reminder for users we need to get back to and unfreeze probably after they recovered their account, so would be nice to have it highlighted in the menu still.

Also needs a rebase now :)

Seldaek:
And needs porting to bootstrap 5 of course.. I got a claude review incoming with some more nitpicks.

src/Entity/UserRepository.php
```diff
@@ -30,23 +30,76 @@ public function __construct(ManagerRegistry $registry)
...
-                ->setParameter('reason', $reason);
+        if ($search !== null && $search !== '') {
+            $qb->andWhere('u.usernameCanonical LIKE :search OR u.emailCanonical LIKE :search')
+                ->setParameter('search', '%'.mb_strtolower($search).'%');
+        }
+
+        match ($frozenFilter) {
+            'none' => $qb->andWhere('u.frozen IS NULL'),
+            'any' => $qb->andWhere('u.frozen IS NOT NULL'),
+            'spam', 'bad_actor', 'temporary' => $qb->andWhere('u.frozen = :frozenReason')
```
Seldaek: **Hardcoded enum values turn a future freeze reason into a silent no-op.**

The template renders one option per `UserFreezeReason::cases()`, but these match arms hardcode the values. The enum's own docblock anticipates growth ("New cases can be added over time"), so say someone adds `case Impersonation = 'impersonation'`. The dropdown gains the option, selecting it sends `?frozen=impersonation`, and this arm doesn't match — so it falls to `default => null` and **no `WHERE` clause is added at all**. The page returns every user, including unfrozen ones, while the dropdown still shows "Impersonation" as selected. An admin reads that as "these are the impersonation freezes".

`UserFreezeReason::tryFrom($frozenFilter)` and matching on the enum rather than duplicating its values would make this immune to new cases — and pairs with returning a 400 on genuinely unknown values (separate comment below).

src/Entity/UserRepository.php
```diff
@@ -30,23 +30,76 @@ public function __construct(ManagerRegistry $registry)
...
+            ->from(Package::class, 'pkg')
+            ->innerJoin('pkg.maintainers', 'maint')
+            ->where('maint = u');
+
         $qb = $this->createQueryBuilder('u')
-            ->where('u.frozen IS NOT NULL')
-            ->orderBy('u.frozenAt', 'DESC')
+            ->select('u')
+            ->addSelect('('.$packageCountQb->getDQL().') AS packageCount')
+            ->orderBy('u.createdAt', 'DESC')
```
Seldaek: **Two problems with this ordering change — one behavioural, one performance.**

*The frozen review queue no longer surfaces new holds.* The old query was `WHERE u.frozen IS NOT NULL ORDER BY u.frozenAt DESC`. Now `frozen=temporary` orders by registration date, so a hold placed yesterday on a 2015 account sorts below an untouched 2026 signup — exactly backwards for a review queue. Compounding it, the menu entry now lands on the unfiltered list rather than defaulting to the Temporary queue the way `FrozenUserController` did.

That also leaves `frozenAt` orphaned: `git grep frozenAt` on this branch finds only the entity field, its accessors, the migration and `tests/Entity/UserTest.php` — no query and no template reads it. `migrations/2026_07_user_frozen_at.sql:17` states the index's only purpose: "Composite index serving `WHERE frozen = :reason ORDER BY frozenAt DESC` (the default queue view)". And `testGetFrozenUsersQueryBuilderOrdersByMostRecentlyFrozen` is deleted here without a replacement. Either order by `frozenAt DESC` when a freeze filter is active (and port that test), or drop the column and index in a migration — but leaving both in place with no reader is the worst of the three.

*`createdAt` has no index.* The only non-unique index on `fos_user` is `user_frozen_idx (frozen, frozenAt)` (`src/Entity/User.php:51`). The default `/admin/users` view has no `WHERE` at all plus `ORDER BY u.createdAt DESC`, so MySQL scans and filesorts the whole user table on every page load, degrading further with deep `OFFSET`. `ORDER BY u.id DESC` alone is equivalent for an autoincrement PK and is served by the primary key — or add an index on `createdAt` via a `migrations/*.sql` file.

src/Entity/UserRepository.php
```diff
@@ -30,23 +30,76 @@ public function __construct(ManagerRegistry $registry)
...
             ->addOrderBy('u.id', 'DESC');
 
-        if ($reason !== null) {
-            $qb->andWhere('u.frozen = :reason')
-                ->setParameter('reason', $reason);
+        if ($search !== null && $search !== '') {
+            $qb->andWhere('u.usernameCanonical LIKE :search OR u.emailCanonical LIKE :search')
+                ->setParameter('search', '%'.mb_strtolower($search).'%');
```
Seldaek: **`LIKE` metacharacters in the search term aren't escaped.**

Searching `a_b` produces `%a_b%`, which matches `axb`, `a9b`, ...; `100%` produces `%100%%`, matching every username or email containing `100`; and a trailing `\` produces an invalid escape sequence.

`src/QueryFilter/AuditLog/AbstractAdminAwareTextFilter.php:47` already has the fix:

```suggestion
                ->setParameter('search', '%'.addcslashes(mb_strtolower($search), '%_\\').'%');
```

src/Entity/UserRepository.php
```diff
@@ -30,23 +30,76 @@ public function __construct(ManagerRegistry $registry)
...
+        }
+
+        match ($frozenFilter) {
```
Seldaek: **Unrecognised filter values silently return the unfiltered list.**

`?frozen=frozen`, `?twofa=on`, `?twofa=1`, `?github_linked=true`, `?github_linked=linked` and `?registered_from=01/2026` all fall through to `default => null` (or `parseFilterDate`'s `return null`) and add no predicate. The page then lists every account while the form shows the filter as blank — so a typo'd or stale bookmarked query string reads as a legitimate filtered result.

The sibling admin listing sets the opposite convention at `src/Controller/Admin/FilterListController.php:61-72`: `FilterLists::tryFrom($listFilter)` then `throw new BadRequestHttpException('Unknown list')`. Worth validating in the controller and returning 400.

src/Entity/UserRepository.php
```diff
@@ -30,23 +30,76 @@ public function __construct(ManagerRegistry $registry)
...
+            ->select('COUNT(pkg.id)')
+            ->from(Package::class, 'pkg')
+            ->innerJoin('pkg.maintainers', 'maint')
+            ->where('maint = u');
+
         $qb = $this->createQueryBuilder('u')
-            ->where('u.frozen IS NOT NULL')
-            ->orderBy('u.frozenAt', 'DESC')
+            ->select('u')
+            ->addSelect('('.$packageCountQb->getDQL().') AS packageCount')
```
Seldaek: **Doctrine has `SIZE()` for this.**

These lines build a throwaway `QueryBuilder` purely to call `getDQL()` and splice the string into `addSelect()`. That construct silently drops parameters if the subquery ever gains one, and its correlation (`where('maint = u')`) depends on an outer alias the subquery builder knows nothing about.

`SIZE()` does the same job in one line, and `SizeFunction` counts against `maintainers_packages` directly — skipping the `INNER JOIN package` the current version pays for on every row:

```suggestion
        $qb = $this->createQueryBuilder('u')
            ->addSelect('SIZE(u.packages) AS packageCount')
```

`User::$packages` is the inverse side of `Package::$maintainers` (`ManyToMany`, join table `maintainers_packages`), so this counts the same rows.

The suggestion also drops `->select('u')`: `createQueryBuilder('u')` already selects `u`, and `select()` resets rather than appends, so that call is redundant either way.

src/Controller/Admin/UserController.php
```diff
@@ -0,0 +1,87 @@
...
+            $search !== '' ? $search : null,
+            $frozenFilter !== '' ? $frozenFilter : null,
+            $registeredFrom,
+            $registeredTo,
+            $twoFactorFilter !== '' ? $twoFactorFilter : null,
+            $githubId !== '' ? $githubId : null,
+            $githubLinkedFilter !== '' ? $githubLinkedFilter : null,
+        );
+
+        $users = new Pagerfanta(new QueryAdapter($qb, false));
```
Seldaek: **Missing `useOutputWalkers: false` makes the count query run the package subquery once per user.**

With the third argument left unset, `Paginator::useOutputWalker()` returns true and `CountOutputWalker` emits `SELECT COUNT(*) FROM (SELECT DISTINCT id_0 FROM (<the full original SQL>) dctrn_result) dctrn_table`. That inner SQL still carries the correlated `(SELECT COUNT(pkg.id) FROM package pkg INNER JOIN maintainers_packages ... WHERE maint = u)` select expression *and* the `ORDER BY`, so on the default unfiltered view MySQL evaluates the subquery for every row of `fos_user` just to produce a row count. `{{ users.nbResults }}` in the template means this runs on every page load.

Both sibling listings already pass it — `new QueryAdapter($qb, false, false)` in `FilterListController.php:88` and `TransparencyLogController.php:68` — which selects `CountWalker` instead: `COUNT(DISTINCT u.id)`, with the subquery and `ORDER BY` dropped.

```suggestion
        $users = new Pagerfanta(new QueryAdapter($qb, false, false));
```

src/Controller/Admin/UserController.php
```diff
@@ -0,0 +1,87 @@
...
+    private function parseFilterDate(string $value, bool $endOfDay): ?\DateTimeImmutable
+    {
+        if ($value === '') {
+            return null;
+        }
+
+        $date = \DateTimeImmutable::createFromFormat('!Y-m-d', $value);
+        if ($date === false) {
+            return null;
+        }
```
Seldaek: **`createFromFormat` doesn't reject out-of-range dates, it rolls them over.**

`\DateTimeImmutable::createFromFormat('!Y-m-d', '2026-02-31')` doesn't return `false` — it returns 2026-03-03. `'2026-13-01'` returns 2027-01-01. So the filter silently applies a date the admin never typed, and the corrected value gets echoed back into the form as though it were the input.

Check `getLastErrors()`, or round-trip the format and treat a mismatch as invalid:

```suggestion
        $date = \DateTimeImmutable::createFromFormat('!Y-m-d', $value);
        if ($date === false || $date->format('Y-m-d') !== $value) {
            return null;
        }
```

`DateTimeFromFilter::getDateTime()` sidesteps this by using `new \DateTimeImmutable($value)` in a try/catch, which throws on both inputs — another argument for reusing it.

src/Menu/MenuBuilder.php
```diff
@@ -83,9 +83,9 @@ public function createAdminMenu(): ItemInterface
             ]);
         }
         if ($this->security->isGranted('ROLE_DISABLE_USERS')) {
-            $menu->addChild('Frozen users', [
-                'label' => '<span class="icon-lock"></span>Frozen users',
-                'route' => 'admin_frozen_users',
+            $menu->addChild('Users', [
+                'label' => '<span class="icon-vcard"></span>Users',
+                'route' => 'admin_users',
```
Seldaek, line 88: **`/admin/frozen-users` disappears with no redirect.**

`GET /admin/frozen-users?reason=temporary` is the URL this menu entry used to produce, so it's what moderators working the Temporary hold queue will have bookmarked. After this PR it 404s with no hint that the page moved.

A small `#[Route('/admin/frozen-users')]` action that 301s to `admin_users` with `frozen=temporary` keeps the bookmarks working and restores the old default-queue landing spot at the same time.

templates/admin/users.html.twig
```diff
@@ -0,0 +1,158 @@
...
+                            <th>2FA</th>
+                            <th>GitHub ID</th>
+                            <th>Packages</th>
+                            <th>Registered</th>
+                            <th>Last login</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        {% for row in users %}
+                            {% set user = row[0] %}
```
Seldaek, line 131: **The freeze context is gone from the row, not just relocated.**

`templates/admin/frozen_users.html.twig` showed four freeze-context columns per row — frozen at, frozen by, the public reason, and a `ROLE_AUDITOR`-gated internal note — and `FrozenUserController::latestFreezeAuditByUserId()` batch-loaded the latest `UserFrozen` `AuditRecord` per displayed page to fill them. This row shows only `Frozen: <reason>`.

So an admin working the Temporary hold queue can no longer see when a hold was placed, who placed it, or the note explaining why, without opening each user's transparency log one at a time — which is the workflow the queue existed for. The PR description says nothing is lost, but this part is.

Porting `latestFreezeAuditByUserId()` and rendering those columns when a freeze filter is active (keeping the `ROLE_AUDITOR` gate on the note) would close the gap.

tests/Entity/UserRepositoryTest.php
```diff
@@ -37,34 +45,96 @@ public function testGetFrozenUsersQueryBuilderExcludesUnfrozenAndFiltersByReason
...
+    }
+
+    public function testGetUsersQueryBuilderIncludesPackageCount(): void
+    {
+        $user = self::createUser('withpkgs', 'withpkgs@example.org');
+        $this->store($user);
+
+        $rows = $this->userRepository->getUsersQueryBuilder(search: 'withpkgs')->getQuery()->getResult();
+        self::assertCount(1, $rows);
+        self::assertSame(0, (int) $rows[0]['packageCount']);
```
Seldaek: **This assertion passes on a broken subquery.**

The test is named for the package count but never creates a package, so only the zero path is covered — and after the `(int)` cast, a broken subquery, a `NULL`, or an empty string all satisfy `assertSame(0, ...)` equally well. `UserControllerTest::testListsUsersWithPackageCountAndProfileLink` is likewise named for the count but asserts nothing about it.

Nothing currently exercises `maint = u` returning non-zero, the outer-alias correlation, or the `AS packageCount` result key — which is the riskiest part of the new query. A user maintaining two packages asserting `2` would cover it.

Seldaek (commented):
Nice consolidation — the filter set is a real improvement over the reason-only frozen queue. A pile of notes below, but first the one that shapes the rest.

### Most of the filter findings share a root cause

Four of the inline comments (hardcoded freeze reasons, unescaped `LIKE`, unknown values silently ignored, rolled-over dates) are symptoms of one thing: the filtering is hand-rolled rather than going through `App\QueryFilter`.

`getUsersQueryBuilder(?string, ?string, ?DateTimeImmutable, ?DateTimeImmutable, ?string, ?string, ?string)` is seven positional parameters with stringly-typed magic values — `'none'`, `'any'`, `'enabled'`, `'disabled'`, `'yes'`, `'no'` — hand-duplicated between `UserRepository.php` and the `<option value=...>` lists in the template, with nothing keeping the two in sync. The docblock legend explaining which strings each parameter accepts only needs to exist because the API is stringly-typed.

Meanwhile `src/QueryFilter/QueryFilterInterface.php` plus `src/QueryFilter/AuditLog/{DateTimeFromFilter,DateTimeToFilter,AbstractAdminAwareTextFilter,...}` already implement `fromQuery(InputBag)`, `filter(QueryBuilder)`, `getKey()` and `getSelectedValue()` — the pattern `TransparencyLogController.php:41-60` drives in a loop, for the very page whose filter layout this PR is copying. Those filters are already generic apart from the hardcoded `a.` alias.

Suggestion: parameterise the alias, lift those filters out of the `AuditLog` namespace, add `UserFrozenFilter` / `TwoFactorFilter` / `GithubLinkedFilter` alongside them, and drive the set from `UserController` in the same loop the transparency log uses. That gets you enum-typed `tryFrom()` parsing, the existing `LIKE` escaping, `BadRequestHttpException` on unknown values, date parsing that actually rejects `2026-02-31`, and one source of truth for the template's option values — all four inline findings at once.

If you'd rather keep the current shape for now, each of those comments stands on its own and can be fixed in place.

The rest are independent: two perf items, the frozen-queue regressions, a bookmark-breaking route removal, and two tests that pass for the wrong reason.

tests/Controller/Admin/UserControllerTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $mod = self::createUser('mod2', 'mod2@example.org', roles: ['ROLE_DISABLE_USERS']);
+        $alice = self::createUser('alice', 'alice@findme.example.org');
+        $bob = self::createUser('bob', 'bob@example.org');
+        $this->store($mod, $alice, $bob);
+
+        $this->client->loginUser($mod);
+
+        $byUsername = $this->client->request('GET', '/admin/users?search=alice')->html();
+        self::assertResponseIsSuccessful();
+        self::assertStringContainsString('alice', $byUsername);
```
Seldaek: **This matches the form field, not a table row.**

After `GET /admin/users?search=alice` the template renders `<input type="text" name="search" ... value="alice">`, so `assertStringContainsString('alice', ...)` is satisfied by the echoed filter value regardless of what the listing contains. Break the search filter so it returns zero rows and this assertion still passes — only the `assertStringNotContainsString('>bob<')` line below keeps the test honest.

```suggestion
        self::assertStringContainsString('>alice<', $byUsername);
```

which matches the profile link text, the way the negative assertion already does for bob.

src/Entity/UserRepository.php
```diff
@@ -30,21 +30,17 @@ public function __construct(ManagerRegistry $registry)
...
-            ->orderBy('u.frozenAt', 'DESC')
-            ->addOrderBy('u.id', 'DESC');
+            ->addSelect('SIZE(u.packages) AS packageCount')
+            ->orderBy($orderByFrozenAt ? 'u.frozenAt' : 'u.id', 'DESC');
 
-        if ($reason !== null) {
-            $qb->andWhere('u.frozen = :reason')
-                ->setParameter('reason', $reason);
+        if ($orderByFrozenAt) {
+            $qb->addOrderBy('u.id', 'DESC');
```
IonBazan, line 43: We are using ordering by `id` here because `createdAt` doesn't have an index yet. This however would mean a table scan when using the registration date filter. I have not added it yet as this could be quite heavy operation and I don't have information about the table size.

Seldaek: Sounds fine really.. I don't think the order matters all too much in that table, rather aim for fast responses incl when filters are used.

IonBazan: Exactly, but the filter might still cause slow responses (no index). This should be fine for most cases but long term worth adding an index on `createdAt`

Seldaek:
Thanks :)

PR #1826, opened 2026-09-03
[RFC][UI] Add more package details to package list

### LINKED ISSUE #1290: Search Enhancements

### LINKED ISSUE #1368: Add a package maintainability score

IonBazan:
Package listings (a user's packages, explore/popular, vendor pages, dependents) and Algolia search results now show npmjs-inspired details:

- latest version, release date, license
- keyword tags

Adds filters for release date and licence:

image

Download/star counters are vertically centered:

image

## Caveats

- Search records only expose the new `meta.release`/`released`/`license` after their next indexing run; the hit template guards on it, so stale records render as before. A full `packagist:index --force` backfills everything.
- `bin/console algolia:configure` required for the new `meta.license` facet. The released numeric filter needs no settings change (Algolia allows numeric filtering on any numeric attribute).

- Algolia client upgraded: PHP `algolia/algoliasearch-client-php` 3 -> 4 (API rewrite - index name now passed per call) and JS `algoliasearch` v4 -> v5 / `instantsearch.js`-> v4.112. 
The search hit template moves to Preact JSX (`search.js` becomes `search.jsx`), with `preact` declared explicitly and the esbuild target raised for v5's async generators.

- V4's client rejects an empty appId, so .`env` Algolia keys are now `changeme `placeholders and .`env.test` carries dummy values (real values still come from `.env.local` / deploy env).

Replaces #1410
Somewhat fixes #1290 and fixes #1368

### DIFF
--- a/.env
+++ b/.env
@@ -43,10 +43,10 @@ SPAM_MODEL_FILE=
 MAILER_DSN=null://null
 ###< symfony/mailer ###
 
-ALGOLIA_APP_ID=
-ALGOLIA_ADMIN_KEY=
-ALGOLIA_SEARCH_KEY=
-ALGOLIA_INDEX_NAME=
+ALGOLIA_APP_ID=changeme
+ALGOLIA_ADMIN_KEY=changeme
+ALGOLIA_SEARCH_KEY=changeme
+ALGOLIA_INDEX_NAME=changeme
 
 # set those to values obtained by creating an application at https://github.com/settings/applications
 APP_GITHUB_CLIENT_ID=
--- a/.env.test
+++ b/.env.test
@@ -7,3 +7,7 @@ REDIS_URL=redis://localhost/14
 APP_MAILER_FROM_EMAIL=packagist@example.org
 APP_HOSTNAME=packagist.wip
 DEFAULT_URI=http://packagist.wip
+ALGOLIA_APP_ID=dummy
+ALGOLIA_ADMIN_KEY=dummy
+ALGOLIA_SEARCH_KEY=dummy
+ALGOLIA_INDEX_NAME=packagist
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-apcu": "*",
         "ext-ctype": "*",
         "ext-iconv": "*",
-        "algolia/algoliasearch-client-php": "^3.4",
+        "algolia/algoliasearch-client-php": "^4.0",
         "babdev/pagerfanta-bundle": "^4.2",
         "beelab/recaptcha2-bundle": "^2.3",
         "cebe/markdown": "^1.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,56 +4,48 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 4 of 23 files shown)


Seldaek:
I'm not sure this really adds so much value or just wastes screen real estate and search index space: 

- The license seems not super relevant given the prevalence of MIT in the PHP world, it is almost never a concern.
- Last version number gives you exactly zero information.
- Last release date is.. dubious IMO as a quality meter. Some things are very stable and released rarely, others might have  had a release a week ago, but it was the first and then the maintainer abandons the project.. So just a single date is not a great datapoint to make a decision on which to click. 

- Tags might add value tho, need to think some more about that one

This is mainly why these issues are still open, because I don't know what to do about them :)

IonBazan:
The screen space wastage is real concern here but... IMHO versions and release dates are quite helpful, while I would get rid of tags 😂 
Perhaps we could remove all of them from the display to save space but include them in Search filter facets like we do with tags now? I'm open for feedback and also feel free to experiment with this. Just a pattern I noticed in some of the issues past few years and a gap vs npmjs:

image

EdouardCourty:
Hi, I think this redesign makes sense. However, tags seem to use space for no added value.
Latest version and latest update date is very useful IMO.

Seldaek:
Why is it useful to you? What does this tell you?

EdouardCourty:
It is a triage signal. When you're comparing several packages in a list, opening each one just to check "is this still maintained?" doesn't scale.

Last release date lets you filter out abandoned packages in seconds. Same reason npm, PyPI and every other major registry surface it right on the listing.

Seldaek:
Ok but as per my comment above:

> * Some things are very stable and released rarely, others might have  had a release a week ago, but it was the first and then the maintainer abandons the project.. So just a single date is not a great datapoint to make a decision on which to click.

So I'd say maybe showing `released in the last year` vs `released over a year ago`, `released over 2/n years ago` is a good signal without being too specific?

And also easier to scan than distinguishing which is the newer of many exact dates. I see npm has relative dates which is also more readable than plain dates, but still I'd argue this might nudge you the wrong way just because by chance something was released shortly and another not for 6months because it's very stable.

IonBazan:
I agree that the release date is not the only parameter we should look at as there are plenty of solid packages which do not require constant updates (`psr/log` was last updated in 2024). Similarly, the version is just a number, but seeing a new tag on the package we previously known might inspire users to upgrade if they are stuck on old versions. 

Considering all above, and conserving the screen space, what do you think about revering the package list changes but keeping only the search filter/facets to allow users filtering by license and approximate latest release date? Once this change is deployed and packages are re-indexed, we can always revisit this in the future and modify the UI as needed.

Edit: We could also expose Aikido maintainability score, if that's possible - this parameter takes into account multiple variables and produces an easy to understand number. We won't need to re-invent the wheel.

IonBazan:
Replaced with #1863 as a subset of changes.

PR #1827, opened 2026-09-04
Replace React notifier with Bootstrap 5 toasts

IonBazan:
The notifier component was the only consumer of `react`, `react-dom` and `zustand`. 
Reimplement it on Bootstrap 5's Toast component keeping the same `log()`/`remove()` API, so all call sites stay unchanged. Drop the three dependencies and most of the custom `.notifications` styling. 

Using React to mount this tiny component was an overkill.

image

image

## Additional changes

- Fixes inconsistent border colour and thickness (because of the opacity) in nav login form:

Before:
Screen Recording 2026-09-04 at 16 22 39 (1)

After: 
Screen Recording 2026-09-04 at 16 31 05


Seldaek:
Thanks

IonBazan:
One thing I noticed is that long output toasts are not scrollable and because of fixed position, you can't see the whole content.   It's probably a one-line fix so I'll leave it to you. Sorry for the issue!

Seldaek:
No worries, it was bound to happen that we cause regressions here given all the changes :D https://github.com/composer/packagist/commit/06bcbd6d4a7918a7f1eb74aa96b55b3ca4ae257c

PR #1830, opened 2026-09-04
Catch the right Redis exception type in graceful-degradation blocks

### LINKED ISSUE #1829: ExploreController catches the wrong ConnectionException, so /explore 500s on a Redis outage
`src/Controller/ExploreController.php:19` imports `Doctrine\DBAL\ConnectionException` and uses it
for the two catches at `:56` and `:99` — but both of those `try` blocks wrap **Redis** calls, not
DB calls.

In DBAL 4.4.4, `Doctrine\DBAL\ConnectionException` is an empty class body that nothing in DBAL
ever throws:

```php
// vendor/doctrine/dbal/src/ConnectionException.php
class ConnectionException extends \Exception implements Exception
{
}
```

```
$ grep -rn "ConnectionException::" vendor/doctrine/dbal/src/
(no matches)
```

(It's the legacy root-namespace class, distinct from `Doctrine\DBAL\Exception\ConnectionException`,
which is the one the driver actually raises.)

So both catches are dead code, and the graceful degradation they were written for never happens:

- `exploreAction` (`:47-58`) — `$redis->zrevrange('downloads:trending', 0, 9)` was meant to fall
  back to `$popular = []`.
- `popularAction` (`:78-101`) — `$redis->zrevrange(...)` / `$redis->zcard(...)` were meant to fall
  back to an empty Pagerfanta.

Both raise `Predis\Connection\ConnectionException`, which propagates → **`/explore` and
`/explore/popular` 500 on a Redis outage** instead of rendering without the popular list.

### Suggested fix

Catch `\Predis\PredisException` rather than `Predis\Connection\ConnectionException`. Predis raises
`Response\ServerException` for `-LOADING`, `-MISCONF`, `-OOM` and `-READONLY`, and that is a
*sibling* of the connection error, not a subclass:

```
PredisException (abstract)
├── CommunicationException (abstract)
│   └── Connection\ConnectionException
└── Response\ServerException
```

So `catch (ConnectionException)` misses a Redis restart/failover (`LOADING Redis is loading the
dataset in memory`) even where the import is correct.

### Same narrow-catch pattern elsewhere

These all use `Predis\Connection\ConnectionException` correctly (unlike ExploreController) but
still let `ServerException` through, so a `LOADING` / `OOM` / `READONLY` reply escapes:

- `src/Controller/WebController.php:184` (`/statistics`)
- `src/Controller/PackageController.php:421`, `:584`, `:684`, `:802`, `:810`
- `src/Controller/Controller.php:70` (`getPackagesMetadata`, so every package listing)
- `src/Security/RecaptchaHelper.php:55`
- `src/Security/Voter/PackageVoter.php:88`
- `src/Validator/PopularPackageSafetyValidator.php:64`
- `src/Model/PackageManager.php:161`

Widening them to `\Predis\PredisException` looks safe everywhere — in every case the catch body is
already the "Redis is unavailable, carry on without it" path.

Found while reviewing #1823; unrelated to that PR.

lazerg:
ExploreController imports Doctrine\DBAL\ConnectionException for its two Redis-related catches, but that class is an empty legacy type nothing in DBAL throws. The actual failure is Predis\Connection\ConnectionException, so a Redis outage produced an uncaught 500 on /explore instead of degrading gracefully. Predis also raises Response\ServerException as a sibling of the connection error (for LOADING/MISCONF/OOM/READONLY replies), so even the narrower Predis\Connection\ConnectionException catch used elsewhere would still miss a restart or failover.

This switches ExploreController to catch Predis\PredisException, and widens the same narrow catch in WebController, PackageController, Controller, RecaptchaHelper, PackageVoter, PopularPackageSafetyValidator and PackageManager, since each of those catch bodies is already a plain "Redis unavailable, continue without it" fallback.

Fixes #1829

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -67,7 +67,7 @@ abstract class Controller extends AbstractController
             if ($search) {
                 $favorites = $favMgr->getFaverCounts($ids);
             }
-        } catch (\Predis\Connection\ConnectionException $e) {
+        } catch (\Predis\PredisException $e) {
         }
 
         return ['downloads' => $downloads, 'favers' => $favorites];
--- a/src/Controller/ExploreController.php
+++ b/src/Controller/ExploreController.php
@@ -16,10 +16,10 @@ use App\Entity\Package;
 use App\Entity\Version;
 use App\Model\DownloadManager;
 use App\Model\FavoriteManager;
-use Doctrine\DBAL\ConnectionException;
 use Pagerfanta\Adapter\FixedAdapter;
 use Pagerfanta\Pagerfanta;
 use Predis\Client as RedisClient;
+use Predis\PredisException;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -53,7 +53,7 @@ class ExploreController extends Controller
                     return array_search($a->getId(), $popularIds) > array_search($b->getId(), $popularIds) ? 1 : -1;
                 });
             }
-        } catch (ConnectionException $e) {
+        } catch (PredisException $e) {
             $popular = [];
         }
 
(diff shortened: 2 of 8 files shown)


Seldaek:
Thanks

PR #1831, opened 2026-09-04
Minor BS5 and chart migration fixes

IonBazan:
1. Fixes chart responsiveness issue reported in https://github.com/composer/packagist/pull/1812#issuecomment-5539960365
>happens on mobile:
image
3. Fixes nav login and other form inputs - inconsistent border colour and thickness (because of the opacity) - cherry-picked from #1827

Before:
Screen Recording 2026-09-04 at 16 22 39 (1)

After: 
Screen Recording 2026-09-04 at 16 31 05

### DIFF
--- a/css/charts.css
+++ b/css/charts.css
@@ -35,10 +35,6 @@
     display: block;
     margin: 2px 0px 1px;
 }
-.version-stats-chart {
-  width: 70%;
-  margin-right: 5%;
-}
 .stats-toggle-wrapper {
     display: flex;
 }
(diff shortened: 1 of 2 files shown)


css/charts.css
```diff
@@ -35,10 +35,6 @@
     display: block;
     margin: 2px 0px 1px;
 }
-.version-stats-chart {
-  width: 70%;
```
IonBazan, line 39: This was causing the issue as width is now controlled by `col-12 col-lg-9` classes on the same div

Seldaek:
Thanks

PR #1835, opened 2026-09-07
Upgrade algolia/algoliasearch-client-php to 4.x

Seldaek:
v4 removes initIndex() and the SearchIndex object: every operation is now a method on the client taking the index name. Rather than translate each call site, all contact with the client moves behind App\Search\PackageIndex, so a future client upgrade touches one class and tests stop subclassing vendor code.

Search parameters now go in the request body argument rather than the request options. v3 folded unrecognised request options into the body; v4 discards them without complaint, which would have silently dropped filters and pagination. Query::getOptions() is renamed getSearchParams() so the two are not confused.

packagist:index --force is disabled. It called $index->clear(), which never existed on the v3 client either, so it has always died before clearing anything. Pointing it at clearObjects() would genuinely empty the live index for the hours a full reindex takes, so that wants doing atomically as its own change. --all still reindexes everything in place.

packagist:clean-index now browses with a cursor instead of paging by hand. The old loop drifted every time it deleted a record and could not see past paginationLimitedTo, so it only ever examined the first 300 of the ~557 stale virtual packages. It also checks each record's type before deleting, so it stays safe even if the query's filters were to stop applying.

The client is registered lazily because it throws when constructed without credentials. PackageManager depends on it transitively, so every package page would otherwise fail for anyone with no Algolia config; now only search itself degrades, and .env needs no placeholder credentials.

saveRecords() rejects records with no objectID, which v3 did client-side. v4's saveObjects() would instead let Algolia generate one, creating a record we could never update or delete again.

WebController and PackageManager also catch \InvalidArgumentException, which v4 throws for missing required parameters and for malformed response bodies.

### DIFF
--- a/src/Search/PackageIndex.php
+++ b/src/Search/PackageIndex.php
@@ -1,0 +1,57 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Packagist.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *     Nils Adermann <naderman@naderman.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Search;
+
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+
+/**
+ * Read/write access to the package search index.
+ *
+ * All contact with the search vendor's client goes through implementations of this, so that a client
+ * upgrade touches one class instead of every call site.
+ */
+interface PackageIndex
+{
+    /**
+     * @param array<string, mixed> $searchParams the request body: query, filters, hitsPerPage, page, ...
+     *
+     * @return array<string, mixed> the raw search response; callers narrow it to the shape they asked for
+     *
+     * @throws AlgoliaException          on a transport or API error
+     * @throws \InvalidArgumentException on a malformed response body
+     */
+    public function search(array $searchParams): array;
+
+    /**
+     * Iterates the whole index with a cursor, unaffected by records being deleted while iterating.
+     *
+     * @param array<string, mixed> $browseParams
+     *
+     * @return iterable<array<string, mixed>>
+     */
+    public function browse(array $browseParams): iterable;
+
+    /**
+     * @param list<array<string, mixed>> $records each must carry an objectID
+     */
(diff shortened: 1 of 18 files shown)


src/Search/PackageIndex.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * All contact with the search vendor's client goes through implementations of this, so that a client
+ * upgrade touches one class instead of every call site.
+ */
+interface PackageIndex
+{
+    /**
+     * @param array<string, mixed> $searchParams the request body: query, filters, hitsPerPage, page, ...
+     *
+     * @return array<string, mixed> the raw search response; callers narrow it to the shape they asked for
```
stof: Why not defining a phpstan type alias for the array shape of the algolia object ?

the `AlgoliaSearchResponse` type in ResultTransformer already encodes this type nested in its shape anyway.

src/Search/PackageIndex.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Read/write access to the package search index.
+ *
+ * All contact with the search vendor's client goes through implementations of this, so that a client
+ * upgrade touches one class instead of every call site.
+ */
+interface PackageIndex
+{
+    /**
+     * @param array<string, mixed> $searchParams the request body: query, filters, hitsPerPage, page, ...
```
stof: Should this use the `SearchParams` phpstan type of the `Query` class ?

src/Search/PackageIndex.php
```diff
@@ -0,0 +1,57 @@
...
+     * Iterates the whole index with a cursor, unaffected by records being deleted while iterating.
+     *
+     * @param array<string, mixed> $browseParams
+     *
+     * @return iterable<array<string, mixed>>
+     */
+    public function browse(array $browseParams): iterable;
+
+    /**
+     * @param list<array<string, mixed>> $records each must carry an objectID
```
stof: `list<array{objectID: int, ...<string, mixed>}>` would enforce this in static analysis  (and maybe the full shape of the object could be described instead, enforcing consistent records)

src/Search/PackageIndex.php
Seldaek: Yup, more types are in now

PR #1841, opened 2026-09-07
Drop the Composer 1 default branch fallback from download tracking

Seldaek:
`trackDownloadsAction` resolved Composer 1's `9999999-dev` default branch with a `SELECT` against `package` + `package_version` — **per package, inside the per-package loop**, on an endpoint taking ~2M requests per APM period.

It was the only synchronous MySQL left there (everything else resolves ids through a single Lua call against Redis, and the write is a single `EVALSHA`), and with `PDO::ATTR_TIMEOUT` at 1.3s it's the most likely explanation for the endpoint's 10.6s tail.

Support for Composer 1 was shut down on 2025-09-01, which `packages.json` has been announcing via its `warning` field since.

> [!IMPORTANT]
> This is a behaviour change, not a pure removal. The few remaining Composer 1 clients that hit this path now get a `partial` response and their downloads stop being counted, instead of being resolved via the package's default branch. Given you said a few requests do still land here, that's your call — happy to close this one if you'd rather keep the resolution and instead seed a `9999999-dev` alias into the Redis `ids:` hash, which would keep the behaviour with no MySQL.

### Verification

`composer phpstan` clean, full suite green. The endpoint had **no** test coverage at all, so this adds the two cases that matter: a resolvable version is counted (201 `success`), and an unresolvable one is reported back (200 `partial`).

### DIFF
--- a/src/Controller/ApiController.php
+++ b/src/Controller/ApiController.php
@@ -305,15 +305,6 @@ class ApiController extends Controller
         ];
         $jobs = $failed = [];
         foreach ($payload as $package) {
-            // support legacy composer v1 normalized default branches
-            if ($package['version'] === '9999999-dev' && !isset($package['id'], $package['vid'])) {
-                $result = $this->getDefaultPackageAndVersionId($package['name']);
-                if ($result) {
-                    $package['id'] = $result['id'];
-                    $package['vid'] = $result['vid'];
-                }
-            }
-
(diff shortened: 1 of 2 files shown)


Seldaek:
I think this is an acceptable loss, composer 1 client are not so many anymore and this only affects stats for the default branch so 🤷🏻‍♂️

Issue #1817, opened 2026-08-26
PIE package details not matching the rest of UI


IonBazan:
Details like language, Ext name, etc are not sharing the same style as number of installs.
![image](https://github.com/user-attachments/assets/b4d5788d-b8ea-465f-bde6-ae2191313325)

Similarly, install description could need some restyling too:
![image](https://github.com/user-attachments/assets/402cc4cd-e870-4ca1-a6d4-c6f227ceed33)

Issue #1750, opened 2026-06-02
modern design for packagist


shyim:
Hey,

I would like to use this issue to start a discussion around redesigning packagist. When you use packagist, you see and feel it's already got aged pretty well and was built by backend developers (dont understand me wrong, before AI + bootstrap my pages looked similar). 

I believe packagist deserves to look more modern, to represent that PHP is still cool in 2026. With this issue, I would like to push to get the community opinions, and maybe find a designer(s), developer(s) which implement it (I volumeteer already to help out).

## My PoC with AI

I tried my luck of redesigning with Claude Design and build a prototype as maybe an example how I would imagine a future.

Pages:
- [Startpage](https://packagist.shyim.workers.dev/)
- [Package Detail](https://packagist.shyim.workers.dev/p/laravel/framework)
- [Vendor page](https://packagist.shyim.workers.dev/v/laravel)
- [Maintainer page](https://packagist.shyim.workers.dev/u/taylorotwell)

Of course many things are missing:
- Entire account stuff
- Detailed Package Stats
- Docs, how to submit packages and so on

**It's just an starting idea, where the design could go**

Seldaek:
I agree it'd be great obviously to have a wonderful new design. BUT we have a ton of way more important things to do at the moment, feature & security wise, so I really don't want to get sidetracked by this.

The problem with those PoC pages is that yes at first sight it looks great, but if I spend 2min looking at it I start to notice things that don't make any sense. It feels like stuff was just dropped in random locations, there is not much coherence or big picture view that went into it. That makes total sense for an early draft especially if it was AI driven but even if done by yourself I don't expect you'll have the same background knowledge that I do here.. What I am however trying to say is that this is not so easy to outsource to someone else IMO, at least not 100%. 

So I don't really want to say "yes sure go for it" *right now* because I know it'll end up being a time drain for me anyway. But hopefully we can keep this in mind and come back to it in a while.

In the meantime if you do want to help, there are some small pieces that already could use some love like the download stats pages for ex that's a fairly isolated thing, and it is in dire need of a rewrite with a newer charting lib + redesigning it would also be great while at it. That might be a good starting point, and definitely easier to digest/review than a full rewrite of the whole site's UI :D

Docs pages are also good to improve in isolation IMO, or looking at the profile/user pages too afterwards as we plan on working on organization stuff that'll need a lot of new pages so having account pages redesigned would help. But while tempting I don't think the package page should be the first candidate as these need much more thought :)

PR #1844, opened 2026-09-07
Package page reflow

Seldaek:
Not really a huge redesign but trying to make things flow a bit better to avoid this huge gap between description and version info, that leads to wasted vertical space..

image

### DIFF
--- a/templates/package/view_package.html.twig
+++ b/templates/package/view_package.html.twig
@@ -26,14 +26,16 @@
 {%- endblock %}
 
 {% set hasActions = is_granted('ROLE_EDIT_PACKAGES') or is_granted('ROLE_DISABLE_PACKAGES') or is_granted('ROLE_UPDATE_PACKAGES') or package.maintainers.contains(app.user) %}
+{% set repoUrl = package.browsableRepository %}
+{% set isGitHubRepoUrl = package.isGitHub() %}
+{% set displayOpenIssues = version and version.support is not null and version.support.issues is defined and isGitHubRepoUrl and version.support.issues starts with repoUrl %}
 
 {% block content %}
     <div class="row" id="view-package-page">
         <div class="col-12 package"{% if app.user and package.crawledAt is null and is_granted('edit', package) and not package.isFrozen() %} data-force-crawl="true"{% endif %}>
             <div class="package-header">
-
-                <div class="row">
-                    <div class="col-lg-9">
+                <div class="row align-items-end g-2">
+                    <div class="col-12 col-md">
                         <h2 class="title">
                             {% if is_favorite is defined %}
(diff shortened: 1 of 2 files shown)


EdouardCourty:
Why not push the README and current version content where you put the 3 boxes, and move the boxes's data in the right panel ?

Right now the Security and Fund package maintenance cards are quite empty and lead to empty space again

templates/package/view_package.html.twig
```diff
@@ -87,288 +99,282 @@
...
 
-                    <div class="row package-aside">
-                        <div class="facts col-12 col-md-6 col-lg-12">
-                            <h5>Statistics</h5>
-                            {% if downloads.total is defined %}
-                                <div class="package-downloads-chart js-downloads-chart"></div>
-                            {% endif %}
+            <div class="row package-facts">
+                <div class="col-12 col-md-6 col-lg">
+                    <div class="package-aside details">
```
IonBazan, line 257: While here, we could add `.facts` here too. That should solve #1817

IonBazan (commented):
I agree with @EdouardCourty - especially for PIE packages, funding and security info will waste too much space. I would take some inspiration from #1750 and move the stats below the name instead. Stars, Forks, Issues and Website could be easily replaced with icons and placed inline to save space like it's done in https://packagist.shyim.workers.dev/p/symfony/console. That design has its flaws but I like the "stats strip" with License, Dependencies, PHP version etc.

templates/package/view_package.html.twig
Seldaek: Hah yeah I just noticed that problem as well and got a fix locally :)

Seldaek:
I mean just as a reference.. this is the amount of empty space we have today :D 
image

So I think it's a lot better already, and yes the security block is a bit empty right now but that may have more later.

Anyway I will do some more improvements to gain a few more vertical pixels, but moving the readme on top is IMO a no-go because the length of the readme is very unpredictable.

IonBazan:
How about we move version info and readme within that Left column itself? Currently it starts only after "facts-boxes" but we could treat that whole right section as a sidebar. The consequence would be that readme has a little less horizontal space, and responsiveness could be a bit more tricky but maybe worth a shot.

Something like this:
image

Looks a bit too crowded but with right paddings should render quite well.

Seldaek:
Yeah problem with that is if you want to go through a few versions to see how metadata changed, the version list might require you to scroll so far that you do not see the version metadata anymore, and then you have to scroll back and forth. So nope from me. Anyway I'll deploy this PR already and then see. I don't really want to go down the rabbit hole of redesigning everything right now as I have lots of other stuff I gotta work on too.

Seldaek:
before
image

after
image

IonBazan:
Looks good, and I love the maintainer controls collapsed into dropdowns 👏🏻

PR #1845, opened 2026-09-08
chore: Update Rector config

IonBazan:
The Rector config was not touched since early 2023, when we migrated to PHPUnit 10 and seems to be never used since then. In the meantime, Rector v2 was released with simpler config and easier framework version upgrade control. 

However, since default rules and new framework-specific rules have changed, there are a lot of changes that would be executed at the moment, hence they are not included in this change. I leave it to the maintainer to run afterwards.

Additionally, replaced class FQCN strings in tests with `::class` pseudo-constants.

Note: Rector is not currently required via `composer.json` - it relies on globally installed version instead. We could change that and add a dry-run in the CI if that's necessary.


Seldaek:
Thanks yeah I'll see about running it myself, I don't really like having this stuff in CI

PR #1852, opened 2026-09-15
Add support request workflows and an admin queue

Seldaek:
Adds a /contact page with a guided form per type, feeding a task list at /admin/support/ with internal notes, replies and resolution. The plain email address stays, at the top of the page.

Lost 2FA is requested from the two-factor login prompt, where the firewall has already verified the password, so the request provably comes from someone holding it. That is the first path on the site turning password-only into full account access, so it also gets:

 - an alert to the account's address the moment the request is filed, carrying no requester text, with a single-use link that cancels the request and drops every session
 - a 24h hold before approval on high-value accounts (>=10k downloads on a package, or org membership), enforced server-side and frozen onto the row
 - a risk panel showing downloads, sole-maintainer status, 2FA age and recent security events, so "usually just approve" stays informed

The other three types deep-link to the existing UI rather than growing a second mutation path: transfers to the package transfer form, deletions to the profile delete form. Accounts with no packages are sent to self-service deletion instead of filing a ticket, and vendor claims are rejected up front when the requester can already publish in that namespace.

Queue access is per request type, routed to the role that can carry the action out, so a delegated moderator sees only their own work. ROLE_EDIT_PACKAGES and ROLE_DISABLE_2FA join ADMIN_ROLES for that reason; neither exposes anything new, as both already grant their capability outside /admin/.

At most one open request per user and type is a database invariant, via a generated column under a unique index, so a resubmit cannot duplicate a row or re-notify. Nothing about support requests reaches audit_log, which every logged-in user can read: the free text stays in the support tables, and the actions we take are already audited by the code they call into.

Also fixes twoFactorAuthenticationDeactivated() to record userId alongside actorId. Without it the risk panel's userId-filtered query misses every prior deactivation. Purely internal: the transparency log excludes both 2FA types unconditionally, and its user filters resolve through audit_log_search.


Seldaek:
Follow-up ideas:

- requesting package unfreeze (should be link from package page for maintainers)
- requesting URL change on edit of the URL for popular packages, instead of blocking the edit it would file a ticket and tell the user (ideally without causing noise when our github redirect detection would anyway update the URL by itself at next update) and once ticket is approved the edit actually happens
- requesting package deletion in bulk (ideally support vendor/* and vendor/package list like for ownership claim)
- moving the typosquatting detection from an email to admins to a new item in the support queue (created by nobody), so it can be marked resolved and done

PR #1856, opened 2026-09-17
Move the log display layer to App\Log so both audit and transparency log can share it

IgorBenko:
Groundwork for the follow-up transparency log PR, which projects audit log entries into a dedicated transparency log that renders most events through the same display classes. Move and rename only, no behaviour changes.
104 of the 133 files are renames, the other 29 are changed only for imports and include paths.

## Changes

**Display layer: `App\Audit\Display` -> `App\Log\Display`**
- The 44 per-event classes go into `src/Log/Display/Event/`.
- The shared pieces (`ActorDisplay`, `OrganizationDisplay`, `AuditLogDisplayFactory`) stay
  one level up in `src/Log/Display/`, so the event classes and the factory now import each
  other explicitly.
- `AuditLogDisplayInterface` -> `LogDisplayInterface` and `AbstractAuditLogDisplay` ->
  `AbstractLogDisplay`. Neither is specific to the audit log any more.

**Templates**
- `templates/audit_log/display/` -> `templates/log/display/`, same for `macros.html.twig`.
- `_transparency_log_table.html.twig` -> `log/_table.html.twig`.
- The macro import alias goes from `auditLog` to `log`.

**`AuditRecordType` -> `App\Log\AuditLogEventType`**


Seldaek:
Sorry this needs a rebase.

IgorBenko:
No worries, I rebased it now.

Seldaek:
Thanks

PR #1857, opened 2026-09-17
Show update progress outside the Manage dropdown

Seldaek:
The package page reflow moved the Update action into the Manage menu, and its spinner went with it. Bootstrap keeps the menu open when an input is clicked so it showed briefly, but updates run for many seconds and the indicator vanished as soon as the menu closed. The data-force-crawl auto-update on never-crawled packages showed nothing at all, since nobody opens the menu for it.

The Manage toggle now swaps its glyph for a spinner while a job runs, and a toast reports the same thing, so progress is visible with the menu closed.

The loading state was previously cleared only on job completion, so a failed PUT or job poll left the spinner running forever - and since that class doubled as the re-entrancy guard, Update stayed dead until a reload. Both paths now clear it. They use .fail() rather than an error option, which would have replaced the global $.ajaxSetup handler that surfaces the server's message.

Retires the loader GIFs for bootstrap's spinner-border, via a mixin painting it onto a pseudo-element. It inherits currentColor, so one definition covers the dark toggle and the light contexts. The submit buttons become `<button>` because an `<input>` cannot host the pseudo-element - that is why these used a background image in the first place.


stof:
The PR description misses some code spans around HTML tag names, so that they don't become HTML elements.

PR #1865, opened 2026-09-19
Degrade to an unsorted dependents listing when the sort times out

Seldaek:
phpunit/phpunit has ~163k dependents, so ordering them exceeds the 5s statement timeout and the page 503s until the buffer pool happens to be warm. Fetching the rows is cheap though; it is only the ordering that is unaffordable. So on ER_QUERY_TIMEOUT the page now re-runs the query with no ORDER BY, dropping the download join along with it since that join exists only to sort on, and renders the listing with a warning that it could not be ordered.

Not offered for .json, which keeps returning the 503: a client cannot see the warning, and silently handing it arbitrarily ordered rows is worse than an error. The count is guarded separately because a count has no cheaper variant to fall back to, and the fallback itself stays bounded so a pathological case moves the stall rather than removing it.

Row order is formally undefined without ORDER BY, so paging a degraded listing may repeat or skip a row. Taken deliberately: ORDER BY p.id would reintroduce the filesort over all 163k rows that this avoids.

Also selects p.type, which macros.html.twig reads and every other listing supplies via LISTING_FIELDS. Without it the dependents page is a 500 under strict_variables (dev/test) and in production silently loses the PIE extension indicator, since package.type evaluates to null there. This is what made the page renderable in a test at all.

Tying the ORDER BY to the join that supports it also lets phpstan-dba analyse the query, so its baselined "Unknown column 'd.total'" entry is gone.


templates/package/dependents.html.twig
```diff
@@ -16,8 +16,8 @@
                     <span class="filters">
                         <span class="filter">
                             Order by:
-                            {% if order_by == 'name' %}<span class="active">name</span>{% else %}<a href="{{ path("view_package_dependents", {name: name, order_by: 'name', requires: requires}) }}">name</a>{% endif %} |
-                            {% if order_by == 'downloads' %}<span class="active">downloads</span>{% else %}<a href="{{ path("view_package_dependents", {name: name, order_by: 'downloads', requires: requires}) }}">downloads</a>{% endif %}
+                            {% if order_by == 'name' and not unsorted|default(false) %}<span class="active">name</span>{% else %}<a href="{{ path("view_package_dependents", {name: name, order_by: 'name', requires: requires}) }}">name</a>{% endif %} |
```
stof, line 19: using `|default` with a boolean value is risky, because `|default(true)` does not work at all (`false` is considered empty in Twig, and triggers the default).
It is recommended to use the `??` operator rather than the `default` filter in most cases to avoid surprising behavior.

Seldaek: Followed up in #1871. You're right about the trap; worth noting the current code is correct either way, since `|default(false)` is the safe direction — `false` being empty means the default returns `false` too. It's `|default(true)` that misfires, and there are none in the codebase. Switched to `??` anyway so it reads as what it does rather than relying on that.

